Issue #2, opened 2017-10-12
fortio improvements


ldemailly:
Fortio Improvement list:

- [ ] a request id in either headers or query arg or both, way to get cached/non cached request
- [ ] calculate bytes sent/received including headers (to be researched)
- [ ] h2 testing (#452)
- [ ] go over the TODOs
- [ ] warm up time
- [ ] payload option (exists in grpcping)
- [ ] fortio's own perf regression test (ie > 100k qps)

Done:
- [x] curl / single request mode - #21
- [x] ^C support to stop / option to run continuously until ^C - #5 
- [x] struct, json or other structured output for the histogram - #9
- [x] graph (picture) generation - #14
- [x] evaluate/use https://github.com/spf13/cobra -> no - min dependencies
- [x] fixed number of requests mode - #82 
- [x] consider passing a context and/or using channels - #64 

originally from https://github.com/istio/istio/issues/453

ymesika:
@ldemailly `h2` == `HTTP/2` ?

ldemailly:
yes. last time I checked, the default go client/server only did h2 (http 2.0) on tls (https) connections
grpc does exercise h2 though

ymesika:
That’s correct in most implementations and all browsers. I guess TLS is prerequisite to h2 and I noticed it’s still not supported in fortio.

ldemailly:
@ymesika you can use -grpc or -stdclient for tls/h2 today, does it not work for you ?

ps: h2 without tls (h2c) is part of the spec, just that go client isn't supporting it
( https://github.com/golang/go/issues/14141 )

ldemailly:
maybe also calculate https://en.wikipedia.org/wiki/Apdex

ldemailly:
another couple of ideas:
- warm up time, and use the (end of the) warm up to auto set the histogram parameters
- ability to add datasets to the UI / add runs - maybe pop up the form with run parameters and
- label the runs

ldemailly:
> ability to add datasets to the UI / add runs / label the runs

done in #37

ldemailly:
- consider passing a context and/or using channels done in #64 
- fixed number of requests mode done in #82

ldemailly:
getting close to 1.0 - I think after #85, #88, #89, #91, (#92) it'd be enough feature for 1.0 (after some stability baking)

ldemailly:
closing this one 🎉 - filled remaining ideas as seperate new tasks

PR #5, opened 2017-11-02
Support for ^C and otherwise continuous run

### LINKED ISSUE #6: ^C support to stop / option to run continuously until ^C

ymesika:
This PR adds the following new features (as suggested in Issue #2):
- Support stopping running load with ^C
- New 0 == infinity duration `-t 0` flag will run the load without set duration time (until ^C)

In both interrupts the info collected until interrupt will be printed.

Fixes #6


fortio_main.go
```diff
@@ -94,6 +95,15 @@ func main() {
 		usage("Unable to extract percentiles from -p: ", err)
 	}
 
+	// Following code determines whether the t flag has been provided together with the endless flag as they are
+	// both mutual exclusive. durationFlag can't be tested because if t flag is omitted the durationFlag will hold
+	// the default value.
```
ldemailly: it's nice but that seems like a lot of code just to have an error, maybe it's fine to just have endless overwrite the duration
(or we could test if the duration isn't the default anymore)

fortio_main.go
```diff
@@ -113,12 +123,19 @@ func fortioLoad() {
...
 	url := flag.Arg(0)
+	duration := *durationFlag
 	prevGoMaxProcs := runtime.GOMAXPROCS(*goMaxProcsFlag)
-	fmt.Printf("Fortio %s running at %g queries per second, %d->%d procs, for %v: %s\n",
-		fhttp.Version, *qpsFlag, prevGoMaxProcs, runtime.GOMAXPROCS(0), *durationFlag, url)
+	fmt.Printf("Fortio %s running at %g queries per second, %d->%d procs",
+		fhttp.Version, *qpsFlag, prevGoMaxProcs, runtime.GOMAXPROCS(0))
+	if *endlessFlag {
+		fmt.Printf(", until interrupted: %s\n", url)
+		duration = 0 // Endless mode is determined by having duration value equal to 0
```
ldemailly: maybe we can just use that ? like -qps 0 means infinite/max qps; -t 0

yes, let's do that - it's simpler no ?
(remove the new flag altogether, just add to the help text 0 for infinite - you can still have the if for the nice message like happens for qps 0 too)

periodic/periodic.go
```diff
@@ -219,9 +233,24 @@ func runOne(id int, funcTimes *stats.Histogram, sleepTimes *stats.Histogram, num
...
+	terminated := false
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		<-c
+		terminated = true
+	}()
+
 	for {
+		if terminated {
```
ldemailly: can you run fortio -qps 0 against itself on your machine before/after this change, I would like to know how expensive the channel/synchronization is

ldemailly (commented):
thanks very much for this, see below for a change request (use -t as the only flag) and a question about cost/perf impact of the change

ldemailly:
also please change fhttp/http.go
to
```
        Version       = "v0.3.0"
```

ldemailly:
(created a separate issue #6 and updated your summary)

periodic/periodic.go
```diff
@@ -219,9 +233,24 @@ func runOne(id int, funcTimes *stats.Histogram, sleepTimes *stats.Histogram, num
...
 	useQPS := (perThreadQPS > 0)
 	f := r.Function
+
+	// Catch SIGINT signals from the OS and raise a flag to terminate the run loop
+	terminated := false
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		<-c
+		terminated = true
```
ldemailly: more importantly: while this works on x86 I don't think it's safe (unprotected read/write across go routines)

edit: checked that indeed running with -race shows the race

periodic/periodic.go
```diff
@@ -117,7 +118,7 @@ func newPeriodicRunner(opts *RunnerOptions) *periodicRunner {
 	if r.Resolution <= 0 {
 		r.Resolution = DefaultRunnerOptions.Resolution
 	}
-	if r.Duration <= 0 {
+	if r.Duration < 0 {
```
ldemailly: this breaks the tests which relied on uninitialized duration to get the default duration

so let's use negative as the "infinite" value and we can change this if into == 0

periodic/periodic.go
```diff
@@ -238,13 +267,19 @@ func runOne(id int, funcTimes *stats.Histogram, sleepTimes *stats.Histogram, num
...
+				// This next line is tricky - such as for 2s duration and 1qps there is 1
+				// sleep of 2s between the 2 calls and for 3qps in 1sec 2 sleep of 1/2s etc
+				targetElapsedInSec = (float64(i) + float64(i)/float64(numCalls-1)) / perThreadQPS
+			} else {
+				// Calculate the target elapsed when in endless execution
+				targetElapsedInSec = float64(i) / perThreadQPS
+			}
 			targetElapsedDuration := time.Duration(int64(targetElapsedInSec * 1e9))
 			sleepDuration := targetElapsedDuration - elapsed
 			log.Debugf("%s target next dur %v - sleep %v", tIDStr, targetElapsedDuration, sleepDuration)
```
ldemailly, line 277: let's put that for the case we'd be sleeping (also let's ^C stop right away instead of after sleep) : 
```golang
			select {
			case <-c:
				break MainLoop
			case <-time.After(sleepDuration):
				// continue normal execution
			}
```

ymesika:
@ldemailly Thanks a lot for your detailed review and suggestions. They all make sense and will simplify code an launching the tool.

I will modify the code based on your feedback.

fortio_main.go
ymesika: I initially thought about the 0 value for infinite but then figured it's the same value as the default so tried a different flag.

Your suggestion below for a negative t for infinite is good alternative.

periodic/periodic.go
ymesika: @ldemailly Thanks for the -race tip!

periodic/periodic.go
ymesika: Although sync is no longer used due to use of switch on the channel instead of a go routine here is the load output for 5s run with -qps 0:

Before:

> Fortio v0.2.8 running at 0 queries per second, 8->8 procs, for 5s: http://localhost:8080
> Starting at max qps with 4 thread(s) [gomax 8] for 5s
> 13:59:23 I httprunner.go:79> Starting http test for http://localhost:8080 with 4 threads at 0.0 qps
> Ended after 5.000131225s : 206945 calls. qps=41388
> Aggregated Function Time : count 206945 avg 9.6066258e-05 +/- 2.04e-05 min 4.4506e-05 max 0.000975281 sum 19.8804318
> \# range, mid point, percentile, count
> < 0.001 , 0.001 , 100.00, 206945
> \# target 50% 0.000509894
> \# target 75% 0.000742587
> \# target 99% 0.000965973
> \# target 99.9% 0.00097435
> Code 200 : 206945
> Response Header Sizes : count 206945 avg 116 +/- 0 min 116 max 116 sum 24005620
> Response Body/Total Sizes : count 206945 avg 116 +/- 0 min 116 max 116 sum 24005620
> All done 206945 calls (plus 4 warmup) 0.096 ms avg, 41387.9 qps
> 13:59:28 I periodic.go:257> T003 ended after 5.000038522s : 51711 calls. qps=10342.120320168206
> 13:59:28 I periodic.go:257> T001 ended after 5.000097966s : 51763 calls. qps=10352.397163411897
> 13:59:28 I periodic.go:257> T002 ended after 5.000099538s : 51729 calls. qps=10345.59404405201
> 13:59:28 I periodic.go:257> T000 ended after 5.000101263s : 51742 calls. qps=10348.190422238655
> 

After  (after v0.3.0 commit):

> Fortio v0.3.0 running at 0 queries per second, 8->8 procs, for 5s: http://localhost:8080
> 13:58:10 I httprunner.go:79> Starting http test for http://localhost:8080 with 4 threads at 0.0 qps
> Starting at max qps with 4 thread(s) [gomax 8] for 5s
> Ended after 5.000094758s : 207429 calls. qps=41485
> Aggregated Function Time : count 207429 avg 9.5778898e-05 +/- 1.952e-05 min 4.632e-05 max 0.000439454 sum 19.867321
> \# range, mid point, percentile, count
> < 0.001 , 0.001 , 100.00, 207429
> \# target 50% 0.000242887
> \# target 75% 0.000341171
> \# target 99% 0.000435523
> \# target 99.9% 0.000439061
> Code 200 : 207429
> Response Header Sizes : count 207429 avg 116 +/- 0 min 116 max 116 sum 24061764
> Response Body/Total Sizes : count 207429 avg 116 +/- 0 min 116 max 116 sum 24061764
> All done 207429 calls (plus 4 warmup) 0.096 ms avg, 41485.0 qps
> 13:58:15 I periodic.go:296> T000 ended after 5.000007637s : 51894 calls. qps=10378.784147445092
> 13:58:15 I periodic.go:296> T002 ended after 5.000052854s : 51799 calls. qps=10359.69048978377
> 13:58:15 I periodic.go:296> T003 ended after 5.000057289s : 51867 calls. qps=10373.281145019297
> 13:58:15 I periodic.go:296> T001 ended after 5.000056364s : 51869 calls. qps=10373.683059545607
>

ldemailly:
thx! can you also do the CLA ?

ldemailly:
it’s great thanks, ideally you’d also add a test in periodic_test that indeed it stops all the run after 1 signal (and if you don’t mind the 0 -> -1 change for the flag only) and the CLA so I can merge this

ymesika:
Laurent, as an IBMer I probably need to be part of the IBM signed CLA. As the working days in Israel are Sun-Thu I can consult my colleagues about the right process only next week. Sorry for the delay.

linsun:
@ymesika thanks for the PR!  Yes, pls get in touch with @elevran or @vadimeisenbergibm on the CLA.   Any questions, Doug Davis is avail for help.

ldemailly (approved):
Thank you!

PR #7, opened 2017-11-03
adding test to docker cloud build

ldemailly:


Dockerfile.test
```diff
@@ -0,0 +1,5 @@
+# Getting the source tree ready for running tests (sut)
+FROM golang:1.8.3 as build
+WORKDIR /go/src/istio.io
+RUN go get google.golang.org/grpc
+COPY . fortio
```
ldemailly, line 5: Note: all those steps are identical to first phase of the regular Dockerfile so they are cached

PR #9, opened 2017-11-04
New HistogramData / Histogram.Export() that returns the data in consumable form + json export

### LINKED ISSUE #8: struct, json or other structured output for the histogram
right now fortio even as a library doesn't export the histogram data except through an io.Writer

which makes graphing and recording into db/other systems rely on script/parsing of the text output

ldemailly:
Plus a couple more Makefile targets for testing/installing
Changed Periodic return correspondingly

Also let's callers pick where the (text) output goes, used by fortio command line to send it all to stderr so `fortio load -json - ...` gets only the JSON to stdout

Fixes #8 

Example output:
```
{
"DurationHistogram": {
"Count": 40,
"Min": 0.054895218,
"Max": 0.153494493,
"Sum": 3.145985655,
"Avg": 0.078649641375,
"StdDev": 0.02462510369930899,
"Data": [
{
"Start": 0.054895218,
"End": 0.06,
"Percent": 15,
"Count": 6
},
{
"Start": 0.06,
"End": 0.07,
"Percent": 50,
"Count": 14
},
{
"Start": 0.07,
"End": 0.08,
"Percent": 77.5,
"Count": 11
},
{
"Start": 0.08,
"End": 0.09,
"Percent": 80,
"Count": 1
},
{
"Start": 0.1,
"End": 0.12,
"Percent": 92.5,
"Count": 5
},
{
"Start": 0.12,
"End": 0.14,
"Percent": 95,
"Count": 1
},
{
"Start": 0.14,
"End": 0.153494493,
"Percent": 100,
"Count": 2
}
],
"Percentiles": [
{
"Percentile": 50,
"Value": 0.07
},
{
"Percentile": 75,
"Value": 0.0790909090909091
},
{
"Percentile": 99,
"Value": 0.1507955944
},
{
"Percentile": 99.9,
"Value": 0.15322460314000003
}
]
},
"ActualQPS": 7.895627575902356,
"ActualDuration": 5066095078
}
```

```release-note
Fortio now supports exporting the data in programmatically consumable
form - and the command line can use -json path to export in json
```


fortio_main.go
```diff
@@ -150,12 +156,35 @@ func fortioLoad() {
...
+				log.Fatalf("Unable to create %s: %v", jsonFileName, err)
+			}
+		}
+		n, err := f.Write(j)
+		if err != nil {
+			log.Fatalf("Unable to write json to %s: %v", jsonFileName, err)
+		}
+		f.Write([]byte{'\n'})
+		fmt.Fprintf(out, "Succesfully wrote %d bytes of Json data to %s\n", n, jsonFileName)
+	}
```
ymesika: Shouldn't the json file be closed with `f.Close()`?
P.S. I understand that there is a preference in Go to avoid deferring the file close calls to be able to catch its return error.

fortio_main.go
```diff
@@ -150,12 +156,35 @@ func fortioLoad() {
...
+			f, err = os.Create(jsonFileName)
+			if err != nil {
+				log.Fatalf("Unable to create %s: %v", jsonFileName, err)
+			}
+		}
+		n, err := f.Write(j)
+		if err != nil {
+			log.Fatalf("Unable to write json to %s: %v", jsonFileName, err)
+		}
+		f.Write([]byte{'\n'})
```
ymesika: Error of this write call is not being checked

periodic/periodic.go
```diff
@@ -23,20 +23,23 @@
...
-	"strconv"
-	"strings"
 	"sync"
 	"time"
 
 	"istio.io/fortio/log"
 	"istio.io/fortio/stats"
 )
 
+const (
```
ymesika, line 38: Just wondering whether `fortio_main.go` is better place to define the version

Makefile
```diff
@@ -8,8 +9,18 @@ TAG:=$(USER)$(shell date +%y%m%d_%H%M%S)
...
-all: docker-version docker-push-internal
+# Local targets:
+install: test
+	go install ./...
+
+test:
+	go test -timeout 30s -race ./...
+
+lint:
+	gometalinter ./...
```
ymesika: Running `make lint` exits with an error while all I see are warnings:
> make: *** [lint] Error 1

Do you also get the same result?

periodic/periodic.go
ldemailly: the main isn't part of the library so that wouldn't work (circular dependencies) - I considered renaming the periodic package to fortio - not sure

Makefile
ldemailly: you probably need to run:
```
 go get -u gopkg.in/alecthomas/gometalinter.v1
```
let me add that to the file

ldemailly: or maybe you meant there are some errors, yes:
```
log/logger.go:166:6:warning: func name will be used as log.LogVf by other packages, and that stutters; consider calling this Vf (golint)
log/logger.go:196:6:warning: func name will be used as log.LogDebug by other packages, and that stutters; consider calling this Debug (golint)
log/logger.go:201:6:warning: func name will be used as log.LogVerbose by other packages, and that stutters; consider calling this Verbose (golint)
periodic/periodic.go:93:6:warning: type name will be used as periodic.PeriodicRunner by other packages, and that stutters; consider calling this Runner (golint)
```
going to update and make a pass at fixing once I find the flags

fortio_main.go
ldemailly: yes it's cosmetic last write - I wonder why the linter doesn't complain though
fixing it creatively - lmk what you think of upcoming update

fortio_main.go
ldemailly: it adds quite a lot of rather ugly code and unnecessary code (the program will exit and close open fds) but you're right, changing/adding

ymesika:
Looks good. I ran it locally with both stdout and json file and it works well.

This structured data is important for generating a visual histogram image and embedding it in the UI page.

Let me know if you want an additional pair of reviewer eyes or should I close it.

ldemailly:
@ymesika thanks ! yes #10 is a start of that and graphing (and DB recording) are next

you can mark the changes as approved if you're happy with them

ymesika (approved):
@ldemailly Maybe just increase the version for this change as it introduces new functionality.

ldemailly:
@ymesika  thx ! yeah version wise, as I haven't tagged yours yet I consider 0.3.0 to include both the ^C and this

@douglas-reid I'm going to merge this but if you have any comments I'll fix in followups

PR #10, opened 2017-11-05
Initial http UI for fortio

### LINKED ISSUE #11: fortio (web) UI to trigger tests from within the cluster, etc...
- allows remote trigger of load tests
- Json output option to programmatically gather the data

ldemailly:
~Stacked diff on top of #9 Please review that one first~ DONE

This adds a /fortio URI with very basic/early UI to trigger load test and get the results

Fixes #11


ldemailly:
It looks like this for now:
Initial menu:
screen shot 2017-11-04 at 8 52 50 pm
When running a test:
screen shot 2017-11-04 at 8 53 05 pm

ui/uihandler.go
```diff
@@ -0,0 +1,234 @@
...
+		if r.FormValue("json") == "on" {
+			JSONOnly = true
+			log.Infof("Starting JSON only load request from %v for %s", r.RemoteAddr, url)
+		} else {
+			log.Infof("Starting load request from %v for %s", r.RemoteAddr, url)
+		}
+	}
+	if !JSONOnly {
+		// Normal html mode
+		const templ = `<!DOCTYPE html><html><head><title>Φορτίο v{{.Version}} control UI</title></head>
```
ymesika, line 74: I wonder whether it makes more reason to have this template HTML in an external file and use template.ParseFiles() because once it will become a more complex HTML(s) it can be difficult to maintain it inline. In an external file you will also enjoy validators and syntax highlighting.

ui/uihandler.go
```diff
@@ -0,0 +1,234 @@
...
+	debugPath = debugpath
+	startTime = time.Now()
+	httpPort = port
+	if uiPath != "" {
+		http.HandleFunc(uiPath, Handler)
+		fmt.Printf("UI starting - visit:\nhttp://localhost:%d%s\n", port, uiPath)
+		logoPath = uiPath + "/logo.svg"
+		http.HandleFunc(logoPath, LogoHandler)
+	}
+	fhttp.Serve(port, debugpath)
```
ymesika, line 233: Do you think that using the same server / port for both the echo server and UI might affect the 'purity' of the stat results? E.g. if in the future you will show some real-time dynamic graph that fetches it from the server, won't it affect server response time to echo requests?

ui/uihandler.go
ldemailly: yes but I would need to embed it, I want to keep a single self contained binary (though in the image I could add data files but for “go get” ?)

ui/uihandler.go
ldemailly: that’s why there is a sub path for the UI, with the default config:

/debug is a fast debug
/fortio is this UI
everything else is echo server including / and / echo etc

ymesika (approved):

ui/uihandler.go
ymesika: I stumbled upon this utility to "pack" data files into go file: https://github.com/jteeuwen/go-bindata
It's a nice approach as I'm not aware of something similar in the golang itself.

PR #14, opened 2017-11-08
Initial version of integration with chartjs

### LINKED ISSUE #12: pick tiny, opensource, embeddable charting lib (for drawing the histograms)
Good source of info:

https://thenextweb.com/dd/2015/06/12/20-best-javascript-chart-libraries/

http://www.chartjs.org/ seems like a good candidate

@ymesika mentioned http://visjs.org/graph2d_examples.html too

### LINKED ISSUE #13: fortio charting/graphing

ldemailly:
Draw the histogram and cumulative %

Fixes #13
Fixes #12


ldemailly:
it looks like this:
screen shot 2017-11-08 at 11 34 43 am

ymesika (approved):
Tested with Chrome / Firefox and works well.
I still think the embedded HTML / JS is not the best way to handle web resources but it's a great start.

P.S. Tested it with duration -1s (for unlimited) but there is no way to stop it. In this case we may want to switch the start button into stop and keep it there. I can do it later on.

ui/chartjs.go
```diff
@@ -0,0 +1,14 @@
+// This file is https://github.com/chartjs/Chart.js/releases/download/v2.7.1/Chart.min.js
```
douglas-reid, line 1: curious: any reason not to just include this as `Chart.min.js`?  And serve it directly (via http.FileServer or similar)?

ldemailly: I want to be able to ‘go get istio.io/fortio‘  and have everything - do you know a standard way for go binaries to find the source tree that build them?
(That would solve the go get case but not the scp the binary around and use it)

ldemailly:
Btw you can already today interrupt a -1 run by opening the main page in another tab and clicking Exit but yes we can make that better, this was just the first version I got working

PR #17, opened 2017-11-11
Ability to fix X axis min,max and switch to log on both axis

ldemailly:
Ability to fix X axis max and switch to log on both axis

Also show the target url on page

(Needed to patch in
https://github.com/chartjs/Chart.js/pull/4198
In order to get https://github.com/chartjs/Chart.js/issues/4939
Axis type switching to work)


ymesika (approved):
Nice addition.
Works flawlessly on both Chrome/Firefox.

ldemailly:
Thx ! I test on chrome and safari mac myself

Added a custom build from my branch with all the fixes and removing the unused modules,
that shaved about 1/3rd of the minified size

https://github.com/ldemailly/Chart.js/commit/d71f4184bdd017a680a04c6168cd52a147724bac

ldemailly:
note that you need to ctrl or shift reload to get the .js otherwise log won't work

PR #18, opened 2017-11-14
Switch to relative path for logo and js library

ldemailly:
So the fortio can be used easily behind ingress paths


ldemailly:
thank ! @bochunz but this repo doesn't have bots, so you need to github approve :-)

bochunz (approved):

PR #21, opened 2017-11-14
Adding -curl flag and /fortio/fetch/... to fetch 1 url

ldemailly:
Not to be used as a proxy but convenient for debugging


fhttp/http.go
```diff
@@ -48,7 +48,7 @@ var (
 	// Host is treated specially, remember that one separately.
 	hostOverride string
 	// BufferSizeKb size of the buffer (max data) for optimized client in kilobytes defaults to 32k.
-	BufferSizeKb = 32
+	BufferSizeKb = 128
```
ldemailly, line 51: changed so fortio can proxy itself without error (proxy the shrunk chart.js which is about 122k)

andraxylia (approved):

PR #23, opened 2017-11-14
Separate static from dynamic

bochunz:
Move Chart JS and logo svg files to static dir.


ui/uihandler.go
```diff
@@ -472,13 +453,17 @@ func Serve(port int, debugpath string, uiPath string) {
...
-		chartJSPath = uiPath + chartjsURI
-		http.HandleFunc(chartJSPath, ChartJSHandler)
-		// Use relative paths after that (in the html template):
-		logoPath = "./" + logoURI
-		chartJSPath = "./" + chartjsURI
 	}
+
+	// Serve static contents in the ui/static dir.
+	// We use directory relative to this file to find the static contents, so no matter where is the working dir, static dir could be found.
+	_, filename, _, ok := runtime.Caller(0)
```
ldemailly: so file here is fully qualified source file at build time?
where would we place the files in docker ?
how is the content type set?

ui/uihandler.go
```diff
@@ -472,13 +453,17 @@ func Serve(port int, debugpath string, uiPath string) {
...
+	if ok {
+		fs := http.FileServer(http.Dir(path.Join(path.Dir(filename), "static")))
```
ldemailly: this would need to be relative

ldemailly (commented):
thx for the PR!
if we were to add a lot of static content this would definitely make sense (if we can resolve the issue/question below) but for just the textual svg logo and js I am not sure it’s worth it?

ui/uihandler.go
bochunz: Yes, and modified Dockerfile too, and verified it's working. Content type is automatically detected by file extension.

ui/uihandler.go
bochunz: Because the go binary can be anywhere, I feel putting the static files in a static location is easier - for both local dev or docker.

ui/static/js/Chart.min.js
```diff
@@ -0,0 +1,10 @@
+/*!
+ * Chart.js
+ * http://chartjs.org/
+ * Version: 2.7.1
```
ldemailly: this is not the same file/content - it needs to be the one I patched above

ui/uihandler.go
```diff
@@ -85,12 +83,12 @@ func Handler(w http.ResponseWriter, r *http.Request) {
 	if !JSONOnly {
 		// Normal html mode
 		const templ = `<!DOCTYPE html><html><head><title>Φορτίο v{{.Version}}</title>
-<script src="{{.ChartJSPath}}"></script>
+<script src="../static/js/Chart.min.js"></script>
```
ldemailly: let's keep the variable (set to the value)

ui/uihandler.go
```diff
@@ -85,12 +83,12 @@ func Handler(w http.ResponseWriter, r *http.Request) {
...
 </head>
 <body style="background: linear-gradient(to right, #d8aa20 , #c75228);">
-<img src="{{.LogoPath}}" alt="." height="69" width="45" align="right" />
-<img src="{{.LogoPath}}" alt="." height="92" width="60" align="right" />
-<img src="{{.LogoPath}}" alt="istio logo" height="123" width="80" align="right" />
+<img src="../static/img/logo.svg" alt="." height="69" width="45" align="right" />
+<img src="../static/img/logo.svg" alt="." height="92" width="60" align="right" />
+<img src="../static/img/logo.svg" alt="istio logo" height="123" width="80" align="right" />
```
ldemailly: same

ui/uihandler.go
```diff
@@ -456,15 +443,12 @@ func LogRequest(r *http.Request) {
...
-	w.Header().Set("Content-Type", "application/javascript; charset=utf-8")
-	w.Header().Set("Cache-Control", "max-age=365000000, immutable")
-	_, err := w.Write([]byte(chartjs))
-	if err != nil {
-		log.Errf("Error writing JS lib to %v: %v", r.RemoteAddr, err)
-	}
+// HTTP handler wrapper to add a Cache-Control header for static files.
+func AddCacheControl(h http.Handler) http.Handler {
+	return http.HandlerFunc (func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Cache-Control", "max-age=3600")
```
ldemailly: can you put the same cache-control I had, or why not ?

ui/uihandler.go
ldemailly: sorry I had missed the docker change

in which case runtime.Caller() would be not available ? is there a chance stripped binaries don't have the source code information for instance?

edit: I checked that strip for instance makes it fail...

ui/uihandler.go
ldemailly: I'm still concerned that one can't cp the binary around anymore if we go that way. I may get over that :-)
maybe we can make packages

speaking of which - there should probably be an alternate location, like
/usr/local/lib/fortio-0.3.6/static
for when we have packages

bochunz: We have to have statics delivered via some sort of packaging. Otherwise the page cannot scale. HTMLs need to be moved too. 

I'm thinking whether we should just have static HTML and JS. Then use the JSON API to retrieve data from browser

ui/uihandler.go
bochunz: sure, done

ui/static/js/Chart.min.js
bochunz: Oh, are you using some self-patched secret feature? :P

ldemailly: yes, not a secret though, the changes vs stock chart.js are described both in #17 and in the header of the current file

ui/uihandler.go
ldemailly: yes some ajaxy thing would be good, we should meet to talk about plans for that

but I'm still a bit attached to self contained binary ... working on getting over it if there is a benefit though

ui/static/js/Chart.min.js
bochunz: Updated

ldemailly:
can you make sure you add commits to your branch and never use push -f to a PR (so I can see the history/incremental changes)

bochunz:
Oh ok. Will do serial of commits next time

ui/uihandler.go
```diff
@@ -40,7 +42,7 @@ var (
 	// UI and Debug prefix/paths (read in ui handler).
 	uiPath      string // absolute (base)
 	logoPath    string // relative
-	chartJSPath string // relative
+  chartJSPath string // relative
```
ldemailly: make sure to run `go fmt` at save time with your editor (or run it manually)

bochunz: turns out go fmt doesn't automatically turn spaces to tabs. Done manually.

ldemailly:
can you rebase to latest and run  `make lint` and fix any issues it mentions
(you may need to run `make install-linters` the first time)
I was trying to get/try your pr to see and got errors

also please confirm that you can:
`make docker-internal TAG=test_23`
and then
` docker run -p 8080:8080 istio/fortio:test_23 server &`
and then control/shift-reload on http://localhost:8080/fortio/
(make sure your browser isn't caching the chart.js it needs to load)

bochunz:
Done with those checks :)

ldemailly:
you will need to modify the URLs in Webtest.sh to match new location of the logo and the js
and let's see it all pass in circleci

ui/uihandler.go
```diff
@@ -513,16 +501,24 @@ func Serve(port int, debugpath string, uipath string) {
...
 		http.HandleFunc(fetchPath, FetcherHandler)
 		fhttp.CheckConnectionClosedHeader = true // needed for proxy to avoid errors
 	}
+	logoPath = "../static/img/logo.svg"
+	chartJSPath = "../static/js/Chart.min.js"
+
+	// Serve static contents in the ui/static dir.
+	// We use directory relative to this file to find the static contents,
+	// so no matter where the generate go binary is, the static dir could be found.
+	_, filename, _, ok := runtime.Caller(0)
```
ldemailly: it can probably wait the next pr / when we make packages but we should have a variable for that directory and only fall back to caller if not set (we would set that variable when making a package)

ldemailly (approved):

ui/uihandler.go
```diff
@@ -513,16 +501,24 @@ func Serve(port int, debugpath string, uipath string) {
...
+	if ok {
+		fs := http.FileServer(http.Dir(path.Join(path.Dir(filename), "static")))
+		http.Handle("/static/", AddCacheControl(http.StripPrefix("/static/", fs)))
```
ldemailly: what does the strip prefix do?
does this fileserver logs requests ?(or can we make it do so)

ui/uihandler.go
```diff
@@ -513,16 +501,24 @@ func Serve(port int, debugpath string, uipath string) {
...
-		chartJSPath = uiPath + chartjsURI
-		http.HandleFunc(chartJSPath, ChartJSHandler)
-		// Use relative paths after that (in the html template):
-		logoPath = "./" + logoURI
-		chartJSPath = "./" + chartjsURI
+
 		fetchPath = uiPath + fetchURI
 		http.HandleFunc(fetchPath, FetcherHandler)
 		fhttp.CheckConnectionClosedHeader = true // needed for proxy to avoid errors
 	}
```
ldemailly: those must be inside the if

or rather: flip the if to return when uipath is empty

ldemailly (changes requested):
almost there, minor change request and Qs

ui/uihandler.go
ldemailly: from the doc it doesn't seem stripping the same prefix you expect is useful ?

let's leave all the resources rooted under uiPath (so by default `fortio/static/Chart.min.js` instead of ../)

ldemailly (commented):
@bochunz can we finish this PR so I can make some refactoring ? (ie make the minor changes I asked)

ui/uihandler.go
```diff
@@ -513,16 +501,24 @@ func Serve(port int, debugpath string, uipath string) {
...
+	logoPath = "../static/img/logo.svg"
```
ldemailly: `./static` instead of `../static` (as a side note I think `static/img/logo.svg` would also work as that's relative too, I forgot why I've put ./ in the current version)

bochunz: Statics are on "/static/js/...". The fortio page is on "/fortio" so we should use "../"

Did you mean putting statics under "/fortio/static/..."?

ldemailly: yeah I meant everything related to the ui under the uiPath (/fortio by default) prefix
but I can change it in subsequent PR too

ldemailly (approved):

PR #30, opened 2017-11-21
Adding Labels, URL and Run StartTime to JSON and Chart

ldemailly:
So the resulting json can be self contained/ready to store


andraxylia (approved):

bochunz (commented):
LGTM

PR #32, opened 2017-11-21
Add library for uploading results to Cloud Storage

bochunz:


storage/storage.go
```diff
@@ -0,0 +1,76 @@
...
+	"golang.org/x/net/context"
+)
+
+// DatastoreClient is a struct for interacting with Google Cloud Datastore
+type DatastoreClient struct {
+	client *datastore.Client
+}
+
+// IstioPerfResult is the data structure for uploading a test run result to Google Cloud Datastore
+type IstioPerfResult struct {
```
ldemailly, line 37: Can we use the (Http)RunnerResults directly somehow is there a default mapping where the datastore just stores every exported field (like json does) ?

storage/storage.go
```diff
@@ -0,0 +1,76 @@
...
+// (no std output by default)
+
+package storage // import "istio.io/fortio/storage"
+
+import (
+	"time"
+
+	"istio.io/fortio/log"
+
+	"cloud.google.com/go/datastore"
```
ldemailly, line 27: so far fortio has no external dependencies outside of grpc (and the grpc part is optional)

I wonder if we should put that gcp specific part somewhere else (istio/istio/tools/perf or istio/test-infra...)

or is there a way to just POST json directly?

bochunz: There is. But will be ugly in code (we need to somehow manually grab credentials, and manually construct a giant JSON).

bochunz: Can be done somewhat easier with a shell + python script though.

storage/storage.go
bochunz: I'd expect what we upload to be more or less different from what those internal data structures. The ResultJSON here should probably just be the serialization of RunnerResults. And we only put fields we explicitly want to index here.

ldemailly:
closing in favor of tsv export

PR #33, opened 2017-11-22
Fix such as -ui-path /otherpath/ works and support link time setting of data dir

ldemailly:
i.e. that static content is always under the UI path

Also simplified the code a bit

Updated the test to not need the curl image and use fortio -curl instead

### DIFF
--- a/ui/uihandler.go
+++ b/ui/uihandler.go
@@ -49,11 +49,35 @@ var (
 	httpPort int
 	// Start time of the UI Server (for uptime info).
 	startTime time.Time
+	// Directory where the static content and templates are to be loaded from.
+	// This is replaced at link time to the packaged directory (e.g /usr/local/lib/fortio/)
+	// but when fortio is installed with go get we use RunTime to find that directory.
+	// (see Dockerfile for how to set it)
+	dataDirectory string
 )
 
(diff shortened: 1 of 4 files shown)


ui/uihandler.go
```diff
@@ -502,40 +502,41 @@ func FetcherHandler(w http.ResponseWriter, r *http.Request) {
...
-
-		if servingPath != "" {
-			fs := http.FileServer(http.Dir(servingPath))
-			http.Handle("/fortio/static/", LogAndAddCacheControl(http.StripPrefix("/fortio", fs)))
-		}
-		fhttp.Serve(port, debugpath)
 	}
+	if staticPath != "" {
+		fs := http.FileServer(http.Dir(staticPath))
+		http.Handle(uiPath+"static/", LogAndAddCacheControl(http.StripPrefix(uiPath, fs)))
```
bochunz: use path.Join instead of bare string addition?

ldemailly: the first argument is a URI (I believe) so a string that works with http - always forward /
while Join() would create OS specific paths, for instance use \ on windows which would not be a valid URL

bochunz (approved):

PR #35, opened 2017-11-22
Move main page to templates directory, split JS into static and data

ldemailly:
Also changed/found while testing:
- in -curl mode, exit with non 0 for non 200 http code (like curl -f)
- check the base page in the test

### DIFF
--- a/ui/uihandler.go
+++ b/ui/uihandler.go
@@ -213,6 +213,13 @@ func Handler(w http.ResponseWriter, r *http.Request) {
 		res.DurationHistogram.Count,
 		1000.*res.DurationHistogram.Avg,
 		res.ActualQPS)))
+	ResultToJsData(w, res)
+	ResultToChart(w, res)
+	w.Write([]byte("</script></body></html>\n"))
+}
+
+// ResultToJsData converts a result object to chart data arrays.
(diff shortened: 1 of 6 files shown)


ui/uihandler.go
```diff
@@ -203,269 +138,139 @@ Use with caution, will interrupt/end this server: <input type="submit" name="exi
...
+	ResultToJsData(w, res)
+	ResultToChart(w, res)
+	w.Write([]byte("</script></body></html>\n"))
 }
-</script>
-</body></html>`))
-			}
+
+// ResultToJsData converts a result object to chart data arrays.
+func ResultToJsData(w io.Writer, res *fhttp.HTTPRunnerResults) {
```
bochunz, line 220: this looks hacky and hard to maintain. Could we just get a json string and do JSON.parse()? - such that we use go code to generate as less JS code as possible. This way we can just reuse the code from JSONOnly.

ldemailly: It's turning histogramdata -> chart.js compatible array of data points so I don't think it's that hacky or hard to maintain, _but_
yes, moving toward processing from JS the actual Jason result is the next step, to move the code to JS but for now it's still incrementally much much better than the previous version, no ?

ldemailly:
@bochunz ^

bochunz (approved):

PR #37, opened 2017-11-24
Move result -> chart logic entirely in javascript; fix thread safety; allow save/retrieve/graph saved data

### LINKED ISSUE #22: ui / server mode: make thread safe

### LINKED ISSUE #36: cache the linters
there are random failures and it takes a long while to run the linters on circleci

ldemailly:
So a page can directly graph a fortio JSON

Also added the version in the virtual static prefix (so fortio
upgrades/version bump yield a reload of the code)

Fixes #22 
Fixes #36 

Fixes bug in chunked mode


ldemailly:
(refactor4 would be to load json from url instead of writing it into the page)
EDIT: done in this PR

.circleci/config.yml
```diff
@@ -6,13 +6,14 @@ jobs:
   build:
     docker:
       # specify the version
-      - image: circleci/golang:1.8
+      - image: fortio/fortio.build:v4
```
douglas-reid, line 9: fortio now has its own build image? :stuck_out_tongue_closed_eyes: 

curious: how will we track golang versioning in this image?

ui/uihandler.go
```diff
@@ -144,18 +146,18 @@ func Handler(w http.ResponseWriter, r *http.Request) {
...
 	}
 	if DoExit {
-		syscall.Kill(syscall.Getpid(), syscall.SIGINT) // nolint: errcheck
+		syscall.Kill(syscall.Getpid(), syscall.SIGINT) // nolint: errcheck,gas
 		return
 	}
-	resolution, _ := strconv.ParseFloat(r.FormValue("r"), 64)
-	percList, _ := stats.ParsePercentiles(r.FormValue("p"))
-	qps, _ := strconv.ParseFloat(r.FormValue("qps"), 64)
+	resolution, _ := strconv.ParseFloat(r.FormValue("r"), 64) // nolint: gas
```
bochunz, line 152: what is this lint?

Makefile
```diff
@@ -4,29 +4,37 @@
...
-install-linters:
-	go get -u github.com/alecthomas/gometalinter
-	gometalinter -i -u
+# To debug linters, uncomment
+#DEBUG_LINTERS="--debug"
+
+local-lint:
+	gometalinter $(DEBUG_LINTERS) \
+	--deadline=180s --enable-all --aggregate \
+	--exclude=.pb.go --disable=gocyclo --line-length=132 $(LINT_PACKAGES)
```
douglas-reid, line 27: curious: is this `line-length` consistent with istio/istio ?

fhttp/httprunner_test.go
```diff
@@ -57,8 +56,32 @@ func TestHTTPRunner(t *testing.T) {
 	}
 }
 
+func TestHTTPRunnerClientRace(t *testing.T) {
+	http.HandleFunc("/echo1/", EchoHandler)
+	port := DynamicHTTPServer(false)
+	URL := fmt.Sprintf("http://localhost:%d/echo1/", port)
+
+	opts := HTTPRunnerOptions{
```
douglas-reid: non-blocking: it might be nice to have a func just for setting qps options. maybe: `WithQPS(qps int32)` or similar?

stats/stats.go
```diff
@@ -348,6 +348,7 @@ func (e *HistogramData) Print(out io.Writer, msg string) {
 		if i == len(e.Data)-1 {
 			sep = "<=" // last interval is inclusive (of max value)
 		}
+		// nolint: gas
```
douglas-reid, line 351: what does the `gas` linter do?  and do we want it, considering we've added a ton of `// nolint: gas` statements all over this codebase?

ui/uihandler.go
```diff
@@ -187,95 +189,110 @@ func Handler(w http.ResponseWriter, r *http.Request) {
...
-		}
-		// nolint: errcheck
-		w.Write([]byte(fmt.Sprintf("{x: %.12g, y: %.3f},\n", x, it.Percent)))
+// ResultToJsData converts a result object to chart data arrays and title
+// and creates a chart from the result object
+func ResultToJsData(w io.Writer, json []byte) {
+	// nolint: errcheck, gas
+	w.Write([]byte("var res = "))
+	// nolint: errcheck, gas
+	w.Write(json)
```
bochunz, line 233: Is JSON always convertible to JS code? Safest should be "JSON.parse('%s')" % json

fortio_main.go
```diff
@@ -32,6 +32,12 @@ import (
 	"istio.io/fortio/ui"
 )
 
+var httpOpts fhttp.HTTPOptions
+
+func init() {
+	httpOpts.Init("")
```
douglas-reid, line 38: any reason to prefer this over: `httpOpts = NewHTTPOptions("")` ?

bochunz (approved):
Let's merge this for agility and fix any problem (if any) in a later PR :)

fhttp/http.go
```diff
@@ -40,28 +40,32 @@ type Fetcher interface {
...
-	ResetHeaders()
-	extraHeaders.Add("User-Agent", userAgent)
+// NewHTTPOptions creates and initialize a HTTPOptions object.
+func NewHTTPOptions(url string) *HTTPOptions {
+	h := HTTPOptions{}
+	return h.Init(url)
+}
+
+// Init initializes the headers in an HTTPOptions (User-Agent).
+func (h *HTTPOptions) Init(url string) *HTTPOptions {
```
douglas-reid, line 63: does this need to be exposed?  can we just enforce calls through `NewHTTPOptions()`?

douglas-reid (commented):
Looks decent to me. A few small questions.

ui/uihandler.go
ldemailly: it's complaining about ignoring (the _) the possible error

.circleci/config.yml
ldemailly: the linters were being a pain so yes I made a build image
it's in the repo, it's currently using go1.8.5 (I have an issue open to evaluate switching to 1.9)

Makefile
ldemailly: it's shorter (I think mixer was 160 when I looked)

stats/stats.go
ldemailly: it usually finds the same stuff as errcheck + more... it's security oriented:
https://github.com/GoASTScanner/gas
which given fortio does take user input and turn into file and html may be useful

ui/uihandler.go
ldemailly: I though json was safe (and it works so far) - I tried to insert JSON.parse(' ... ') but that errors out with
Uncaught SyntaxError: Invalid or unexpected token

will leave as is for now

fortio_main.go
ldemailly: it avoids a copy, initializes in place the existing struct. but I could I guess make httpOpts a pointer

fhttp/http.go
ldemailly: it's to set / initialize options inside an existing variable/struct/memory

fhttp/httprunner_test.go
ldemailly: you're right - changing those to `opts.QPS = `

ldemailly:
changes since you last looked / ptal:

https://github.com/istio/fortio/compare/9f6d67400c3f6a771d8a3851152494256c749753...refactor3

Makefile
douglas-reid: i wonder if we should use a standard length across the effort (silly, i know, but might make it easier to hop repos).

stats/stats.go
douglas-reid: OK. Just seems to case more code harm than expected. Hopefully, it is catching useful things.

douglas-reid (commented):
/lgtm

PR #41, opened 2017-11-29
Allow in-page load of any results from a select box; Multi graph of latencies and percentiles

ldemailly:
While still also having single result link supported


ldemailly:
it looks like this (but is dynamic, you can use arrows or mouse to switch which is graphed)

screen shot 2017-11-29 at 12 06 39 pm

ldemailly:
You can now graph multiple runs against one another:

screen shot 2017-11-29 at 7 08 44 pm

(updated with p xx)

yutongz:
Thanks! It's awesome!

ui/templates/browse.html
```diff
@@ -6,13 +6,72 @@
 <img src="{{.LogoPath}}" alt="." height="92" width="60" align="right" />
 <img src="{{.LogoPath}}" alt="istio logo" height="123" width="80" align="right" />
 <h1>Φορτίο (fortio) v{{.Version}}</h1>
+<script src="{{.Version}}/static/js/fortio_chart.js"></script>
+<script>
```
bochunz, line 10: Feels like this could be another js file. But feel free to move it in later PRs.

bochunz (approved):

ui/templates/browse.html
ldemailly: 👍

PR #56, opened 2017-12-18
Add status probability and delay optional argument to the echo server

### LINKED ISSUE #50: add a /status/xxx and delay/microseconds endpoint
like http://httpbin.org/status/xxx (eg http://httpbin.org/status/418)

ldemailly:
http://localhost:8080/echo/?delay=1ms&status=567:33,568:33
For instance delays the response by 1ms and has 33% chance of returning
567, 33% 568 and 34% 200

Fixes #50


fhttp/http.go
```diff
@@ -750,9 +752,81 @@ var (
...
+// Format: status=503 for 100% 503s
+// status=503:20,404:10 for 20% 503s, 10% 404s, 70% 200s
+func parseStatus(status string) int {
+	lst := strings.Split(status, ",")
+	log.Debugf("Parsing status %s -> %v", status, lst)
+	// Simple non probabilistic status case:
+	if len(lst) == 1 && !strings.ContainsRune(status, ':') {
+		s, err := strconv.Atoi(status)
+		if err != nil {
+			log.Warnf("Bad input status %v, not a number nor coma and colon separated % list", status)
```
douglas-reid: nit: s/coma/comma

fhttp/http.go
```diff
@@ -750,9 +752,81 @@ var (
...
+			log.Warnf("Bad input status %v -> %v, not a number before colon", status, l2[0])
+			return http.StatusBadRequest
+		}
+		percStr := l2[1]
+		p, err := strconv.Atoi(percStr)
+		if err != nil || p < 0 || p > 100 {
+			log.Warnf("Percentage is not a [0 - 100] number in %v -> %v : %v %d", status, percStr, err, p)
+			return http.StatusBadRequest
+		}
+		if lastPercent+p > 100 {
```
douglas-reid: [thinking out loud] this seems fine. i wonder if it might be nice to autoscale to 100 down-the-road. like if say:

`?status:404:95,302:15,200:90`

we would adjust down to something like:

```
404 -> 48
302 -> 7
200 -> 45
```

That might be a bit surprising, but might help correct for silly math blunders too...

fhttp/http.go
```diff
@@ -750,9 +752,81 @@ var (
...
+// Format: status=503 for 100% 503s
+// status=503:20,404:10 for 20% 503s, 10% 404s, 70% 200s
+func parseStatus(status string) int {
+	lst := strings.Split(status, ",")
+	log.Debugf("Parsing status %s -> %v", status, lst)
+	// Simple non probabilistic status case:
+	if len(lst) == 1 && !strings.ContainsRune(status, ':') {
+		s, err := strconv.Atoi(status)
+		if err != nil {
+			log.Warnf("Bad input status %v, not a number nor coma and colon separated %% list", status)
```
douglas-reid: nit: s/coma/comma

fhttp/http.go
```diff
@@ -750,9 +752,81 @@ var (
...
+			log.Warnf("Sum of percentage is greater than 100 in %v %d %d", status, lastPercent, p)
+			return http.StatusBadRequest
+		}
+		for i := 0; i < p; i++ {
+			probTable[lastPercent+i] = s
+		}
+		lastPercent += p
+	}
+	res := rand.Intn(100)
+	log.Debugf("[0-100[ for %s roll got %d -> %d", status, res, probTable[res])
```
douglas-reid: do we need a full 100 entry array for this?

it seems like we could track this with a map of `len(lst)` with additive indices. we expect the number of status entries to be small and so running through the list each time is probably efficient enough.  this would also allow arbitrary precision on the percentages.

something like:
```
// for ?status=503:20,404:10

table := map[int]int{ 20:503, 30:404 }
res := rand.Intn(100)
for percent, code := range table {
  if res < percent {
     return code
  }
}
return http.StatusOK
```

not sure this matters all that much, but thought I'd ask.

fhttp/http_test.go
```diff
@@ -248,6 +248,57 @@ func TestDebugSummary(t *testing.T) {
...
+		}
+	}
+}
+
+// Round down to the nearest hundred
+func roundhundred(x int) int {
+	return int(float64(x)+50.) / 100
+}
+
+func TestParseStatus2(t *testing.T) {
```
douglas-reid: nit: maybe call TestStatusDistribution ?  that would help immediately clarify the intent.

fhttp/http.go
douglas-reid: or, maybe we can do this as we parse status codes and avoid any secondary storage altogether?  generate the number first, walk through as we go... (haven't thought through all the error conditions in this approach, however).

douglas-reid (approved):
/lgtm

overkill is the best kill.

PR #57, opened 2017-12-18
Calculate % for each retcode and allow initial errors

ldemailly:
Also handle 100% error case in graph title

(Stacked diff on top of #56)

### DIFF
--- a/fhttp/httprunner.go
+++ b/fhttp/httprunner.go
@@ -61,8 +61,9 @@ func (httpstate *HTTPRunnerResults) Run(t int) {
 // options.
 type HTTPRunnerOptions struct {
 	periodic.RunnerOptions
-	HTTPOptions        // Need to call Init() to initialize
-	Profiler    string // file to save profiles to. defaults to no profiling
+	HTTPOptions               // Need to call Init() to initialize
+	Profiler           string // file to save profiles to. defaults to no profiling
+	AllowInitialErrors bool   // whether initial errors don't cause an abort
 }
 
 // RunHTTPTest runs an http test and returns the aggregated stats.
@@ -87,7 +88,7 @@ func RunHTTPTest(o *HTTPRunnerOptions) (*HTTPRunnerResults, error) {
 			return nil, fmt.Errorf("unable to create client %d for %s", i, o.URL)
 		}
 		code, data, headerSize := httpstate[i].client.Fetch()
-		if code != http.StatusOK {
+		if !o.AllowInitialErrors && code != http.StatusOK {
 			return nil, fmt.Errorf("error %d for %s: %q", code, o.URL, string(data))
 		}
 		if i == 0 && log.LogVerbose() {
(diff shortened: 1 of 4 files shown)


ldemailly:
with `http://localhost:8080/echo?status=503:10,504:20`
```
Code 200 : 2060 (68.7 %)
Code 503 : 318 (10.6 %)
Code 504 : 622 (20.7 %)
```

ui/static/js/fortio_chart.js
```diff
@@ -99,7 +99,11 @@ function makeTitle(res) {
     var total = res.DurationHistogram.Count
     var errStr = "no error"
     if (httpOk != total) {
-      errStr = myRound(100.*(total-httpOk)/total, 2) + '% errors'
+      if (httpOk) {
+        errStr = myRound(100.*(total-httpOk)/total, 2) + '% errors'
+      } else {
+        errStr = "100% errors!"
```
ldemailly, line 105: was showing NaN% without this fix

fhttp/httprunner.go
```diff
@@ -61,8 +61,9 @@ func (httpstate *HTTPRunnerResults) Run(t int) {
 // options.
 type HTTPRunnerOptions struct {
 	periodic.RunnerOptions
-	HTTPOptions        // Need to call Init() to initialize
-	Profiler    string // file to save profiles to. defaults to no profiling
+	HTTPOptions               // Need to call Init() to initialize
+	Profiler           string // file to save profiles to. defaults to no profiling
+	AllowInitialErrors bool   // whether initial errors don't cause an abort
```
douglas-reid, line 66: curious: what do initial errors signify?

ldemailly: like a 503 during the first fetch, the previous (and now still default cmd line) behavior is to abort on non 200

the UI will now not abort on error - that way you can use say `&status=503:40` to get 40% 503s and not risk having it stop right away if one of the initial requests hit that probability

Do you think there is a way I can clarify the description/comment ?

douglas-reid: This more-or-less aligns with my understanding. It seems like if any of the request number of threads fails an initial request, fortio will fail altogether. Is that a correct regurgitation?

Do we want to default to allowing connection errors?  Is the concern repeatedly failing connections (and a need to fail-fast)?

In other words, should this be `FailFast` instead of `AllowErrors` ?  I don't know that it really matters, just curious if you ever intend to run it with this option turned off.

ldemailly: I guess the reason for the failfast (good name! thx!) current/past behavior is to quickly  and effectively catch a typo in a url (404) or an error in setup quickly (service doesn't exist)

if you start a 1h run to have it scroll with errors for 1h it's usually not what you want

with istio though it seems failfast means it'd give up too early :-) so maybe I should change the default

douglas-reid: ha. no that's ok. if running for an hour, we would want to catch a silly error first. 👍

douglas-reid (approved):
/lgtm

PR #61, opened 2017-12-22
Merging histograms on different scale parameters

olimpias:
Merging histograms on different scale parameters has become possible and a unit test is added.
In makefile pwd is changed with shell pwd

### DIFF
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -177,6 +177,10 @@ type HistogramData struct {
 func NewHistogram(Offset float64, Divider float64) *Histogram {
 	h := new(Histogram)
 	h.Offset = Offset
+	if Divider == 0 {
+		log.Fatalf("Divider can not be %f",Divider)
+		return nil
+	}
 	h.Divider = Divider
 	h.Hdata = make([]int32, numBuckets+1)
 	return h
@@ -206,14 +210,7 @@ func init() {
 func (h *Histogram) Record(v float64) {
 	h.Counter.Record(v)
 	// Scaled value to bucketize:
-	scaledVal := (v - h.Offset) / h.Divider
-	idx := 0
-	if scaledVal >= lastValue {
-		idx = numBuckets
-	} else if scaledVal >= firstValue {
-		idx = val2Bucket[int(scaledVal)]
-	} // else it's <  and idx 0
-	h.Hdata[idx]++
+	h.appendRecordToHistogram(v)
 }
 
 // CalcPercentile returns the value for an input percentile
@@ -393,21 +390,56 @@ func (h *Histogram) Clone() *Histogram {
 // CopyFrom sets the content of this object to a copy of the src.
 func (h *Histogram) CopyFrom(src *Histogram) {
 	h.Counter = src.Counter
-	// we don't copy offset/divider as this assumes compatible src/dest
-	for i := 0; i < len(h.Hdata); i++ {
-		h.Hdata[i] += src.Hdata[i]
+	h.CopyHDataFrom(src)
+}
+
+func (h *Histogram) CopyHDataFrom(src * Histogram)  {
+	if h.Divider == src.Divider && h.Offset == src.Offset {
+		for i := 0; i < len(h.Hdata); i++ {
+			h.Hdata[i] += src.Hdata[i]
+		}
+		return;
+	}
+	//Scale values are not equal for histograms.
+	//Need to receive record values from src and add them to h according to h's offset and divider
+	var scaledValue float64
+	for i := 0; i < len(src.Hdata); i++ {
+		if src.Hdata[i] == 0 {
+			continue
+		}
+		scaledValue = -1
+		//Find scaleValue in val2Bucket
+		for j := 0; j<numBuckets; j++ {
+			if val2Bucket[j] == i {
+				scaledValue = float64(j)
+			}
+		}
+		if scaledValue == -1{
+			continue
+		}
+		v := scaledValue * src.Divider + src.Offset
+		for src.Hdata[i] != 0{
+			h.appendRecordToHistogram(v)
+			src.Hdata[i] -= 1
+		}
 	}
 }
 
+func (h*Histogram) appendRecordToHistogram(v float64)  {
+	// Scaled value to bucketize:
+	scaledVal := (v - h.Offset) / h.Divider
+	idx := 0
+	if scaledVal >= lastValue {
+		idx = numBuckets
+	} else if scaledVal >= firstValue {
+		idx = val2Bucket[int(scaledVal)]
+	} // else it's <  and idx 0
+	h.Hdata[idx]++
+}
+
+
 // Transfer merges the data from src into this Histogram and clears src.
 func (h *Histogram) Transfer(src *Histogram) {
-	// TODO potentially merge despite different offset/scale
-	if src.Offset != h.Offset {
-		log.Fatalf("Incompatible offsets in Histogram Transfer %f %f", src.Offset, h.Offset)
-	}
-	if src.Divider != h.Divider {
-		log.Fatalf("Incompatible scale in Histogram Transfer %f %f", src.Divider, h.Divider)
-	}
 	if src.Count == 0 {
 		return
 	}
@@ -417,9 +449,7 @@ func (h *Histogram) Transfer(src *Histogram) {
 		return
 	}
 	h.Counter.Transfer(&src.Counter)
-	for i := 0; i < len(h.Hdata); i++ {
-		h.Hdata[i] += src.Hdata[i]
-	}
+	h.CopyHDataFrom(src)
 	src.Reset()
 }
 
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -178,7 +178,7 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
 	h := new(Histogram)
 	h.Offset = Offset
 	if Divider == 0 {
-		log.Fatalf("Divider can not be %f",Divider)
+		log.Fatalf("Divider can not be zero")
 		return nil
 	}
 	h.Divider = Divider
@@ -209,8 +209,19 @@ func init() {
 // Record records a data point.
 func (h *Histogram) Record(v float64) {
 	h.Counter.Record(v)
-	// Scaled value to bucketize:
 	h.appendRecordToHistogram(v)
+}
+
+func (h *Histogram) appendRecordToHistogram(v float64) {
+	// Scaled value to bucketize:
+	scaledVal := (v - h.Offset) / h.Divider
+	idx := 0
+	if scaledVal >= lastValue {
+		idx = numBuckets
+	} else if scaledVal >= firstValue {
+		idx = val2Bucket[int(scaledVal)]
+	} // else it's <  and idx 0
+	h.Hdata[idx]++
 }
 
 // CalcPercentile returns the value for an input percentile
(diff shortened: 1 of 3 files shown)


stats/stats.go
```diff
@@ -177,6 +177,10 @@ type HistogramData struct {
 func NewHistogram(Offset float64, Divider float64) *Histogram {
 	h := new(Histogram)
 	h.Offset = Offset
+	if Divider == 0 {
+		log.Fatalf("Divider can not be %f",Divider)
```
ldemailly: might as well say "Divider can not be zero"

stats/stats_test.go
```diff
@@ -380,8 +380,51 @@ func TestHistogramNegativeNumbers(t *testing.T) {
...
+# target 75% 58.25
+h2 should now be empty : no data
+`
+	if actual != expected {
+		t.Errorf("unexpected:\n%s\tvs:\n%s", actual, expected)
+	}
+}
+
 func TestTransferHistogram(t *testing.T) {
-	tP := []float64{100.} // TODO: use 75 and fix bug
```
ldemailly: the TODO may not have been very clear but when using 75% instead of 100% some results didn't match my expectations - basically when using 75% I expect the value given to be == the 3rd point / 4 points

stats/stats_test.go
```diff
@@ -415,19 +458,19 @@ func TestTransferHistogram(t *testing.T) {
...
-# target 100% 90
+# target 75% 90
 merged h2 -> h1 : count 4 avg 50 +/- 35.36 min 10 max 90 sum 200
 # range, mid point, percentile, count
 >= 10 < 20 , 15 , 25.00, 1
 >= 20 < 30 , 25 , 50.00, 1
 >= 80 < 90 , 85 , 75.00, 1
 >= 90 <= 90 , 90 , 100.00, 1
-# target 100% 90
+# target 75% 90
```
ldemailly: this one should be 80 not 90

stats/stats_test.go
```diff
@@ -436,22 +479,22 @@ merged h1a -> h2a : count 5 avg 50 +/- 31.62 min 10 max 90 sum 250
 >= 50 < 60 , 55 , 60.00, 1
 >= 80 < 90 , 85 , 80.00, 1
 >= 90 <= 90 , 90 , 100.00, 1
-# target 100% 90
+# target 75% 87.5
```
ldemailly: this one should be between 50 and 80

stats/stats_test.go
```diff
@@ -436,22 +479,22 @@ merged h1a -> h2a : count 5 avg 50 +/- 31.62 min 10 max 90 sum 250
...
 h1 should now be empty : no data
 h3 after merge - 1 : count 4 avg 50 +/- 35.36 min 10 max 90 sum 200
 # range, mid point, percentile, count
 >= 10 < 20 , 15 , 25.00, 1
 >= 20 < 30 , 25 , 50.00, 1
 >= 80 < 90 , 85 , 75.00, 1
 >= 90 <= 90 , 90 , 100.00, 1
-# target 100% 90
+# target 75% 90
```
ldemailly: should be 80

stats/stats_test.go
```diff
@@ -436,22 +479,22 @@ merged h1a -> h2a : count 5 avg 50 +/- 31.62 min 10 max 90 sum 250
...
 h3 after merge - 2 : count 4 avg 50 +/- 35.36 min 10 max 90 sum 200
 # range, mid point, percentile, count
 >= 10 < 20 , 15 , 25.00, 1
 >= 20 < 30 , 25 , 50.00, 1
 >= 80 < 90 , 85 , 75.00, 1
 >= 90 <= 90 , 90 , 100.00, 1
-# target 100% 90
+# target 75% 90
```
ldemailly: should be 80

stats/stats.go
```diff
@@ -393,21 +390,56 @@ func (h *Histogram) Clone() *Histogram {
...
+		}
+		v := scaledValue * src.Divider + src.Offset
+		for src.Hdata[i] != 0{
+			h.appendRecordToHistogram(v)
+			src.Hdata[i] -= 1
+		}
+	}
+}
+
+func (h*Histogram) appendRecordToHistogram(v float64)  {
```
ldemailly: can you move that function to line 215, it would minimize the diff

stats/stats.go
```diff
@@ -393,21 +390,56 @@ func (h *Histogram) Clone() *Histogram {
 // CopyFrom sets the content of this object to a copy of the src.
 func (h *Histogram) CopyFrom(src *Histogram) {
 	h.Counter = src.Counter
-	// we don't copy offset/divider as this assumes compatible src/dest
-	for i := 0; i < len(h.Hdata); i++ {
-		h.Hdata[i] += src.Hdata[i]
+	h.CopyHDataFrom(src)
+}
+
+func (h *Histogram) CopyHDataFrom(src * Histogram)  {
```
ldemailly: can you add a comment for the method, including that in your current implementation the destination histogram "wins" 
(maybe we can add later some other method that tries to calculate the best/most compatible offset/divider pair for the 2 or N input histograms)

stats/stats.go
```diff
@@ -393,21 +390,56 @@ func (h *Histogram) Clone() *Histogram {
...
+	if h.Divider == src.Divider && h.Offset == src.Offset {
+		for i := 0; i < len(h.Hdata); i++ {
+			h.Hdata[i] += src.Hdata[i]
+		}
+		return;
 	}
+	//Scale values are not equal for histograms.
+	//Need to receive record values from src and add them to h according to h's offset and divider
+	var scaledValue float64
+	for i := 0; i < len(src.Hdata); i++ {
```
ldemailly: this may be easier and more correct if you use Export() on src first

or make a function on how to return the interval for each bucket that both export and this can use

you want to use at least the mid point of the range as the value (though it's not quite the best that can be done when re-bining) there are some subtleties around min and max too (ie if 1 bucket is [min, x] and has a count of 1 the value to use is min, not (min+x)/2))

ldemailly (changes requested):
Thanks a lot for the PR!

Please see the test (formatting) failures and fix those as well as comments below:

stats/stats.go
olimpias: Okay I added

stats/stats.go
olimpias: As you mention I used Export method on src to receive histogramData. And I processed its Datas with formula that you mention above.

stats/stats.go
olimpias: Sure

stats/stats_test.go
olimpias: Could bug be on CalcPercentile method in stats.go? because When I ran TestHistogram test, CalcPercentile method calculates 775 for 75 percentile however 775 is in range of 80 percentile. As you correct below, I have expected to be min value near target percentile.

stats/stats.go
olimpias: sure

stats/stats_test.go
olimpias: If you approve that there is a bug on CalcPercentile method, I made solution, I would like to commit. Also, merry Christmas @ldemailly :)

ldemailly:
Did you forget to push your updates? I don’t see them

Merry Christmas

olimpias:
Merry Christmas @ldemailly,  I am waiting for your comment about  CalcPercentile method in stats.go.

stats/stats_test.go
ldemailly: yes I expect there is a bug ("fix bug" part) though maybe the bug fix can be a different PR for just that todo/bug fix if you don't spot the bug easily

btw those tests relying on the text output are old, written before there was an Export, new test can be more easily written against that or directly calling CalcPercentile

olimpias:
Btw I had a problem for running make lint in sudo mode. I solve problem by changing $(PWD) to $(CURDIR) in MakeFile. If you wish, I could PR it too. Also according to https://sourceforge.net/p/ipt-netflow/bugs-requests-patches/53/ $(CURDIR) guarantees to give relative path.

ldemailly:
> Btw I had a problem for running make lint in sudo mode. I solve problem by changing $(PWD) to $(CURDIR) in MakeFile. If you wish, I could PR it too. Also according to https://sourceforge.net/p/ipt-netflow/bugs-requests-patches/53/ $(CURDIR) guarantees to give relative path.

you shouldn't need sudo for any command (you may need your user to be in the docker group though) but sure (though I think the current makefile isn't gnu make specific - I guess we could use $(shell pwd) too if $(PWD) isn't always set

olimpias:
I just realized that I forgot to reverse todo about changing percentile to 75. in test class. Now, I pushed. Also I think we should change $(PWD) to $(shell pwd), $(shell pwd) worked in sudo.

stats/stats.go
```diff
@@ -393,21 +401,35 @@ func (h *Histogram) Clone() *Histogram {
...
+		for i := 0; i < len(h.Hdata); i++ {
+			h.Hdata[i] += src.Hdata[i]
+		}
+		return
+	}
+
+	hData := src.Export([]float64{})
+	for _, v := range hData.Data {
+		if v.Count == 1 {
+			h.appendRecordToHistogram(v.Start)
```
ldemailly: that's not correct, if the value is neither min nor max, there could be a single count in the bucket and the best value is still the midpoint, not the first one

it'd be good to preserve min and max (but if you don't the loop below is all you need)

stats/stats_test.go
```diff
@@ -380,6 +380,49 @@ func TestHistogramNegativeNumbers(t *testing.T) {
...
+func TestTransferHistogramWithDifferentScales(t *testing.T) {
+	tP := []float64{75.}
+	var b bytes.Buffer
+	w := bufio.NewWriter(&b)
+	h1 := NewHistogram(2, 15)
+	h1.Record(30)
+	h1.Record(40)
+	h1.Record(50)
+	h2 := NewHistogram(0, 10)
+	h2.Record(20)
```
ldemailly, line 392: add another point in same bucket for h2 so you see the bug that you only added 1 count (do it before fixing the bug so you know you have a test before the fix)

stats/stats.go
```diff
@@ -393,21 +401,29 @@ func (h *Histogram) Clone() *Histogram {
...
+	if h.Divider == src.Divider && h.Offset == src.Offset {
+		for i := 0; i < len(h.Hdata); i++ {
+			h.Hdata[i] += src.Hdata[i]
+		}
+		return
+	}
+
+	hData := src.Export([]float64{})
+	for _, data := range hData.Data {
+		v := float64(data.Count) * (data.Start + data.End) / 2
```
ldemailly: this is wrong

stats/stats.go
```diff
@@ -393,21 +401,29 @@ func (h *Histogram) Clone() *Histogram {
...
-	// we don't copy offset/divider as this assumes compatible src/dest
-	for i := 0; i < len(h.Hdata); i++ {
-		h.Hdata[i] += src.Hdata[i]
+	h.CopyHDataFrom(src)
+}
+
+// CopyHDataFrom appends histogram data values to this object from the src.
+// Src histogram data values will be appended according to this object's
+// offset and divider
+func (h *Histogram) CopyHDataFrom(src *Histogram) {
```
ldemailly: let's not expose this, rename to `copyHDataFrom()`

stats/stats.go
```diff
@@ -205,6 +209,10 @@ func init() {
 // Record records a data point.
 func (h *Histogram) Record(v float64) {
 	h.Counter.Record(v)
+	h.appendRecordToHistogram(v)
+}
+
+func (h *Histogram) appendRecordToHistogram(v float64) {
```
ldemailly: let's call this lowercase `record()` it's consistent with what it does - you could consider adding a count param to avoid a loop in the other one

ldemailly (changes requested):

stats/stats.go
olimpias: I hope, this time it will be okay :)

ldemailly (approved):
thanks!

olimpias:
Thank you for your comments @ldemailly

PR #64, opened 2017-12-27
interrupt properly each run through channel (and side effect fix compilation error and works on windows)

### LINKED ISSUE #63: don't rely on syscall.Kill to interrupt a run / make fortio works on Windows
I just discovered fortio doesn't run on windows:

```
C:\Users\l\go\src\istio.io>go get -u istio.io/fortio
# istio.io/fortio/ui
fortio\ui\uihandler.go:181:3: undefined: syscall.Kill
```

Ignoring for a second that go should provide a kill emulation (imo) - we probably should cleanup fortio anyway as it's only to interrupt itself

(we can also remove the exit option from the UI and only interrupt a given run or maintain a stack or list of them - or interrupt all of them through a shared channel)

ldemailly:
fixes #63

### DIFF
--- a/periodic/periodic.go
+++ b/periodic/periodic.go
@@ -130,6 +130,11 @@ type periodicRunner struct {
 	RunnerOptions
 }
 
+var (
+	globalAbort chan os.Signal
+	abortMutex  sync.Mutex
+)
+
 // Normalize initializes and normalizes the runner options. In particular it sets
 // up the channel that can be used to interrupt the run later.
 func (r *RunnerOptions) Normalize() {
@@ -163,15 +168,39 @@ func (r *RunnerOptions) Normalize() {
 	}
 	if r.Stop == nil {
 		r.Stop = make(chan struct{}, 1)
-		// Catch SIGINT signals from the OS and close the channel to terminate the run loops
-		c := make(chan os.Signal, 1)
-		signal.Notify(c, os.Interrupt)
 		// TODO: for long running server this slowly leaks go rountines... do better
 		go func() {
-			// in case the signal happens after r.Stop was already closed
-			defer recover() // nolint: errcheck
-			<-c
-			close(r.Stop)
+			abortMutex.Lock()
+			runnerChan := r.Stop
+			if globalAbort == nil {
+				log.Infof("First outstanding run starting, catching signal")
+				globalAbort = make(chan os.Signal, 1)
+				signal.Notify(globalAbort, os.Interrupt)
+			}
+			abortMutex.Unlock()
+			log.LogVf("WATCHER starting new watcher for signal! %d", runtime.NumGoroutine())
+			select {
+			case _, ok := <-globalAbort:
+				log.LogVf("WATCHER Got interrupt signal! %v", ok)
+				abortMutex.Lock()
+				if r.Stop != nil {
+					close(r.Stop)
+					r.Stop = nil
+				}
+				if ok {
+					log.Infof("Closing to notify all and resetting signal handler")
+					if globalAbort != nil {
+						close(globalAbort)
+						globalAbort = nil
+					}
+					signal.Reset(os.Interrupt)
+				}
+				abortMutex.Unlock()
+			case <-runnerChan:
+				log.LogVf("WATCHER r.Stop readable")
+				// nothing to do, stop happened
+			}
+			log.LogVf("WATCHER End of go routine")
 		}()
 	}
 }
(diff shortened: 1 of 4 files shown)


ymesika:
@ldemailly I'm using Mac and can't verify that it indeed fixed the compilation error on Windows

ldemailly:
@ymesika welcome back ! it's not as much about windows than about per run channel now, PTAL

ui/uihandler.go
```diff
@@ -101,10 +104,11 @@ func (w *HTMLEscapeWriter) Write(p []byte) (int, error) {
 // Handler is the main UI handler creating the web forms and processing them.
 func Handler(w http.ResponseWriter, r *http.Request) {
 	LogRequest(r, "UI")
-	DoExit := false
-	if r.FormValue("exit") == "Exit" {
-		log.Critf("Exit request from %v", r.RemoteAddr)
-		DoExit = true
+	DoStop := false
```
ymesika: Seems like you can't actually get a form with both 'stop' and 'load' set together? In this case maybe DoStop and DoLoad can be combined just to hold request type.

ui/uihandler.go
```diff
@@ -231,6 +264,9 @@ func Handler(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			log.Errf("Unable to write json output for %v: %v", r.RemoteAddr, err)
 		}
+		if DoSave {
```
ymesika: Adding one more type to the request type variable I mentioned in my previous comment - stop / load / save

ymesika (approved):
Beside my comments below, which are cosmetics, I tested this new button and it works fine.
Wasn't able to check if it indeed got rid of the compilation error on Windows.

periodic/periodic.go
```diff
@@ -59,14 +59,14 @@ type Runnable interface {
 
 // MakeRunners creates an array of NumThreads identical Runnable instances.
 // (for the (rare/test) cases where there is no unique state needed)
-func (ro *RunnerOptions) MakeRunners(r Runnable) {
-	log.Infof("Making %d clone of %+v", ro.NumThreads, r)
-	if len(ro.Runners) < ro.NumThreads {
-		log.Infof("Resizing runners from %d to %d", len(ro.Runners), ro.NumThreads)
-		ro.Runners = make([]Runnable, ro.NumThreads)
+func (r *RunnerOptions) MakeRunners(rr Runnable) {
```
ldemailly, line 62: this change is to minimize the diff for Normalize (as go linters want the variable name for receivers to be all the same)

ui/uihandler.go
ldemailly: well not save as save is that checkbox is in addition to load / json

ui/uihandler.go
ldemailly: changed to enum, thanks!

ui/uihandler.go
```diff
@@ -155,98 +194,106 @@ func Handler(w http.ResponseWriter, r *http.Request) {
...
+			for _, v := range runs {
+				v.Abort()
+				i++
+			}
+			uiRunMapMutex.Unlock()
+			log.Infof("Interrupted %d runs", i)
+		} else { // Stop one
+			uiRunMapMutex.Lock()
+			v := runs[runid]
+			if v != nil {
```
rkpagadala: Change to
v, ok := runs[runid]
 			if ok {

periodic/periodic.go
```diff
@@ -128,9 +132,15 @@ type periodicRunner struct {
 	RunnerOptions
 }
 
-// internal version, returning the concrete implementation.
-func newPeriodicRunner(opts *RunnerOptions) *periodicRunner {
-	r := &periodicRunner{*opts} // by default just copy the input params
+var (
+	globalAbort     chan os.Signal
```
rkpagadala: Consider changing to
gAbortChan

periodic/periodic.go
```diff
@@ -128,9 +132,15 @@ type periodicRunner struct {
...
+	outstandingRuns int64
```
rkpagadala: gOutstandingRuns

periodic/periodic.go
```diff
@@ -128,9 +132,15 @@ type periodicRunner struct {
...
+	abortMutex      sync.Mutex
```
rkpagadala: gAbortMutex

Or put then all in a structure and call it gSomething

rkpagadala (approved):
Minor comments

periodic/periodic.go
ldemailly: k

periodic/periodic.go
ldemailly: k

periodic/periodic.go
ldemailly: k

ui/uihandler.go
ldemailly: k

PR #65, opened 2017-12-29
Configurable timeout value - merge histograms in  different scale

olimpias:
Configurable timeout value set is added and merging different histogram is done

### DIFF
--- a/fhttp/http.go
+++ b/fhttp/http.go
@@ -66,8 +66,8 @@ func NewHTTPOptions(url string) *HTTPOptions {
 func (h *HTTPOptions) Init(url string) *HTTPOptions {
 	h.URL = url
 	h.NumConnections = 1
-	if h.Timeout == 0 {
-		h.Timeout = DefaultTimeOutValue
+	if h.HTTPReqTimeOut == 0 {
+		h.HTTPReqTimeOut = HTTPReqTimeOutDefaultValue
 	}
 	h.ResetHeaders()
 	h.extraHeaders.Add("User-Agent", userAgent)
@@ -76,9 +76,9 @@ func (h *HTTPOptions) Init(url string) *HTTPOptions {
 
 // Version is the fortio package version (TODO:auto gen/extract).
 const (
-	userAgent           = "istio/fortio-" + periodic.Version
-	retcodeOffset       = len("HTTP/1.X ")
-	DefaultTimeOutValue = 3
+	userAgent                  = "istio/fortio-" + periodic.Version
+	retcodeOffset              = len("HTTP/1.X ")
+	HTTPReqTimeOutDefaultValue = 3 * time.Second
 )
 
 // HTTPOptions holds the common options of both http clients and the headers.
@@ -93,8 +93,8 @@ type HTTPOptions struct {
 	// ExtraHeaders to be added to each request.
 	extraHeaders http.Header
 	// Host is treated specially, remember that one separately.
-	hostOverride string
-	Timeout      int64 // timeout value for http request in terms of second
+	hostOverride   string
+	HTTPReqTimeOut time.Duration // timeout value for http request in terms of second
 }
 
 // ResetHeaders resets all the headers, including the User-Agent one.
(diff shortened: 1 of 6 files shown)


olimpias:
I found a bug. I will look at night

olimpias:
Okay I fixed

fortio_main.go
```diff
@@ -82,6 +82,7 @@ var (
 	profileFlag     = flag.String("profile", "", "write .cpu and .mem profiles to file")
 	keepAliveFlag   = flag.Bool("keepalive", true, "Keep connection alive (only for fast http 1.1)")
 	halfCloseFlag   = flag.Bool("halfclose", false, "When not keepalive, whether to half close the connection (only for fast http)")
+	timeoutFlag     = flag.Int64("timeout", fhttp.DefaultTimeOutValue, "Http request timeout interval value in terms of second")
```
ldemailly: let’s use a duration not an int

ldemailly:
(i have not reviewed all yet)

also there are probably more than 1 timeout: connect timeout, read timeout, ...
also see what the fast client is using
you can use the delay= to test

olimpias:
Hmm, I didn't release that I need to add http req timeout to basic client. Cool:). Btw at stats.go, while I was merging two histograms, I tried to maximize their values scope range. So that I choose to select highest divider and offset. Is this logic meaningful?

fhttp/http.go
```diff
@@ -523,9 +530,10 @@ func (c *BasicClient) Fetch() (int, []byte, int) {
 		log.Debugf("Reusing socket %v", *conn)
 	}
 	c.socket = nil // because of error returns
+	conErr := conn.SetReadDeadline(time.Now().Add(c.reqTimeout))
 	// Send the request:
 	n, err := conn.Write(c.req)
-	if err != nil {
+	if err != nil || conErr != nil {
```
ldemailly, line 536: how did you test this ?
also this is too short for proxy requests in fetch, need to either raise it or set it higher in the uihandler
(see the circle ci failure and/or try:)

http://localhost:8080/fortio/fetch/localhost:8080/fortio/?url=http%3A%2F%2Flocalhost%3A8080%2Fecho&t=10s&load=Start

ie pasting `localhost:8080/fortio/?url=http%3A%2F%2Flocalhost%3A8080%2Fecho&t=10s&load=Start`
in the fetch box of http://localhost:8080/fortio/

will run for 10s thus timeout

stats/stats.go
```diff
@@ -422,6 +422,27 @@ func (h *Histogram) copyHDataFrom(src *Histogram) {
 	}
 }
 
+// Merge two different histogram with different scale parameters
+// Highest offset and divider value will be selected on new Histogram to maximize scope
+func Merge(h1 *Histogram, h2 *Histogram) *Histogram {
+	var divider float64
```
ldemailly: nit: I think:
```
divider := h1.Divider
if  h2.Divider > h1.Divider {
	divider = h2.Divider
}
```
would be a more go/standard way to write this (quite a lot shorter, gets the right type automatically etc)

stats/stats.go
```diff
@@ -422,6 +422,27 @@ func (h *Histogram) copyHDataFrom(src *Histogram) {
...
+	var offset float64
+	if h2.Divider > h1.Divider {
+		divider = h2.Divider
+	} else {
+		divider = h1.Divider
+	}
+	if h2.Offset > h1.Offset {
```
ldemailly: I would think you want the smallest offset to not miss data ?

fhttp/http.go
olimpias: I had tested with TestHTTPRunner method. I put sleep before reading tcp socket to see is timeout working. Also, now I tested with url you gave. It is okey. When I set HTTPReqTimeOutDefaultValue to 10, it failed and when I set to 11 it worked.

stats/stats.go
olimpias: Yes, you are right. I didnt think like that.

ldemailly (approved):
Thanks - can we change the value in UI handler - to say 5 minutes ?
here: https://github.com/istio/fortio/blob/master/ui/uihandler.go#L414

olimpias:
Sure, I changed

ldemailly:
thx!

pro tip: set your IDE to auto run go fmt on save (for instance atom + go plus does that automatically) that way you don't get build failure for minor formatting issues

olimpias:
Okay, Thanks for your tip and comments.  I will try to figure out to do that in GoLang IDE.

PR #70, opened 2018-01-02
Bug fix for CalcPercentile method

olimpias:
I think, I fix the bug. If you confirm my fix @ldemailly , I will change other test methods too.

### DIFF
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -244,6 +244,8 @@ func (h *Histogram) CalcPercentile(percentile float64) float64 {
 	var prevPerc float64
 	var perc float64
 	found := false
+	prevLowest := float64(0)
+	currentPercLowest := float64(0)
 	cur := h.Offset
 	// last bucket is virtual/special - we'll use max if we reach it
 	// we also use max if the bucket is past the max for better accuracy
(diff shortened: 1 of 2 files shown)


olimpias:
Also // TODO: fix the p51 (and p1...), should be 0 not 10 looks like fixed. result return 0.2 for p51

ldemailly:
Thanks for the PR ! can you explain/comment what you expect the change to do/fix ?
given how it seems to make the "small p" test fail: like:
	stats_test.go:166: p0.1: got 0.005, not as expected 1.045
it is not working/fixing properly
(when p gets close to p0 the value returned should get close to data set min;
likewise when p gets close to p100 the value return should get close to data set max

it does do that despite the p75 issue so let's not loose that with any fix)

ldemailly:
hey I gave this some more thoughts and I think I need to change the bucket definition to be )a,b] instead of the current [a,b(

i will start a PR

olimpias:
Hey, okay. To be sure, the for loop starts from 0 to lastIdx will be reversed in Export? Like for loop from lastIdx to 0? Also will you do that for CalcPercentile? I could try to do that If I understand you right.

ldemailly:
I added #74

it's not about the loop order but whether a buck like [1,2] includes 1 or 2. right now it's setup where 1 is included but 2 isn't except in the last bucket where "max" is included
this is like that because that way 0 is included in first bucket, but we can switch which one is the exception (or add a bucket)

olimpias:
I could not understand the solution that you proposed by adding a bucket or switching which one is the exception. Did you mean deleting instead of switching? I will be waiting for you PR. It could be safer :).

olimpias:
#77 is solved the bug about CalcPercentile. So I m closing this PR. I will go with //TODO consider using an interval search for the last N big buckets

Issue #31, opened 2017-11-21
increase coverage


ldemailly:
it's pretty low in some parts but can be increased fairly easily

except for the ui handler - for which some sort of recording of Webtest.sh would help

ldemailly:
it's now pretty good - except for ui/ and grpc/

PR #71, opened 2018-01-02
Increase coverage: logger 62->100%, periodic 80 -> 98.6%, fhttp 66->76% + bug fix

ldemailly:
And found and fixed a tricky bug in watcher logic thanks to this

Coverage increase is for #31


periodic/periodic.go
```diff
@@ -174,39 +174,42 @@ func (r *RunnerOptions) Normalize() {
...
+			n := gOutstandingRuns
 			runnerChan := r.Stop // need a copy to not race with assignement to nil
-			abortChan := gAbortChan
 			if gAbortChan == nil {
-				log.Infof("First outstanding run starting, catching signal")
+				log.Infof("WATCHER %d First outstanding run starting, catching signal", n)
 				gAbortChan = make(chan os.Signal, 1)
 				signal.Notify(gAbortChan, os.Interrupt)
 			}
+			abortChan := gAbortChan
```
ldemailly, line 184: this is the serious bug fix, the rest of the changes are just to make logging easier to debug with a watcherId

fhttp/httprunner_test.go
```diff
@@ -89,3 +94,29 @@ func TestHTTPRunnerBadServer(t *testing.T) {
...
+		log.Fatalf("Unable to listen to dynamic port: %v", err)
+	}
+	port := listener.Addr().(*net.TCPAddr).Port
+	log.Infof("Using port: %d", port)
+	listener.Close()
+	url := fmt.Sprintf("http://localhost:%d/debugx1", port)
+	go func() {
+		Serve(port, "/debugx1")
+	}()
+	time.Sleep(100 * time.Millisecond)
```
douglas-reid, line 113: anything we can do here to not have a magic sleep ?

fhttp/httprunner_test.go
```diff
@@ -89,3 +94,29 @@ func TestHTTPRunnerBadServer(t *testing.T) {
 	}
 	log.Infof("Got expected error from mismatch/bad server: %v", err)
 }
+
+// need to be the last test as it installs Serve() which would make
+// the error test for / url above fail:
+
+func TestServe(t *testing.T) {
```
douglas-reid, line 101: instead of doing all of the logic ourselves, could we use `httptest.NewServer` instead (see: http://devdocs.io/go/net/http/httptest/index#NewServer)?

periodic/periodic_test.go
```diff
@@ -152,3 +171,88 @@ func TestID(t *testing.T) {
...
+	if actual != expected {
+		t.Errorf("Extra high qps executed unexpected number of times %d instead %d", actual, expected)
+	}
+	if count != expected {
+		t.Errorf("Extra high qps executed unexpected number of times %d instead %d", count, expected)
+	}
+}
+
+func Test2Watchers(t *testing.T) {
+	time.Sleep(200 * time.Millisecond)
```
douglas-reid: why does the test start with a `time.Sleep` ?

periodic/periodic_test.go
```diff
@@ -152,3 +171,88 @@ func TestID(t *testing.T) {
...
+	{
```
douglas-reid: do you need this block for anything?  is this purely for org?  should it be a separate fn ?

periodic/periodic_test.go
```diff
@@ -152,3 +171,88 @@ func TestID(t *testing.T) {
...
+		gAbortMutex.Unlock()
+		gAbortChan <- os.Interrupt
+	}
+	time.Sleep(200 * time.Millisecond)
+	// once out of scope
+	gAbortMutex.Lock()
+	if gOutstandingRuns != 0 {
+		t.Errorf("expecting 0 watches, found %d", gOutstandingRuns)
+	}
+	gAbortMutex.Unlock()
```
douglas-reid: this test appears to rely on a good deal of internals. is there any way we could make this less reliant on implementation details?

can we gather stats on number of outstanding runs?  or redirect logs to validate that there were two watchers?

periodic/periodic_test.go
```diff
@@ -152,3 +171,88 @@ func TestID(t *testing.T) {
...
+	time.Sleep(200 * time.Millisecond)
+	{
+		o1 := RunnerOptions{}
+		r1 := newPeriodicRunner(&o1)
+		o2 := RunnerOptions{}
+		r2 := newPeriodicRunner(&o2)
+		time.Sleep(200 * time.Millisecond)
+		gAbortMutex.Lock()
+		if gOutstandingRuns != 2 {
+			t.Errorf("expecting 2 watches, found %d for (%v %v)", gOutstandingRuns, r1, r2)
```
douglas-reid: super nit: typical go style is `got, want` (as opposed to `want, got`)

periodic/periodic_test.go
ldemailly: it was for the 2 requests objects to get out of scope but I ended up using abort instead so I'm removing it

fhttp/httprunner_test.go
ldemailly: the proper way would be to have a ready channel but this is a quick test - though I suppose the sleep does make the test likely brittle...

periodic/periodic_test.go
ldemailly: because the previous test still has watchers running - I suppose I can add a "wait until 0"

periodic/periodic_test.go
ldemailly: fixing

periodic/periodic_test.go
ldemailly: yes it's quite white box and partially to increase coverage while testing something
this being said those state are ok to test/use in tests I think

rkpagadala (approved):
gAbortChan race was interesting. Nice

fhttp/httprunner_test.go
ldemailly: the test is to test serve and it just gets a local free port first
(but I added todo on how fhttp.Serve can be improved)

PR #77, opened 2018-01-05
Add delay probability, Change bucket definition and fix calcPerc bugs, drawing improvements

### LINKED ISSUE #74: incorrect/misleading cumulative % graphing
for instance:

screen shot 2018-01-04 at 10 45 55 am

```
      {
        "Start": 0.04,
        "End": 0.045000000000000005,
        "Percent": 38.41961852861036,
        "Count": 43
      },
      {
        "Start": 0.045000000000000005,
        "End": 0.05,
        "Percent": 43.596730245231605,
        "Count": 57
      },
      {
        "Start": 0.05,
        "End": 0.060000000000000005,
        "Percent": 51.316984559491374,
        "Count": 85
      },
      {
        "Start": 0.060000000000000005,
        "End": 0.07,
        "Percent": 59.2188919164396,
        "Count": 87
      },
      {
        "Start": 0.07,
        "End": 0.08,
        "Percent": 65.39509536784742,
        "Count": 68
      },
      {
        "Start": 0.08,
        "End": 0.09000000000000001,
        "Percent": 72.11625794732062,
        "Count": 74
      },
      {
        "Start": 0.09000000000000001,
        "End": 0.1,
        "Percent": 76.93006357856494,
        "Count": 53
      },
      {
        "Start": 0.1,
        "End": 0.2,
        "Percent": 99.09173478655768,
        "Count": 244
      },
      {
        "Start": 0.2,
        "End": 0.246923071,
        "Percent": 100,
        "Count": 10
      }
    ],
  ```

means 99.09% is only reached at 0.2 not 0.15

but if we want to have exact match for round numbers we should change the bucket definition to be inclusive instead of exclusive

### PR DESCRIPTION
**Pull request by ldemailly**
Add delay probability, Change bucket definition and fix calcPerc bugs, drawing improvements

Also adding support for RecordN to record N instance of same data point  
 
Fixes #74

### DIFF
--- a/fhttp/http.go
+++ b/fhttp/http.go
@@ -57,14 +57,19 @@ var (
 )
 
 // NewHTTPOptions creates and initialize a HTTPOptions object.
+// It replaces plain % to %25 in the url. If you already have properly
+// escaped URLs use o.URL = to set it.
 func NewHTTPOptions(url string) *HTTPOptions {
 	h := HTTPOptions{}
 	return h.Init(url)
 }
 
 // Init initializes the headers in an HTTPOptions (User-Agent).
+// It replaces plain % to %25 in the url. If you already have properly
+// escaped URLs use o.URL = to set it.
 func (h *HTTPOptions) Init(url string) *HTTPOptions {
-	h.URL = url
+	// unescape then rescape % to %25 (so if it was already %25 it stays)
+	h.URL = strings.Replace(strings.Replace(url, "%25", "%", -1), "%", "%25", -1)
 	h.NumConnections = 1
 	if h.HTTPReqTimeOut == 0 {
 		h.HTTPReqTimeOut = HTTPReqTimeOutDefaultValue
@@ -809,6 +814,7 @@ func generateStatus(status string) int {
 			return codes[i]
 		}
 	}
+	log.Debugf("[0.-100.[ for %s roll %f no hit, defaulting to OK", status, res)
 	return http.StatusOK // default/reminder of probability table
 }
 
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: stats/stats_test.go [hunk 1]
```diff
@@ -168,6 +169,34 @@ func TestHistogram(t *testing.T) {
 	}
 }
 
+func XTestHistogramData(t *testing.T) {
```
ldemailly: cc @olimpias - this is what I meant/expect
I will add the code now that I have a test :-)

olimpias: Now I see, Thanks for example :).

ldemailly:
> The test in this WIP commit are broken because the linear interpolation
isn’t there yet - but I wanted a checkpoint

ldemailly:
This is now ready to review. PTAL. TIA

ldemailly:
before:
screen shot 2018-01-06 at 1 58 25 pm

after:
screen shot 2018-01-06 at 2 01 17 pm

fhttp/http.go
```diff
@@ -57,14 +57,19 @@ var (
...
 	return h.Init(url)
 }
 
 // Init initializes the headers in an HTTPOptions (User-Agent).
+// It replaces plain % to %25 in the url. If you already have properly
+// escaped URLs use o.URL = to set it.
 func (h *HTTPOptions) Init(url string) *HTTPOptions {
-	h.URL = url
+	// unescape then rescape % to %25 (so if it was already %25 it stays)
+	h.URL = strings.Replace(strings.Replace(url, "%25", "%", -1), "%", "%25", -1)
```
douglas-reid, line 72: do we want something here like: `url.QueryEscape()` and `url.QueryUnescape()`  or is the `%` special here?

fhttp/http_test.go
```diff
@@ -283,6 +286,38 @@ func TestParseStatus(t *testing.T) {
...
+		{"10ms:-1", -1},
+		{"20ms:101", -1},
+		{"20ms:101%", -1},
+		{"10ms:45,100ms:56", -1},
+		// Max delay case:
+		{"10s:45,10s:55", 1 * time.Second},
+		// Good cases
+		{"100ms", 100 * time.Millisecond},
+		{"100ms:100", 100 * time.Millisecond},
+		{"100ms:100%", 100 * time.Millisecond},
```
douglas-reid, line 308: curious: why allow '%' in the delay string to be valid?

douglas-reid (commented):
looks pretty good. just curious about %s in delay specs.

fhttp/http_test.go
ldemailly: it's the probability percentage - adding the optional % makes it a bit more self documented (I thought, but clearly not ;-) )

douglas-reid: i understood the purpose. i was just wondering if it is worth the hassle.

fhttp/http.go
ldemailly: query escape would mangle / and other characters, % is somewhat special/hacky yes

I'm not sure what to do cleanly here as the url with a raw % is invalid, I'm just making it work despite being invalid so it looks pretty in the UI text field (I think that's also why I didn't have "%" the first time around for the status weights PRs)

PR #82, opened 2018-01-13
Support for exact number of calls

### LINKED ISSUE #75: add support for "exact number of request"
For fortio use in functional test, being able to specify exactly 100 requests for instance would be useful
(right now if you ask for 10 qps, 10 seconds, 8 connections you get 100 requests + 8 warmup requests)

ldemailly:
Fixes #75
If you set RunnerOptions.Exactly to any number that exact number of
calls will be made, and without additional warmup - it will still try
to reach the target QPS but if the service is slower it will take the
time it will take

On the command line this mode is selected using “-n num”
(The replaces/extends the grpc ping count flag)

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,9 +321,9 @@ make lint
 make webtest
 ```
 
-When modifying Javascript:
+When modifying Javascript, check with [standard](https://github.com/standard/standard):
 ```
-standard --fix fortio_chart.js
+standard --fix ui/static/js/fortio_chart.js
 ```
 
 ## See also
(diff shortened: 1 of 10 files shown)


ldemailly:
@mandarjog this could help tracking down the rate limit and metrics issues:
if you ask it to do N requests it will do exactly N whatever the time it takes and without extra warmup calls

andraxylia (approved):

mandarjog:
@ldemailly used it right away with `-n 100`, thanks.

ldemailly:
> @ldemailly used it right away with -n 100, thanks.

@mandarjog cool! I'm glad ! you can also use the ui if you run fortio server either inside or outside your cluster

screen shot 2018-01-12 at 11 19 22 pm

And once https://github.com/istio/istio/pull/2596 goes in, from test code, using
```
Exactly: 100, 
```
or `o.Exactly = 100` in the (HTTP)RunnerOptions

PR #83, opened 2018-01-14
Val2Bucket's memory usage is decreased for upper values

olimpias:
"TODO: consider using an interval search for the last N big buckets" is done.


stats/stats.go
```diff
@@ -198,24 +208,53 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
 			idx++
 		}
-		val2Bucket[i] = idx
+		constantTimeVal2Bucket[i] = idx
 	}
+	idx++ // Not include 1000
+	j := 0
+	for ; j < line5EndToEndLength-1; j++ {
+		currentValue := histogramBucketValues[idx]
+		previosValue := histogramBucketValues[idx-1]
```
ldemailly: typo: `previous`

stats/stats.go
```diff
@@ -198,24 +208,53 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
 	// coding bug detection (aka impossible if it works once)
 	if idx != numValues {
 		log.Fatalf("Bug in creating histogram buckets idx %d vs numbuckets %d (last val %d)", idx, numValues, lastV)
 	}
+}
 
+// lookUpIdx looks for scaledValue's index in histogramBucketValues
+func lookUpIdx(scaledValue int) int {
+	constantOrLinearDeciderValue := int(histogramBucketValues[line5EndIndex])
+	if constantOrLinearDeciderValue > scaledValue { //constant
```
ldemailly: super minor nit but it would read (to me) easier if it was
if scaledValue <= constantOrLinearDeciderValue

also maybe constantOrLinearDeciderValue should just be maxLookupValue or some such

stats/stats.go
```diff
@@ -198,24 +208,53 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
+		return constantTimeVal2Bucket[scaledValue]
+	}
+	iteratorValue := constantOrLinearDeciderValue
+	for _, value := range linearTimeVal2Bucket {
+		iteratorValue += value.Range
```
ldemailly: avoid this sumation by doing the pre calculation in init

ldemailly (commented):
thanks for this !

see below - also

can you consider adding a benchmark (before/after maybe too) ?

olimpias:
Yes, I will try to add benchmark on weekend.

olimpias:
I could not do a benchmark due to work, health and school problems. I will try to finish at the end of week.

ldemailly:
No worries ! I hope things get better!

ldemailly:
ping :)

olimpias:
Configuration that I used for tests
configuration

Master Branch Results
master_branch_datas
 
master_branch_data

My Branch Results 
my_branch_data

my_branch_datas

I hope, you were looking to this :).

ldemailly:
for benchmark I meant something like
https://github.com/istio/fortio/blob/896d2dae1a2f97099292aa50e14791cf0b1c48c6/fhttp/http_test.go#L618

because the values are large (5s) you can't do an e2e test that will test the performance of the library, you need to send large values in a unit test (though it's great you did check functionally it works the same)

olimpias:
I received 100000000 10.8 ns/op from master branch. From my branch, I received 50000000 20.0 ns/op. I push the code that I used in benchmark.

stats/stats_test.go
```diff
@@ -724,3 +724,10 @@ func TestNaN(t *testing.T) {
 // should get 3 buckets 0-1 with count 1
 // 1-2 with count 3
 // 2-2.5 with count 1
+
+func BenchmarkBucketLookUpWithHighestValue(b *testing.B) {
+	testHistogram := NewHistogram(0, 1)
+	for i := 0; i < b.N; i++ {
+		testHistogram.Record(100000)
```
ldemailly: add another one with a bit more varied data (maybe rand though that may test the speed of rand more than histogram’s speed)
and say what you get before/after

olimpias:
I received 20000000   74.2 ns/op from my branch and  20000000   61.0 ns/op from master when I used BenchmarkBucketLookUpWithRandomValues method.

stats/stats.go
```diff
@@ -135,13 +135,24 @@ var (
...
-	numValues  = len(histogramBucketValues)
-	numBuckets = numValues + 1 // 1 special first bucket is <= 0; and 1 extra last bucket is > 100000
-	firstValue = float64(histogramBucketValues[0])
-	lastValue  = float64(histogramBucketValues[numValues-1])
-	val2Bucket []int
+	numValues              = len(histogramBucketValues)
+	numBuckets             = numValues + 1 // 1 special first bucket is <= 0; and 1 extra last bucket is > 100000
+	firstValue             = float64(histogramBucketValues[0])
+	lastValue              = float64(histogramBucketValues[numValues-1])
+	line5EndToEndLength    = 13
```
ldemailly: it's not line5 - 1000 is end of line 8 but that's not a good name anyway

let's add a single constant, on a new line to minimize the diff:

 maxArrayValue = 1000 // Last value looked up as O(1) array, the rest is linear search

and calculate the rest in init as needed

stats/stats.go
```diff
@@ -135,13 +135,24 @@ var (
 		2000, 3000, 4000, 5000, 7500, 10000, // another order of magnitude coarsly covered
 		20000, 30000, 40000, 50000, 75000, 100000, // ditto, the end
 	}
-	numValues  = len(histogramBucketValues)
-	numBuckets = numValues + 1 // 1 special first bucket is <= 0; and 1 extra last bucket is > 100000
-	firstValue = float64(histogramBucketValues[0])
-	lastValue  = float64(histogramBucketValues[numValues-1])
-	val2Bucket []int
```
ldemailly: leave that one unchanged name

stats/stats.go
```diff
@@ -135,13 +135,24 @@ var (
...
+	firstValue             = float64(histogramBucketValues[0])
+	lastValue              = float64(histogramBucketValues[numValues-1])
+	line5EndToEndLength    = 13
+	line5EndIndex          = numValues - line5EndToEndLength //to reach 1000
+	constantTimeVal2Bucket []int                             //up to 1000
+	linearTimeVal2Bucket   []UpperValueIndexer               // later than 1000 to 100000
 )
 
+// UpperValueIndexer holds upper values of histogramBucketValues such as 1000 to 100000
+// Index and Scope are kept. Scope is estimated with current index value - previous index value plus previous
```
ldemailly: I don't get "Scope"

stats/stats.go
```diff
@@ -135,13 +135,24 @@ var (
...
+// sum of scopes.
+type UpperValueIndexer struct {
```
ldemailly: I don;'t think this is needed - we just need the index for maxArrayValue in 	histogramBucketValues

stats/stats.go
```diff
@@ -203,24 +214,54 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
 	return h
 }
 
-// Tradeoff memory for speed (though that also kills the cache so...)
-// this creates an array of 100k (max value) entries
-// TODO: consider using an interval search for the last N big buckets
+// Val2Bucket values are kept in two different structure
+// constantTimeVal2Bucket allows you reach between 0 and 1000 in constant time
+// linearTimeVal2Bucket allows you to reach between 1000 and 100000 in linear time
+// linearTimeVal2Bucket approach saves memory usage but loses speed
```
ldemailly: btw we can add a new TODO to do a binary search instead of linear (though that's maybe overkill depending on benchmark)

stats/stats_test.go
ldemailly: btw the value that is slowest is lastValue-1 so 99999 not 100000

stats/stats.go
```diff
@@ -203,24 +214,54 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
 func init() {
 	lastV := int32(lastValue)
-	val2Bucket = make([]int, lastV+1)
+	line5EndValue := histogramBucketValues[line5EndIndex]
```
ldemailly: that is maxArrayValue and the input to this

stats/stats.go
```diff
@@ -203,24 +214,54 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
+	linearTimeVal2Bucket[j] = UpperValueIndexer{Index: idx, Scope: 1} // for 100000
 	// coding bug detection (aka impossible if it works once)
 	if idx != numValues {
 		log.Fatalf("Bug in creating histogram buckets idx %d vs numbuckets %d (last val %d)", idx, numValues, lastV)
 	}
+}
 
+// lookUpIdx looks for scaledValue's index in histogramBucketValues
+func lookUpIdx(scaledValue int) int {
+	maxLookupValue := int(histogramBucketValues[line5EndIndex])
```
ldemailly: still maxArrayValue (or you can call it maxLookupValue in the contant)

ldemailly (commented):
sorry to be quite strict but this is core code for fortio so here are some changes I think are needed

stats/stats_test.go
olimpias: Last value is selected from histogramBucketsValues array length - 1 which maps to 100000 as value. So the value that I used, allows me to iterate over all values. If scaledVal >= lastValue  is like this in record, you were right.

olimpias:
After the changes, Highest value branchmark runs 50000000 27.4 ns/op and random value lookup runs 20000000 77.5 ns/op

stats/stats.go
olimpias: I used scope for value interval index. As you point out, it was pointless to use. So I wont explain it :D.

olimpias:
Why is that failed, I couldnt see the details. Having 404. Also I realized that now, In my pc, I can not able to run make webtest.

stats/stats.go
```diff
@@ -135,11 +135,13 @@ var (
...
-	numBuckets = numValues + 1 // 1 special first bucket is <= 0; and 1 extra last bucket is > 100000
-	firstValue = float64(histogramBucketValues[0])
-	lastValue  = float64(histogramBucketValues[numValues-1])
-	val2Bucket []int
+	numValues            = len(histogramBucketValues)
+	numBuckets           = numValues + 1 // 1 special first bucket is <= 0; and 1 extra last bucket is > 100000
+	firstValue           = float64(histogramBucketValues[0])
+	lastValue            = float64(histogramBucketValues[numValues-1])
+	maxArrayValue        = int32(1000) // Last value looked up as O(1) array, the rest is linear search
+	val2Bucket           []int         //up to 1000
```
ldemailly: cosmetic/diff minimization: move that line above the maxArrayValue and insert a blank line and re-go-fmt - tia (the unchanged block won't look changed as it shows now just because of whitespace; I wish github would ignore whitespace for go files...)

stats/stats.go
```diff
@@ -203,24 +205,61 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
-// TODO: consider using an interval search for the last N big buckets
+// Val2Bucket values are kept in two different structure
+// val2Bucket allows you reach between 0 and 1000 in constant time
+// linearTimeVal2Bucket allows you to reach between 1000 and 100000 in linear time
+// linearTimeVal2Bucket approach saves memory usage but loses speed
+// TODO: change linear time to O(log(N)) with binary search for linearTimeVal2Bucket
 func init() {
-	lastV := int32(lastValue)
-	val2Bucket = make([]int, lastV+1)
+	maxArrayValueIndex := 0
```
ldemailly: initialize to -1 (because 0 is valid and you want to catch not found fast)

stats/stats.go
```diff
@@ -203,24 +205,61 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
+	for i, value := range histogramBucketValues {
+		if value == maxArrayValue {
+			maxArrayValueIndex = i
+			break
+		}
+	}
+	val2Bucket = make([]int, maxArrayValue)
```
ldemailly: can you insert here (just after the for loop searching for maxArrayValue) an 
```
if maxArrayValueIndex == -1 {
  log.Fatalf("Bug boundary maxArrayValue=%d not found in bucket list %v", maxArrayValue, histogramBucketValues)
}
```
and try it once but putting 1001 instead of 1000 for the constant, temporarily in your code; this is in addition to the other test below


also move that line`val2Bucket = make([]int, maxArrayValue)` back to the top so it's easier to read the diff (the array will be of size maxArrayValue (possibly + or - 1) and there is nothing to calculate first)

stats/stats.go
```diff
@@ -203,24 +205,61 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
+	linearTineVal2BucketLen := numValues - maxArrayValueIndex
```
ldemailly: typo: Tine -> Time but as mentioned above, linearTimeVal2Bucket is not needed

stats/stats_test.go
```diff
@@ -724,3 +724,17 @@ func TestNaN(t *testing.T) {
 // should get 3 buckets 0-1 with count 1
 // 1-2 with count 3
 // 2-2.5 with count 1
+
```
ldemailly, line 772: also add a test around the maxArrayValue boundary, I want to confirm there isn't an off by 1 error 

so test that if you record 999 it ends up in the >= 900 < 1000 bucket
same for 999.99

check that 1000 shows up in the >= 1000 < 2000 bucket
same for 1000.01 and 1001 and 1999.99

check that 2000 and 2001 show up in >= 2000 bucket (the first virtual bucket)

ideally should be table driven like the round test and not like the early tests checking the text output (which is somewhat brittle)

ldemailly:
I filled #159 for the failure which seems unrelated to your PR

How come your PC can't run webtest ? what happens?

stats/stats.go
```diff
@@ -135,11 +135,13 @@ var (
...
-	firstValue = float64(histogramBucketValues[0])
-	lastValue  = float64(histogramBucketValues[numValues-1])
-	val2Bucket []int
+	numValues            = len(histogramBucketValues)
+	numBuckets           = numValues + 1 // 1 special first bucket is <= 0; and 1 extra last bucket is > 100000
+	firstValue           = float64(histogramBucketValues[0])
+	lastValue            = float64(histogramBucketValues[numValues-1])
+	maxArrayValue        = int32(1000) // Last value looked up as O(1) array, the rest is linear search
+	val2Bucket           []int         //up to 1000
+	linearTimeVal2Bucket []int         // later than 1000 to 100000
```
ldemailly: I don't think this is needed, you just need maxArrayValueIndex and use the existing histogramBucketValues for the search

ldemailly:
also please "make pull" in your fork/branch so github doesnt say "This branch is out-of-date with the base branch" - ty!

olimpias:
I'm having below message while trying to run make webtest

```
./Webtest.sh
+ make docker-internal TAG=webtest
make[1]: Entering directory '/home/cemturker/go/src/istio.io/fortio'
### Now building docker.io/istio/fortio:webtest
docker build -f Dockerfile -t docker.io/istio/fortio:webtest .
Sending build context to Docker daemon 13.27 MB
Step 1 : FROM istio/fortio.build:v5 as build
Error parsing reference: "istio/fortio.build:v5 as build" is not a valid repository/tag
Makefile:94: recipe for target 'docker-internal' failed
make[1]: *** [docker-internal] Error 1
make[1]: Leaving directory '/home/cemturker/go/src/istio.io/fortio'
+ exit 1
```

stats/stats.go
olimpias: I thought you mention about struct that I used.

stats/stats.go
olimpias: May I keep maxArrayValueIndex instead of maxArrayValue as constant? because when I use lookup, I will always need to look for maxArrayValue index.

olimpias: Nevermind, the values are not constant. I can set when ever I want. I thought they were constant :).

ldemailly:
it means your docker is too old, it doesn't understand multi stage builds
that's a good addition to the faq I suppose

do "check for updates..." in the docker menu if your machine is a mac or windows

or update to docker CE 17 +

ldemailly:
I added this to the faq:
https://github.com/istio/fortio/wiki/FAQ#i-ran-make-webtest-and-it-failed
thanks for stepping on this issue and mentioning it

stats/stats.go
ldemailly: neither the struct nor a new array is needed

stats/stats.go
ldemailly: ping ^^^

stats/stats.go
```diff
@@ -203,24 +205,46 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
+	// coding bug detection (aka impossible if it works once) until 1000
+	if idx != maxArrayValueIndex {
+		log.Fatalf("Bug in creating histogram index idx %d vs index %d up to %d", idx, int(maxArrayValue), maxArrayValue)
 	}
+}
 
+// lookUpIdx looks for scaledValue's index in histogramBucketValues
+// TODO: change linear time to O(log(N)) with binary search
+func lookUpIdx(scaledValue int) int {
+	scaledVale32 := int32(scaledValue)
```
ldemailly: either "scaledVal32" or "scaledValue32" but not "scaledVale32"

stats/stats_test.go
```diff
@@ -720,7 +720,46 @@ func TestNaN(t *testing.T) {
 	}
 }
 
+func TestBucketLookUp(t *testing.T) {
+	var tests = []struct {
+		inputRecord   float64 // input
+		expectedIndex int     // expected
+
+	}{
+		{inputRecord: 999, expectedIndex: 43},
```
ldemailly: let's not do a complete whitebox test here, it's quite impossible for a human to know if 43 is the right bucket, use HistogramData and check the start and/or end of the first bucket should be as I mentioned 900-1000 etc etc

stats/stats.go
```diff
@@ -135,11 +135,13 @@ var (
...
-	numValues  = len(histogramBucketValues)
-	numBuckets = numValues + 1 // 1 special first bucket is <= 0; and 1 extra last bucket is > 100000
-	firstValue = float64(histogramBucketValues[0])
-	lastValue  = float64(histogramBucketValues[numValues-1])
-	val2Bucket []int
+	numValues          = len(histogramBucketValues)
+	numBuckets         = numValues + 1 // 1 special first bucket is <= 0; and 1 extra last bucket is > 100000
+	firstValue         = float64(histogramBucketValues[0])
+	lastValue          = float64(histogramBucketValues[numValues-1])
+	maxArrayValue      = int32(1000) // Last value looked up as O(1) array, the rest is linear search
```
ldemailly: ping on inserting a new line before this line
and leaving val2Bucket  in the first set

ldemailly (commented):
also again please rebase

stats/stats_test.go
olimpias: When I add one record, it doesnt show the start and end as 900 and 1000 instead it only shows the record value which I added as start and end. For solution, I changed expectedIndex with startRecord. For example, I added two record which are 999 and 900,0004. 900,0004 is the start of the bucket. So when I add these two values, I expect from hData to create only one data.

stats/stats.go
olimpias: Okay I do this now. But I couldn't figure out to do white space stuff with re-go-fmt - tia.

stats/stats_test.go
ldemailly: it is tuned to reduce buckets that contain either min or max you can add "1" and/or "1000000" to get the expected bucket span

stats/stats.go
ldemailly: if you insert a line, go fmt won't try to make it all aligned and thus it will reduce the diff

if your editor doesn't do it out of the box (I recommend you get a better editor: Atom + go-plus is free and great) you can run goimports or gofmt from the command line

olimpias: make lint ask me to format them. It still causes to be diff.

olimpias: btw, I'm using GoLand

stats/stats.go
```diff
@@ -135,11 +135,14 @@ var (
...
-	numBuckets = numValues + 1 // 1 special first bucket is <= 0; and 1 extra last bucket is > 100000
-	firstValue = float64(histogramBucketValues[0])
-	lastValue  = float64(histogramBucketValues[numValues-1])
-	val2Bucket []int
+	numValues          = len(histogramBucketValues)
+	numBuckets         = numValues + 1 // 1 special first bucket is <= 0; and 1 extra last bucket is > 100000
+	firstValue         = float64(histogramBucketValues[0])
+	lastValue          = float64(histogramBucketValues[numValues-1])
+	val2Bucket         []int
+	maxArrayValue      = int32(1000) // Last value looked up as O(1) array, the rest is linear search
```
ldemailly: insert a blank line BEFORE this line so it looks like
```
...
	val2Bucket []int

         maxArrayValue      = int32(1000) 
...
```
then let your editor reformat, and then update the pr - it won't have any red/green for the first block of vars anymore

stats/stats.go
```diff
@@ -135,11 +135,14 @@ var (
...
+	maxArrayValueIndex = -1          // Index of maxArrayValue
+
```
ldemailly: not here which does nothing

stats/stats_test.go
```diff
@@ -720,7 +720,50 @@ func TestNaN(t *testing.T) {
...
+func TestBucketLookUp(t *testing.T) {
+	var tests = []struct {
+		input float64 // input
+		start float64 // start
+		end   float64 // end
+
+	}{
+		{input: 999, start: 900, end: 1000},
+		{input: 999.99, start: 900, end: 1000},
+		{input: 1000.0004, start: 1000, end: 2000},
```
ldemailly: great, but please put 1000 itself too

stats/stats_test.go
```diff
@@ -720,7 +720,50 @@ func TestNaN(t *testing.T) {
...
+	}
+	h := NewHistogram(0, 1)
+	for _, test := range tests {
+		h.Reset()
+		h.Record(1)
+		h.Record(1000000)
+		h.Record(test.input)
+		hData := h.Export()
+		if hData.Data[1].Start != test.start || hData.Data[1].End != test.end {
+			t.Errorf("Testing data %f is matched with different bucket", test.input)
```
ldemailly: please %+v print tests and the actual value

in go tests always print the actual value first 

so... "Got %+v while expected %+v", hData.Data[1], test)

ldemailly (dismissed):
getting closer ;)

stats/stats_test.go
olimpias: I added .0004 because of 0.0001 value in the record. When I put 1000, it becomes 999,999

ldemailly: yes because the intervals are open on the end https://github.com/istio/fortio/blob/master/stats/stats.go#L121
so let's verify - replace the 1000.0004 which is same as next line by 1000 and update the start/end

olimpias: Okay then, for instance; for 1000, start will be 900 and end will be 1000.

stats/stats.go
```diff
@@ -140,6 +140,9 @@ var (
 	firstValue = float64(histogramBucketValues[0])
 	lastValue  = float64(histogramBucketValues[numValues-1])
 	val2Bucket []int
+
```
ldemailly: 🎉

stats/stats.go
```diff
@@ -140,6 +140,9 @@ var (
 	firstValue = float64(histogramBucketValues[0])
 	lastValue  = float64(histogramBucketValues[numValues-1])
 	val2Bucket []int
```
ldemailly: you can add a comment there saying it ends at maxArrayValue

stats/stats.go
```diff
@@ -203,24 +206,46 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
-// Tradeoff memory for speed (though that also kills the cache so...)
-// this creates an array of 100k (max value) entries
-// TODO: consider using an interval search for the last N big buckets
+// Val2Bucket values are kept in two different structure
+// val2Bucket allows you reach between 0 and 1000 in constant time
 func init() {
-	lastV := int32(lastValue)
-	val2Bucket = make([]int, lastV+1)
+	val2Bucket = make([]int, maxArrayValue)
+	for i, value := range histogramBucketValues {
```
ldemailly: I'm a bit paranoid and the initialization to -1 is a bit far up; can you add
maxArrayValueIndex = -1
just before the for - ty

stats/stats.go
```diff
@@ -203,24 +206,46 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
 
+// lookUpIdx looks for scaledValue's index in histogramBucketValues
+// TODO: change linear time to O(log(N)) with binary search
+func lookUpIdx(scaledValue int) int {
+	scaledValue32 := int32(scaledValue)
+	if scaledValue32 < maxArrayValue { //constant
+		return val2Bucket[scaledValue]
+	}
+	for i := maxArrayValueIndex; i < numValues; i++ {
+		if histogramBucketValues[i] > scaledValue32 {
```
ldemailly, line 244: shouldn't it be >= ? I guess the tests should tell

stats/stats.go
```diff
@@ -203,24 +206,46 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
+			return i
+		}
+	}
+	//select max value
+	return numValues - 1
```
ldemailly: this should never happen, if it does there is another bug somewhere (because there is 
`if scaledVal > lastValue ` below in record()

can you also check with your IDE or the coverage link if that line is hit ?

lastly there is some complication between numValues and numBuckets to double check

please add boundary tests for the last bucket too

olimpias: Not hitting, but I can change numValues to numValues -1 to hit in the for loop. I will add boundary tests.

stats/stats.go
```diff
@@ -203,24 +206,47 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
+	if idx != maxArrayValueIndex {
+		log.Fatalf("Bug in creating histogram index idx %d vs index %d up to %d", idx, int(maxArrayValue), maxArrayValue)
 	}
+}
 
+// lookUpIdx looks for scaledValue's index in histogramBucketValues
+// TODO: change linear time to O(log(N)) with binary search
+func lookUpIdx(scaledValue int) int {
+	scaledValue32 := int32(scaledValue)
+	if scaledValue32 < maxArrayValue { //constant
```
ldemailly, line 240: this also should never happen ? we should not add 'defensive' code that never gets executed and hide bugs when it's internal code, instead we should have good boundary tests that it works as expected

stats/stats.go
```diff
@@ -203,24 +206,47 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
+		return val2Bucket[scaledValue]
+	}
+	for i := maxArrayValueIndex; i < numValues-1; i++ {
```
ldemailly: because of https://github.com/istio/fortio/blob/v0.7.1/stats/stats.go#L242
it means the int value of the input it is -1 at the boundary, ie 1000 data enters here as 999
it's to preserve the open to the left part of the interval

stats/stats.go
```diff
@@ -203,24 +206,47 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
+		if histogramBucketValues[i] > scaledValue32 {
```
ldemailly: I'm having a hard time finding time to verify but this loop seems off because of the above

stats/stats.go
ldemailly: replace with a logFatal("should not be reached")
return 0

olimpias: Okay I removed -1 from for loop.

stats/stats.go
olimpias: This is not a defensive code. That check allows us to reach lower than 1000 values in the bucket.

stats/stats.go
olimpias: hmm than you are right for histogramBucketValues[i] > scaledValue32. It has to be more than equals. You told me at top commands ;). So I'm assuming that 1001 will enter the 999's boundary. I mean that 1001 start will be 999 and end will be 1000. And record value of 1001.1's start value will be 1000 and end value will be 2000.

stats/stats.go
olimpias: But we can separate the method if you want. I'm open to suggestions :)

stats/stats.go
olimpias: I'm changing to >=

stats/stats.go
olimpias: Please check TestBucketLookUp method in stats_test.go. I have added some boundary tests for low values as well. If you suggest any values for boundary test, I can add them

ldemailly: if you look in the code for record(), lookUpIdx() which is private can't be called with values < 1000

stats/stats_test.go
```diff
@@ -720,7 +720,64 @@ func TestNaN(t *testing.T) {
...
+		{input: 11, start: 10, end: 11},
+		{input: 171, start: 160, end: 180},
+		{input: 180, start: 160, end: 180},
+		{input: 801, start: 800, end: 900},
+		{input: 900, start: 800, end: 900},
+		{input: 999, start: 900, end: 1000},
+		{input: 999, start: 900, end: 1000},
+		{input: 999.99, start: 900, end: 1000},
+		{input: 1000, start: 900, end: 1000},
+		{input: 1000.01, start: 900, end: 1000},
```
ldemailly: this is wrong :-) I made you break your previously working code, sorry ! 
1000.01 should definitely be in the ] 1000, 2000 ] bucket like 1001 and 1001.1

it's off by 1

olimpias: Okay, I take it back.

stats/stats.go
olimpias: You mean should I do lookUpIdx public? Even lookUpIdx() is private, record and lookUpIdx is in same file. So I'm expecting that, lookUpIdx method is usable in record. Also inside the lookUpIdx, scaledValue32 < maxArrayValue check allows you to reach values < 1000 as well. If you look the master branch, idx = lookUpIdx(scaledVal) has same responsiblility as idx = val2Bucket[int(scaledVal)]

ldemailly: > should I do lookUpIdx public? 

no, no it should stay private, in theory it could just be the end of record() which already private too

you are right and I forgot this replaces 
		idx = val2Bucket[int(scaledVal)]
so it needs to handle the val2Bucket part too; my bad

stats/stats.go
```diff
@@ -203,24 +206,47 @@ func NewHistogram(Offset float64, Divider float64) *Histogram {
...
+	if scaledValue32 < maxArrayValue { //constant
+		return val2Bucket[scaledValue]
+	}
+	for i := maxArrayValueIndex; i < numValues; i++ {
+		if histogramBucketValues[i] > scaledValue32 {
+			return i
+		}
+	}
+	//select max value
+	return numValues - 1
```
ldemailly: does this statement get triggered? the last bucket isn't numValues - 1
I'd prefer something clearly off if it's not supposed to happen (this time the argument about maxValue already checked in record() should apply)

stats/stats_test.go
```diff
@@ -720,7 +720,64 @@ func TestNaN(t *testing.T) {
...
+		end   float64 // end
+
+	}{
+		{input: 11, start: 10, end: 11},
+		{input: 171, start: 160, end: 180},
+		{input: 180, start: 160, end: 180},
+		{input: 801, start: 800, end: 900},
+		{input: 900, start: 800, end: 900},
+		{input: 999, start: 900, end: 1000},
+		{input: 999, start: 900, end: 1000},
```
ldemailly: duplicate line

stats/stats_test.go
```diff
@@ -720,7 +720,64 @@ func TestNaN(t *testing.T) {
...
+		{input: 1000, start: 900, end: 1000},
+		{input: 1000.01, start: 1000, end: 2000},
+		{input: 1001, start: 1000, end: 2000},
+		{input: 1001.1, start: 1000, end: 2000},
+		{input: 1999.99, start: 1000, end: 2000},
+		{input: 2000, start: 1000, end: 2000},
+		{input: 2001, start: 2000, end: 3000},
+		{input: 2001.1, start: 2000, end: 3000},
+		{input: 100000.0001, start: 75000, end: 100000},
+		{input: 7500, start: 5000, end: 7500},
```
ldemailly: you meant 75000 maybe ? and/or 7500.01

stats/stats.go
olimpias: This case will be used when scaledValue is equals to 100000. because in the for loop,  the highest check is 100000 > scaledValue. IF value is bigger than 100000, it will be handled by record method.

stats/stats_test.go
```diff
@@ -720,7 +720,64 @@ func TestNaN(t *testing.T) {
...
+		{input: 999.99, start: 900, end: 1000},
+		{input: 1000, start: 900, end: 1000},
+		{input: 1000.01, start: 1000, end: 2000},
+		{input: 1001, start: 1000, end: 2000},
+		{input: 1001.1, start: 1000, end: 2000},
+		{input: 1999.99, start: 1000, end: 2000},
+		{input: 2000, start: 1000, end: 2000},
+		{input: 2001, start: 2000, end: 3000},
+		{input: 2001.1, start: 2000, end: 3000},
+		{input: 100000.0001, start: 75000, end: 100000},
```
ldemailly: using the exact number that gets subtracted is a bit brittle - stick to X is in the previous bucket while X.01 is in the next bucket
ie change this to test
99999.99
100000
100000.01
(and if you want
100000.99
100001
)

stats/stats_test.go
```diff
@@ -720,7 +720,64 @@ func TestNaN(t *testing.T) {
...
+		{input: 7500, start: 5000, end: 7500},
+		{input: 10000.99, start: 10000, end: 20000},
+		{input: 99999.99, start: 75000, end: 100000},
+		{input: 100000, start: 75000, end: 100000},
```
ldemailly: 👍

stats/stats_test.go
olimpias: Ah sorry for that :)

stats/stats_test.go
olimpias: Should I do that for all boundary values?

ldemailly:
```
$ git diff
diff --git a/stats/stats.go b/stats/stats.go
index 7062a63..59edba9 100644
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -245,8 +245,8 @@ func lookUpIdx(scaledValue int) int {
                        return i
                }
        }
-       //select max value
-       return numValues - 1
+       log.Fatalf("never reached/bug")
+       return 0
 }
 
 // Record records a data point.
diff --git a/stats/stats_test.go b/stats/stats_test.go
index 0010354..a111b86 100644
--- a/stats/stats_test.go
+++ b/stats/stats_test.go
@@ -743,12 +743,14 @@ func TestBucketLookUp(t *testing.T) {
                {input: 2000, start: 1000, end: 2000},
                {input: 2001, start: 2000, end: 3000},
                {input: 2001.1, start: 2000, end: 3000},
-               {input: 100000.0001, start: 75000, end: 100000},
-               {input: 7500, start: 5000, end: 7500},
-               {input: 10000.99, start: 10000, end: 20000},
+               {input: 75000, start: 50000, end: 75000},
+               {input: 75000.01, start: 75000, end: 100000},
                {input: 99999.99, start: 75000, end: 100000},
                {input: 100000, start: 75000, end: 100000},
-               {input: 100000.0002, start: 100000, end: 1e+06},
+               {input: 100000.01, start: 100000, end: 1e+06},
+               {input: 100000.99, start: 100000, end: 1e+06},
+               {input: 100001, start: 100000, end: 1e+06},
+               {input: 523477, start: 100000, end: 1e+06},
        }
        h := NewHistogram(0, 1)
        for _, test := range tests {
@@ -757,8 +759,9 @@ func TestBucketLookUp(t *testing.T) {
                h.Record(1000000)
                h.Record(test.input)
                hData := h.Export()
-               if hData.Data[1].Start != test.start || hData.Data[1].End != test.end {
-                       t.Errorf("Got %+v while expected %+v", hData.Data[1], test)
+               hd := hData.Data[1]
+               if hd.Start != test.start || hd.End != test.end {
+                       t.Errorf("Got %+v while expected %+v", hd, test)
                }
        }
 }
```
I propose this change ^

ldemailly:
let's add 1 more/last test:
```
func TestAllBucketBoundaries(t *testing.T) {
	h := NewHistogram(0, 1)

	for i, value := range histogramBucketValues {
		v := float64(value)
		h.Reset()
		h.Record(-1)
		h.RecordN(v-0.01, 50)
		h.RecordN(v, 700)       // so first interval gets a count of 750 + 1 for first
		h.RecordN(v+0.01, 1003) // second interval gets a count of 1003 + 1 for last
		h.Record(1e6)
		hData := h.Export()
		var firstInterval int64
		if i == 0 {
			firstInterval = 1 // fist interval is [min, 0[
		}
		var lastInterval int64
		if i == len(histogramBucketValues)-1 {
			lastInterval = 1
		}
		if hData.Data[1-firstInterval].End != v || hData.Data[1-firstInterval].Count != 750+firstInterval {
			t.Errorf("= Boundary, got %+v unexpectedly for %d %g", hData, value, v)
		}
		if hData.Data[2-firstInterval].Start != v || hData.Data[2-firstInterval].Count != 1003+lastInterval {
			t.Errorf("> Boundary, got %+v unexpectedly for %d %g", hData, value, v)
		}
	}
}
```
it checks all the boundaries, let me know if you wonder why/how it works ^

ldemailly:
I'll add the test separately ! thanks for this !

ldemailly (approved):

olimpias:
Thanks for your comments @ldemailly!. I learnt a lot. Btw I'm wondering about the test how it is working. I will check it in  details when I have time and ask you, if I have question about it :).

ldemailly:
I added it on master now, so feel free to comment/ask yes ! thanks !

PR #84, opened 2018-01-17
Fix for timeout not set when not calling Init()

ldemailly:


ldemailly:
@mandarjog this will fix the timeout in rate limit
I didn't review #65 carefully enough thinking about earlier code not calling Init()

fhttp/http_test.go
```diff
@@ -80,6 +80,25 @@ func TestNewHTTPRequest(t *testing.T) {
 	}
 }
 
+func TestMultiInitAndEscape(t *testing.T) {
+	// one escaped already 2 not
+	o := NewHTTPOptions("http://localhost:8080/?delay=1s:10%,0.5s:15%25,0.25s:5%")
```
rkpagadala, line 85: suggestion:
expected := "http://localhost:8080/?delay=1s:10%,0.5s:15%25,0.25s:5%"
o := NewHTTPOptions(expected)

fhttp/http.go
```diff
@@ -216,6 +221,10 @@ func NewStdClient(o *HTTPOptions) Fetcher {
 	if o.NumConnections < 1 {
 		o.NumConnections = 1
 	}
+	// 0 timeout for stdclient doesn't mean 0 timeout... so just warn and leave it
```
rkpagadala, line 224: explain what the effects of timeout of 0, negative value

rkpagadala (approved):

fhttp/http_test.go
ldemailly: expected is not same as input

rkpagadala: ok. it was hard to tell the difference.

ldemailly: yeah it’s a bit tricky those %
(sorry, terse as i have been on the phone)

fhttp/http.go
ldemailly: tbh I have to look it up
initially my test didn’t fail when I commented out the code that was setting the timeout to the default like I did for the fast custom client 

and tried with -1 and that also passed locally, I assumed it meant infinite timeout but then it failed on circle, so I set it back to unitialized/0

PR #86, opened 2018-01-18
Adding `-quiet` mode + misc bug fixes

### LINKED ISSUE #85: fortio add a -quiet mode with less output
changing the log level doesn't work

not seeing how it's handled as an argument either?

https://github.com/istio/fortio/blob/master/fortio_main.go#L74

ldemailly:
Fixes #85

Also fixed a bug where `-json -` was sending some of the header to
stdout/mixed with json

- Fixed bug with echo-ing back POST data and added test for that
- DynamicHTTP now returns the mux so multiple handler can be set for
same path on different ports/servers. Also easier to write independent
tests.
- Fixed bug where fast client -H Foo:v1 -H Foo:v2 would loose v2 and
added test for that
- Updated README to document server modes and param a bit more (and
found the above bugs with echo etc)
- Fixed some test/logging data races with the global abort channel

Example:
```
$ fortio load -quiet -n 12  http://localhost:8080/
Fortio 0.6.4 running at 8 queries per second, 8->8 procs, for 12 calls: http://localhost:8080/
Aggregated Function Time : count 12 avg 0.00042976883 +/- 0.0002144 min 0.0001487 max 0.000732863 sum 0.005157226
# target 50% 0.000414229
# target 75% 0.000573546
# target 99% 0.00072649
# target 99.9% 0.000732226
Code 200 : 12 (100.0 %)
All done 12 calls (plus 0 warmup) 0.430 ms avg, 8.0 qps
```

~Ps: need to find how to change the log level without triggering race
test, for now added a new file to trigger quiet test~: solved


christian-posta (commented):
Trying this out locally for me looks good! 

```
ceposta@postamacpro(fortio (tweaks)) $ fortio load -c 1 -n 5  -quiet http://www.google.com
Fortio 0.6.4 running at 8 queries per second, 8->8 procs, for 5 calls: http://www.google.com
Aggregated Function Time : count 5 avg 0.08348766 +/- 0.01366 min 0.074955596 max 0.110622412 sum 0.417438298
# target 50% 0.0774778
# target 75% 0.0795796
# target 99% 0.110091
# target 99.9% 0.110569
Code 200 : 5 (100.0 %)
All done 5 calls (plus 0 warmup) 83.488 ms avg, 7.1 qps
```

ldemailly:
mandar/doug only one of you need to look :-) tia!

douglas-reid (commented):
/lgtm

PR #114, opened 2018-02-07
Adds Support for Configurable gRPC Listen Address

### LINKED ISSUE #113: Configurable fgrpc listen address
Similar to https://github.com/istio/fortio/issues/91 for fhttp, fgrpc needs a way to make it's listen address configurable.

danehans:
Fixes #113 

Converts `-grpc-port` flag from an int to a string representing a `host:port`, `ip:port` or `port`. When `port` is specified, fnet.NormalizePort is used to generate `:port` used by the grpc listener. Since grpcClient() uses the `-grpc-port` flag, the `GRPCDestination` function was added to construct the proper `grpcping` destination.


pingsrv.go
```diff
@@ -68,7 +70,7 @@ func pingServer(port int) {
 	healthServer.SetServingStatus("ping", grpc_health_v1.HealthCheckResponse_SERVING)
 	grpc_health_v1.RegisterHealthServer(grpcServer, healthServer)
 	fgrpc.RegisterPingServerServer(grpcServer, &pingSrv{})
-	fmt.Printf("Fortio %s grpc ping server listening on port %v\n", periodic.Version, port)
+	fmt.Printf("Fortio %s grpc ping server listening on port %s\n", periodic.Version, port)
```
ldemailly: (nit/no action needed:) you could have left that one as is :-)

pingsrv.go
```diff
@@ -163,3 +164,14 @@ func grpcClient() {
 		pingClientCall(dest, count, *payloadFlag)
 	}
 }
+
+func setGRPCDestination(dest string) string {
```
ldemailly: let's change this to GRPCDestination(dest string, port string)  // ie a pure function with no global access (it doesn't set anything and doesn't need to read anything)

that way you can add a test for it too

pingsrv.go
```diff
@@ -57,8 +58,9 @@ func (s *pingSrv) Ping(c context.Context, in *fgrpc.PingMessage) (*fgrpc.PingMes
 	return &out, nil
 }
 
-func pingServer(port int) {
-	socket, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+func pingServer(port string) {
+	port = fnet.NormalizePort(port)
+	socket, err := net.Listen("tcp", port)
```
ldemailly, line 63: 👍 ain't it nice we made that function ?

ldemailly:
let's try to get this in so the readme update can be done too and I can make 0.6.8

pingsrv.go
danehans: Oops... I'll revert.

pingsrv.go
danehans: OK

pingsrv.go
danehans: Good call indeed. I appreciate all your guidance.

pingsrv.go
```diff
@@ -149,11 +149,16 @@ func grpcHealthCheck(serverAddr string, svcname string, n int) {
 }
 
 func grpcClient() {
-	if len(flag.Args()) != 1 {
-		usage("Error: fortio grpcping needs host argument")
+	if len(flag.Args()) < 1 || len(flag.Args()) > 2 {
+		usage("Error: fortio grpcping needs either host or host and port arguments. ",
+			"Fortio's default gRPC server port (8079) is used if port is omitted.")
```
ldemailly: here it should be %v and *grpcPortFlag instead of 8079

also didn't we say I could put `host:port` (1 argument ? we can also do 2 args as extra flexibility if you want)

fnet/network.go
```diff
@@ -26,3 +27,19 @@ func NormalizePort(port string) string {
 	}
 	return ":" + port
 }
+
+// GRPCDestination parses dest and returns dest:port based on dest type
+// being a hostname, IPv4 or IPv6 address.
+func GRPCDestination(dest, port string) string {
```
ldemailly: Need to either move this in fgrpc/ or rename it to something more generic
`NormalizeHostPort()` maybe ?

also rename "port" param to "defaultPort" for clarity that it won't always be used

fnet/network_test.go
```diff
@@ -53,3 +53,42 @@ func TestNormalizePort(t *testing.T) {
...
+			"1.2.3.4",
+			"8078",
+			"1.2.3.4:8078",
+		},
+		{
+			"IPv6 address",
+			"2001:dba::1",
+			"8079",
+			"[2001:dba::1]:8079",
+		},
```
ldemailly: I would add a test
NormalizeHostPort([2001:dba::1]:1234,4567) -> [2001:dba::1]:1234 because there is already a port in 'host'
and
NormalizeHostPort(2001:dba::1234,4567) -> [2001:dba::1234]:4567 because the : is for ipv6 and not a port
and
NormalizeHostPort(foo.bar.com:777,888) -> foo.bar.com:777
NormalizeHostPort(foo.bar.com,888) -> foo.bar.com:888

fnet/network.go
```diff
@@ -26,3 +27,19 @@ func NormalizePort(port string) string {
...
+	if ip := net.ParseIP(dest); ip != nil {
```
ldemailly: I'm tempted to simplify this... and just again check for : (except if the : are inside [])
see the suggested additional tests below, I suggest to write the tests first and then change the code to pass

pingsrv.go
danehans: I'll update the error msg.

When you asked that GRPCDestination take string & port as args, I thought you were looking for separate host/port flags.

ldemailly: nah sorry i wasn’t clear, I meant the flag be the port if the port isn’t in the destination already

fnet/network_test.go
danehans: `NormalizeHostPort([2001:dba::1]:1234,4567) -> [2001:dba::1]:1234` will fail:
```
--- FAIL: TestNormalizeHostPort (0.00s)
	network_test.go:93: Test case IPv6 address and port failed to set gRPC destination
			expected: [2001:dba::1]:1234
			  actual: [2001:dba::1]:1234:5678
```

The func expects host to be an IP address and not an ip:port pair. Do you want the func to support this use case or should I update the test to support expected failures?

fnet/network.go
danehans: Will do. The func will need to be updated. IMO it's best to pass in a single dest string that can be host, host:port or ip:port.

pingsrv.go
danehans: I'm working through the changes and I think I'm about ready to puth the commit. I am sticking with the single `grpcping` flag representing a dest host. The dest can be `hostname`, `ip`, `ip:port` or `hostname:port`. If hostname/ip are specified, then `:*grpcDefaultPort` will be added.

fnet/network_test.go
ldemailly: Those tests should pass (which is why I mentioned to add the test and then fix the code)

I think again you can do a string based

danehans:
@ldemailly I refactored this PR based on your feedback and all tests pass. Let me know if anything is needed to get this merged.

fnet/network_test.go
ldemailly: can you add those 2:

NormalizeHostPort(foo.bar.com:777,888) -> foo.bar.com:777
NormalizeHostPort(foo.bar.com,888) -> foo.bar.com:888

fgrpc/grpcrunner_test.go
```diff
@@ -77,3 +77,53 @@ func TestGRPCRunner(t *testing.T) {
...
+		{
+			"IPv6 address",
+			"2001:dba::1",
+			"[2001:dba::1]:8079",
+		},
+		{
+			"IPv6 address and port",
+			"[2001:dba::1]:1234",
+			"[2001:dba::1]:1234",
+		},
```
ldemailly, line 116: can you add those 2:

GRPCDestination(foo.bar.com:777) -> foo.bar.com:777
GRPCDestination(foo.bar.com) -> foo.bar.com:8079

ldemailly (approved):
thx!

PR #123, opened 2018-02-10
switching fortio to using dev and vendor as a submodule

ldemailly:
Use dep, and submodule

Updated build image to latest go 1.8(.7) and to get grpc from vendor instead of image


ldemailly:
@costinm ptal, this should help demonstrating https://github.com/istio/istio/pull/3348 / what to do for istio/istio

costinm (approved):

PR #129, opened 2018-02-16
Converts -redirect-port flag to a string

danehans:
Previously, the `-redirect-port` flag was an int and therefore could not take `ip/host:port` for binding the service to a specific address. This PR converts the flag to a string in support of this requirement. In doing so, the value to disable the redirector (`-1`), was changed to a string named `disabled`.

Fixes https://github.com/istio/fortio/issues/122

### DIFF
--- a/fortio_main.go
+++ b/fortio_main.go
@@ -104,8 +104,8 @@ var (
 
 	allowInitialErrorsFlag = flag.Bool("allow-initial-errors", false, "Allow and don't abort on initial warmup errors")
 	autoSaveFlag           = flag.Bool("a", false, "Automatically save JSON result with filename based on labels & timestamp")
-	redirectFlag           = flag.Int("redirect-port", 8081,
-		"Redirect all incoming traffic to https URL (need ingress to work properly). -1 means off.")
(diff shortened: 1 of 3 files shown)


fortio_main.go
```diff
@@ -104,8 +104,8 @@ var (
 
 	allowInitialErrorsFlag = flag.Bool("allow-initial-errors", false, "Allow and don't abort on initial warmup errors")
 	autoSaveFlag           = flag.Bool("a", false, "Automatically save JSON result with filename based on labels & timestamp")
-	redirectFlag           = flag.Int("redirect-port", 8081,
-		"Redirect all incoming traffic to https URL (need ingress to work properly). -1 means off.")
+	redirectFlag           = flag.String("redirect-port", "8081", "Redirect all incoming traffic to https URL"+
+		" (need ingress to work properly). Can be in the form of host:port, ip:port, port or disabled to disable the feature.")
```
ldemailly: put \"disabled\" so the value stands out a bit more

ldemailly (approved):
ty

PR #136, opened 2018-02-18
Adding sha and proper dirty/pre label depending on git tag etc

### LINKED ISSUE #135: add sha and/or date to long version and debug url
because I change the version ahead of release - when the perf setup pulls latest for instance it's not 100% clear which version is running and if it's been built clean or not etc

ldemailly:
Fixes #135

dev build (go install .) look like this:
```
$ fortio version 
0.7.0-pre-unknown
$ fortio server
Fortio 0.7.0-pre echo server listening on port :8080
...
```

docker/release when dirty repo:
```
$ docker run istio/fortio:webtest version
0.7.0-pre-180218_2300_1dbde2b225b4c1bfdc6f051e782199e03414f8bd-dirty
```

clean repo but tag not matching
```
$ docker run istio/fortio:webtest version
0.7.0-pre-180218_2315_dea13ce8dc8c8ba3781d3a041ceb60a89af7ee24
$ docker run istio/fortio:webtest version -s
0.7.0-pre
```

releases should be `0.7.0` and full version `0.7.0-date_time_sha`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,10 @@ RUN make -C fortio submodule
 # Demonstrate moving the static directory outside of the go source tree and
 # the default data directory to a /var/lib/... volume
 RUN CGO_ENABLED=0 GOOS=linux go build -a -ldflags \
-  '-X istio.io/fortio/ui.resourcesDir=/usr/local/lib/fortio -X main.defaultDataDir=/var/lib/istio/fortio -s' \
+  "-s -X istio.io/fortio/ui.resourcesDir=/usr/local/lib/fortio -X main.defaultDataDir=/var/lib/istio/fortio \
+  -X istio.io/fortio/version.buildInfo=$(date +%y%m%d_%H%M_)$(cd fortio; git rev-parse HEAD) \
+  -X istio.io/fortio/version.tag=$(cd fortio; git describe --tags) \
+  -X istio.io/fortio/version.gitstatus=$(cd fortio; git status --porcelain | wc -l)" \
   -o fortio.bin istio.io/fortio
 # Just check it stays compiling on Windows (would need to set the rsrcDir too)
 RUN CGO_ENABLED=0 GOOS=windows go build -a -o fortio.exe istio.io/fortio
--- a/release/README.md
+++ b/release/README.md
@@ -14,6 +14,8 @@
 
 - The docker official builds are done automatically based on tag, check https://cloud.docker.com/app/istio/repository/docker/istio/fortio/builds
 
+- Increment the `patch` and commit that right away so the first point is true next time and so master/latest docker images have the correct next-pre version.
+
 ## How to change the build image
 
 Update [../Dockerfile.build](../Dockerfile.build)
--- a/release/README.md
+++ b/release/README.md
@@ -1,33 +1,36 @@
-# How to make a fortio release
+## How to make a fortio release
 
-- Make sure `periodic/periodic.go`'s `Version` is newer than https://github.com/istio/fortio/releases
+- Make sure `version/version.go`'s `major`/`minor`/`patch` is newer than https://github.com/istio/fortio/releases
 
 - Make a release there and document the changes since the previous release
 
-- Make sure to use the same tag format (e.g "0.4.3" - note that there are no "v" in the tag to be consistent with the rest istio)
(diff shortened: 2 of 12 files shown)


ldemailly:
dev build (go install .) look like this:
```
$ fortio version 
0.7.0-pre-unknown
$ fortio server
Fortio 0.7.0-pre echo server listening on port :8080
...
```

docker/release:
```
$ docker run istio/fortio:webtest version
0.7.0-pre-180218_2300_1dbde2b225b4c1bfdc6f051e782199e03414f8bd-dirty
```

release/README.md
```diff
@@ -1,33 +1,36 @@
-# How to make a fortio release
+## How to make a fortio release
 
-- Make sure `periodic/periodic.go`'s `Version` is newer than https://github.com/istio/fortio/releases
+- Make sure `version/version.go`'s `major`/`minor`/`patch` is newer than https://github.com/istio/fortio/releases
 
 - Make a release there and document the changes since the previous release
 
-- Make sure to use the same tag format (e.g "0.4.3" - note that there are no "v" in the tag to be consistent with the rest istio)
+- Make sure to use the same tag format (e.g "v0.7.1" - note that there is "v" in the tag unlike the rest of istio)
```
danehans: What is the reason for adding "v" to the tag?

ldemailly: it’s not new just updating the doc

the interesting question would be why istio doesn’t have a v - it’s pretty standard semver to have tags vX.Y.Z

it also easier to see all the version tag starting with a v and use completion

yutongz (approved):

Dockerfile
```diff
@@ -9,7 +9,10 @@ RUN make -C fortio submodule
 # Demonstrate moving the static directory outside of the go source tree and
 # the default data directory to a /var/lib/... volume
 RUN CGO_ENABLED=0 GOOS=linux go build -a -ldflags \
-  '-X istio.io/fortio/ui.resourcesDir=/usr/local/lib/fortio -X main.defaultDataDir=/var/lib/istio/fortio -s' \
+  "-s -X istio.io/fortio/ui.resourcesDir=/usr/local/lib/fortio -X main.defaultDataDir=/var/lib/istio/fortio \
+  -X istio.io/fortio/version.buildInfo=$(date +%y%m%d_%H%M_)$(cd fortio; git rev-parse HEAD) \
```
mattdelco, line 13: https://reproducible-builds.org/docs/timestamps/ would say to not include dates in the build.  An alternative is to include the date of the source, e.g.<pre>git log -1 --pretty=format:%cd --date=format:+%y%m%d_%H%M_</pre>

release/README.md
mattdelco: I don't see any mention of a v prefix in the semver doc, and don't know of anything else that uses it either (it's a bit redundant to say "version v0.7.1").  I don't see it used in docker tags (including for fortio).

Anyway, I haven't been clear about who/what is supposed to update the fortio tags in the istio repo (e.g., updateVersions.sh).    Most cases I see lazily use "istio/fortio:latest" (or equivalent), which is fine while on master but it's got potential to cause problems when someone is working with an older branch or release of istio.  Occasionally I come across files like samples/httpbin/sample-client/fortio-deploy.yaml that hard-code a specific version and I'm not sure if it should just be using "latest" like everyone else.

Dockerfile
ldemailly: I do want to know when the binary actually gets built because tag/version aren't supposed 
to be reused so if the date change it's interesting to know (ie it's a one time / one use thing)

The source tree TS is also interesting though but I expect people to copy paste the full sha if they really want to know what it is (and not helped by the version enough)

release/README.md
ldemailly: The version in the code is 0.7.0, likewise in the tgz and the docker tag but the github tag is "v0.7.0" maybe I need to clarify which tag is being talked about here (ie git tag)

Example of other projects using vX.Y.Z for their tags : 
https://github.com/grpc/grpc-go/releases
https://github.com/kubernetes/kubernetes/releases

Re updating tag in istio/istio - fortio ends up in vendor now, but that tag in istio.VERSION is used by some e2e test to install an image; :latest would work as long as I keep master pristine

PR #149, opened 2018-02-21
Fixes -H flag panic

danehans:
Updates headers flag to initialize the `httpOpts`. Also updates `Webtest.sh` script to inlcude `-H` tests to avoid future issues with the flag. Fixes Issue: https://github.com/istio/fortio/issues/148

### DIFF
--- a/Webtest.sh
+++ b/Webtest.sh
@@ -44,4 +44,7 @@ docker exec fortio_server /usr/local/bin/fortio load -stdclient -qps 1 -t 2s -c 
 docker exec fortio_server /usr/local/bin/fortio load -qps 1 -t 2s -c 2 http://www.google.com/
 # Do a grpcping
 docker exec fortio_server /usr/local/bin/fortio grpcping localhost
+# Test extra headers (-H) flag
+docker exec fortio_server /usr/local/bin/fortio load -H Foo:Bar -curl http://www.google.com
+docker exec fortio_server /usr/local/bin/fortio curl -H Foo:Bar -H Baz:Zab -curl http://www.google.com
 # TODO: check report mode
--- a/fortio_main.go
+++ b/fortio_main.go
@@ -41,11 +41,17 @@ var httpOpts fhttp.HTTPOptions
 type flagList struct {
 }
 
-// Unclear when/why this is called and necessary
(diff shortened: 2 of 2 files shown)


fortio_main.go
```diff
@@ -41,11 +41,17 @@ var httpOpts fhttp.HTTPOptions
...
+// The String method's output will be used in diagnostics.
 func (f *flagList) String() string {
 	return ""
 }
+
+// Set is the method to set the flag value, part of the flag.Value interface.
+// Set's argument is a string to be parsed to set the flag.
+// It's a comma-separated list, so we split it.
 func (f *flagList) Set(value string) error {
+	httpOpts.Init(value)
```
ldemailly, line 54: this is wrong location for the fix

please put init in http.go and a test in http_test.go
```
func TestGetHeaders(t *testing.T) {
	o := NewHTTPOptions("")
	o.AddAndValidateExtraHeader("FOo:baR")
```
replace first line by
```
        o := HTTPOptions{}
```

not counting that value is the header here while init takes a url...

Webtest.sh
```diff
@@ -44,4 +44,7 @@ docker exec fortio_server /usr/local/bin/fortio load -stdclient -qps 1 -t 2s -c
 docker exec fortio_server /usr/local/bin/fortio load -qps 1 -t 2s -c 2 http://www.google.com/
 # Do a grpcping
 docker exec fortio_server /usr/local/bin/fortio grpcping localhost
+# Test extra headers (-H) flag
+docker exec fortio_server /usr/local/bin/fortio load -H Foo:Bar -curl http://www.google.com
+docker exec fortio_server /usr/local/bin/fortio curl -H Foo:Bar -H Baz:Zab -curl http://www.google.com
```
ldemailly, line 49: It will be a great e2e test in addition to the unit test I mentioned before

but you don't need -curl for curl subcommand - it's also redundant - do one with load and one with curl
in theory you'd want to see the headers are set but that's what the unit test is for

ldemailly:
as this is quite critical and time sensitive, I'm making the fix - thanks a lot for the report though!

ldemailly:
thx! fixed in #150

PR #151, opened 2018-02-22
Add /debug/pprof except for report mode

ldemailly:
manual checked localhost:8080/debug/pprof/ in normal mode and absence of it in report mode

### DIFF
--- a/ui/uihandler.go
+++ b/ui/uihandler.go
@@ -897,6 +897,7 @@ func Serve(baseurl, port, debugpath, uipath, staticRsrcDir string, datadir strin
 // Report starts the browsing only UI server on the given port.
 // Similar to Serve with only the read only part.
 func Report(baseurl, port, staticRsrcDir string, datadir string) {
+	http.DefaultServeMux = http.NewServeMux() // drop the pprof default handlers
 	baseURL = baseurl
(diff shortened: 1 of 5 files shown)


ui/uihandler.go
```diff
@@ -897,6 +897,7 @@ func Serve(baseurl, port, debugpath, uipath, staticRsrcDir string, datadir strin
 // Report starts the browsing only UI server on the given port.
 // Similar to Serve with only the read only part.
 func Report(baseurl, port, staticRsrcDir string, datadir string) {
+	http.DefaultServeMux = http.NewServeMux() // drop the pprof default handlers
```
costinm, line 900: Not sure that's the way to do it - are you using the default mux before ? After ?
I would just use an explicit mux when creating the http listener.

costinm (approved):
/lgtm

It's not worse than before - but could be cleaner :-)

ui/uihandler.go
ldemailly: I tried the my own mux, like in RedirectToHTTPS below and somehow that still had pprof (maybe it tries to attach it to the first started server or some such) - searching around on the web I found that `http.DefaultServeMux = http.NewServeMux()` and tested it does work/removes pprof for the report and only the report

Issue #146, opened 2018-02-21
add grpc to the UI


ldemailly:
as a check box at least

ldemailly:
almost there, great start shipped in 0.7.2, left over #168

ldemailly:
it's quite good enough for a start - thank you !

PR #152, opened 2018-02-22
grpc support for UI

asadali:
- adds option to run grpc load test from the UI
- #TODO add more options for configuring the grpc run
- this PR may need some refactoring, but sending this out for feedback and course-correction

* * *

@ldemailly can you review? w.r.t #146 
thanks a lot for creating this tool!
it has proven quite useful in quick evaluation of grpc services.

### DIFF
--- a/Webtest.sh
+++ b/Webtest.sh
@@ -24,7 +24,7 @@ $CURL ${BASE_FORTIO}browse
 # Check we can connect, and run a http QPS test against ourselves through fetch
 $CURL "${BASE_FORTIO}fetch/localhost:8080$FORTIO_UI_PREFIX?url=http://localhost:8080/debug&load=Start&qps=-1&json=on" | grep ActualQPS
 # Check we can connect, and run a grpc QPS test against ourselves through fetch
-$CURL "${BASE_FORTIO}fetch/localhost:8080$FORTIO_UI_PREFIX?url=localhost:8079&load=Start&qps=-1&json=on&runner=grpc" | grep ActualQPS
+$CURL "${BASE_FORTIO}fetch/localhost:8080$FORTIO_UI_PREFIX?url=localhost:8079&load=Start&qps=-1&json=on&n=100&runner=grpc" | grep '"1": 100'
 # Check we get the logo (need to remove the CR from raw headers)
 VERSION=$(docker exec fortio_server /usr/local/bin/fortio version -s)
 LOGO_TYPE=$($CURL "${BASE_FORTIO}${VERSION}/static/img/logo.svg" | grep -i Content-Type: | tr -d '\r'| awk '{print $2}')
@@ -40,10 +40,12 @@ if [ "$SIZE" -lt 50000 ]; then
 fi
 # Check the main page
 $CURL $BASE_FORTIO
-# Do a small load using std client
+# Do a small http load using std client
 docker exec fortio_server /usr/local/bin/fortio load -stdclient -qps 1 -t 2s -c 1 https://www.google.com/
 # and with normal and with custom headers
 docker exec fortio_server /usr/local/bin/fortio load -H Foo:Bar -H Blah:Blah -qps 1 -t 2s -c 2 http://www.google.com/
 # Do a grpcping
 docker exec fortio_server /usr/local/bin/fortio grpcping localhost
+# Do a grpcping to a scheme-prefixed destination
+docker exec fortio_server /usr/local/bin/fortio grpcping https://fortio.istio.io:443
 # TODO: check report mode and pprof
--- a/Webtest.sh
+++ b/Webtest.sh
@@ -21,8 +21,10 @@ CURL="docker exec fortio_server /usr/local/bin/fortio load -curl -loglevel $LOGL
 $CURL https://istio.io/robots.txt
 # Check that browse doesn't 404s
 $CURL ${BASE_FORTIO}browse
-# Check we can connect, and run a QPS test against ourselves through fetch
+# Check we can connect, and run a http QPS test against ourselves through fetch
 $CURL "${BASE_FORTIO}fetch/localhost:8080$FORTIO_UI_PREFIX?url=http://localhost:8080/debug&load=Start&qps=-1&json=on" | grep ActualQPS
+# Check we can connect, and run a grpc QPS test against ourselves through fetch
+$CURL "${BASE_FORTIO}fetch/localhost:8080$FORTIO_UI_PREFIX?url=localhost:8079&load=Start&qps=-1&json=on&runner=grpc" | grep ActualQPS
 # Check we get the logo (need to remove the CR from raw headers)
 VERSION=$(docker exec fortio_server /usr/local/bin/fortio version -s)
 LOGO_TYPE=$($CURL "${BASE_FORTIO}${VERSION}/static/img/logo.svg" | grep -i Content-Type: | tr -d '\r'| awk '{print $2}')
--- a/fgrpc/grpcrunner.go
+++ b/fgrpc/grpcrunner.go
@@ -26,6 +26,8 @@ import (
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/health/grpc_health_v1"
 
+	"strings"
+
 	"istio.io/fortio/fnet"
 	"istio.io/fortio/log"
 	"istio.io/fortio/periodic"
@@ -34,15 +36,18 @@ import (
 const (
 	// DefaultGRPCPort is the Fortio gRPC server default port number.
 	DefaultGRPCPort = "8079"
+	prefixHTTP      = "http://"
+	prefixHTTPS     = "https://"
 )
 
 // Dial dials grpc either using insecure or using default tls setup.
 // TODO: option to specify certs.
 func Dial(serverAddr string, tls bool) (conn *grpc.ClientConn, err error) {
 	opts := grpc.WithInsecure()
-	if tls {
+	if tls || (strings.HasPrefix(serverAddr, "https://")) {
 		opts = grpc.WithTransportCredentials(credentials.NewTLS(nil))
 	}
+	serverAddr = grpcDestination(serverAddr)
 	conn, err = grpc.Dial(serverAddr, opts)
 	if err != nil {
 		log.Errf("failed to conect to %s with tls %v: %v", serverAddr, tls, err)
@@ -94,6 +99,7 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
 		RetCodes: make(map[grpc_health_v1.HealthCheckResponse_ServingStatus]int64),
 	}
 	grpcstate := make([]GRPCRunnerResults, numThreads)
+
 	for i := 0; i < numThreads; i++ {
 		r.Options().Runners[i] = &grpcstate[i]
 		conn, err := Dial(o.Destination, o.Secure)
@@ -156,9 +162,18 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
 	return &total, nil
 }
 
-// GRPCDestination parses dest and returns dest:port based on dest type
+// grpcDestination parses dest and returns dest:port based on dest type
 // being a hostname, IP address or hostname/ip:port pair.
-func GRPCDestination(dest string) string {
+func grpcDestination(dest string) string {
+	// strip any unintentional http/https scheme prefixes from destination
+	if strings.HasPrefix(dest, prefixHTTP) {
+		dest = strings.Replace(dest, prefixHTTP, "", 1)
+		log.Infof("stripping http scheme. grpc destination: %v", dest)
+	} else if strings.HasPrefix(dest, prefixHTTPS) {
+		dest = strings.Replace(dest, prefixHTTPS, "", 1)
+		log.Infof("stripping https scheme. grpc destination: %v", dest)
+	}
+
 	if _, _, err := net.SplitHostPort(dest); err == nil {
 		return dest
 	}
--- a/fgrpc/grpcrunner_test.go
+++ b/fgrpc/grpcrunner_test.go
@@ -126,6 +126,16 @@ func TestGRPCDestination(t *testing.T) {
 			"[2001:dba::1]:1234",
 		},
 		{
+			"http hostname and port",
+			"http://foo.bar:2567",
+			"foo.bar:2567",
+		},
+		{
+			"https hostname and port",
+			"https://foo.bar:2567",
+			"foo.bar:2567",
+		},
+		{
 			"hostname and no port",
 			"foo.bar.com",
 			"foo.bar.com:8079",
--- a/fgrpc/grpcrunner_test.go
+++ b/fgrpc/grpcrunner_test.go
@@ -96,6 +96,16 @@ func TestGRPCDestination(t *testing.T) {
 			"localhost:1234",
 		},
 		{
+			"http hostname and port",
+			"http://localhost:1234",
+			"localhost:1234",
+		},
+		{
+			"https hostname and port",
+			"https://localhost:1234",
+			"localhost:1234",
+		},
+		{
 			"IPv4 address",
 			"1.2.3.4",
 			"1.2.3.4:8079",
@@ -128,7 +138,7 @@ func TestGRPCDestination(t *testing.T) {
 	}
 
 	for _, tc := range tests {
-		dest := GRPCDestination(tc.dest)
+		dest := grpcDestination(tc.dest)
 		if dest != tc.output {
 			t.Errorf("Test case %s failed to set gRPC destination\n\texpected: %s\n\t  actual: %s",
 				tc.name,
--- a/ui/templates/main.html
+++ b/ui/templates/main.html
@@ -52,6 +52,9 @@ Headers: <br />
 {{end}}{{end}} <!-- 2 extra header lines, TODO: add a JS 'more headers' button -->
 <input type="text" name="H" size=40 value="" /> <br />
 <input type="text" name="H" size=40 value="" /> <br />
+http: <input type="radio" name="runner" value="http" checked/>,
+grpc: <input type="radio" name="runner" value="grpc"/><br/>
+Use secure transport (tls) for GRPC <input type="checkbox" name="grpc-secure"/><br/>
 JSON output: <input type="checkbox" name="json" />,
 Save output: <input type="checkbox" name="save" checked /> <br />
 <input type="submit" name="load" value="Start"/>
--- a/ui/uihandler.go
+++ b/ui/uihandler.go
@@ -37,6 +37,7 @@ import (
 	"sync"
 	"time"
 
+	"istio.io/fortio/fgrpc"
 	"istio.io/fortio/fhttp"
 	"istio.io/fortio/fnet"
 	"istio.io/fortio/log"
@@ -80,6 +81,7 @@ var (
 const (
 	fetchURI    = "fetch/"
 	faviconPath = "/favicon.ico"
+	modegrpc    = "grpc"
 )
 
 // Gets the resources directory from one of 3 sources:
@@ -140,13 +142,14 @@ func Handler(w http.ResponseWriter, r *http.Request) {
 	DoSave := (r.FormValue("save") == "on")
 	url := r.FormValue("url")
 	runid := int64(0)
+	runner := r.FormValue("runner")
 	if r.FormValue("load") == "Start" {
 		mode = run
 		if r.FormValue("json") == "on" {
 			JSONOnly = true
-			log.Infof("Starting JSON only load request from %v for %s", r.RemoteAddr, url)
+			log.Infof("Starting JSON only %s load request from %v for %s", runner, r.RemoteAddr, url)
 		} else {
-			log.Infof("Starting load request from %v for %s", r.RemoteAddr, url)
+			log.Infof("Starting %s load request from %v for %s", runner, r.RemoteAddr, url)
 		}
 	} else {
 		if r.FormValue("stop") == "Stop" {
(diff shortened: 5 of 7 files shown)


ui/uihandler.go
```diff
@@ -265,8 +275,53 @@ func Handler(w http.ResponseWriter, r *http.Request) {
 			}
 			uiRunMapMutex.Unlock()
 		}
-	case run:
-		// mode == run case:
+	case rungrpc:
```
ldemailly: more code should be shareable between the 2 cases

see how fortio_main.go does grpc/http with sharing the interface between the 2

ldemailly (commented):
you need to leave it backward compatible URL wise;
that's why the test is failing because with your PR one must set runner=http - make sure it's optional

also see below

and thanks for your contribution!
please sign the cla though

asadali:
@ldemailly thank you for the pointers. I refactored the shared code and handled the default case.

google cla: I signed it!

ui/uihandler.go
```diff
@@ -275,29 +278,36 @@ func Handler(w http.ResponseWriter, r *http.Request) {
...
+			o := fhttp.HTTPRunnerOptions{
+				HTTPOptions:        *httpopts,
+				RunnerOptions:      ro,
+				AllowInitialErrors: true,
+			}
+			res, err = fhttp.RunHTTPTest(&o)
+		}
 		if err != nil {
-			log.Errf("Init error %+v : %v", o, err)
+			log.Errf("Init error %+v : %v", ro, err)
```
asadali: we do lose some visibility here. but I did not think this merited creation of a `HasRunnerOptions` interface

ldemailly:
can you put just
I signed it!
nothing else in 1 comment (the cla bot is picky)

asadali:
I signed it!

asadali:
I am trying to work out why the bot doesn't seem to pick up the corporate CLA. will ping once I have addressed that

asadali:
I signed it

danehans:
I am trying to test the PR by doing a grpc run through the UI. From the UI, I change URL to `http://localhost:8079` and select the grpc radio button. I get the following Error:
```
20:17:29 I uihandler.go:437> UI: GET /fortio/?labels=Fortio&url=http%3A%2F%2Flocalhost%3A8079&qps=1000&t=3s&n=&c=8&p=50%2C+75%2C+90%2C+99%2C+99.9&r=0.0001&H=User-Agent%3A+istio%2Ffortio-0.7.2-pre&H=&H=&runner=grpc&save=on&load=Start HTTP/1.1 172.17.0.1:53864 ()
20:17:29 V uihandler.go:441> Header Connection: keep-alive
20:17:29 V uihandler.go:441> Header Accept-Encoding: gzip, deflate, br
20:17:29 V uihandler.go:441> Header Accept-Language: en-US,en;q=0.9
20:17:29 V uihandler.go:441> Header Upgrade-Insecure-Requests: 1
20:17:29 V uihandler.go:441> Header User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/63.0.3239.132 Safari/537.36
20:17:29 V uihandler.go:441> Header Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
20:17:29 V uihandler.go:441> Header Referer: http://localhost:8080/fortio/
20:17:29 I uihandler.go:152> Starting grpc load request from 172.17.0.1:53864 for http://localhost:8079
20:17:29 V stats.go:491> Will use [50 75 90 99 99.9] for percentiles
20:17:29 I uihandler.go:207> New run id 2
20:17:29 V http.go:95> URLSchemeCheck &{URL:http://localhost:8079 NumConnections:1 Compression:false DisableFastClient:false HTTP10:false DisableKeepAlive:false AllowHalfClose:false Insecure:false initDone:true https:false extraHeaders:map[User-Agent:[istio/fortio-0.7.2-pre]] hostOverride: HTTPReqTimeOut:15s}
20:17:29 V uihandler.go:278> adding header User-Agent: istio/fortio-0.7.2-pre
20:17:29 V http.go:175> Setting regular extra header User-Agent: istio/fortio-0.7.2-pre
20:17:29 D http.go:177> headers now map[User-Agent:[istio/fortio-0.7.2-pre]]
20:17:29 V http.go:175> Setting regular extra header X-On-Behalf-Of: 172.17.0.1:53864
20:17:29 D http.go:177> headers now map[User-Agent:[istio/fortio-0.7.2-pre] X-On-Behalf-Of:[172.17.0.1:53864]]
20:17:29 I grpcrunner.go:89> Starting grpc test for http://localhost:8079 with 8 threads at 1000.0 qps
20:17:29 V periodic.go:210> WATCHER 1 First outstanding run starting, catching signal
20:17:29 V periodic.go:216> WATCHER 1 starting new watcher for signal! chan  g 0xc42033c960 r 0xc4202d8ba0 (14)
20:17:29 E grpcrunner.go:112> Error in first grpc health check call for http://localhost:8079 rpc error: code = Unavailable desc = all SubConns are in TransientFailure
20:17:29 V periodic.go:253> Abort called for 0xc4202d2880 &{Runners:[0xc420304580 <nil> <nil> <nil> <nil> <nil> <nil> <nil>] QPS:1000 Duration:3s NumThreads:8 Percentiles:[50 75 90 99 99.9] Resolution:0.0001 Out:0xc4202d17c0 Labels:Fortio Stop:0xc4202d7b30 Exactly:0}
20:17:29 V periodic.go:81> Closing 0xc4202d8ba0
20:17:29 E uihandler.go:310> Init error {Runners:[0xc420304580 <nil> <nil> <nil> <nil> <nil> <nil> <nil>] QPS:1000 Duration:3s NumThreads:8 Percentiles:[50 75 90 99 99.9] Resolution:0.0001 Out:0xc4202d17c0 Labels:Fortio Stop:0xc4202d7b30 Exactly:0} : rpc error: code = Unavailable desc = all SubConns are in TransientFailure
20:17:29 V periodic.go:231> WATCHER 1 r.Stop readable
20:17:29 V periodic.go:234> WATCHER 1 End of go routine
20:17:29 V periodic.go:238> WATCHER 1 Last watcher: resetting signal handler
```

However, I am able to `grpcping` the same Fortio instance:
```
$ docker run --name client --rm istio/fortio:webtest grpcping 172.17.0.2:8079
20:18:56 I pingsrv.go:119> Ping RTT 1152190 (avg of 1427595, 969342, 1059635 ns) clock skew 2248
Clock skew histogram usec : count 1 avg 2.248 +/- 0 min 2.248 max 2.248 sum 2.248
>= 2.248 <= 2.248 , 2.248 , 100.00, 1
# target 50% 2.248
RTT histogram usec : count 3 avg 1152.1907 +/- 198.2 min 969.342 max 1427.595 sum 3456.572
>= 969.342 <= 1000 , 984.671 , 33.33, 1
> 1000 <= 1200 , 1100 , 66.67, 1
> 1400 <= 1427.6 , 1413.8 , 100.00, 1
# target 50% 1100
```

asadali:
@danehans can you try without the http prefix?
eg localhost:8079

i am not entirely sure if the http prefix should be honored for grpc calls?

Webtest.sh
```diff
@@ -21,8 +21,10 @@ CURL="docker exec fortio_server /usr/local/bin/fortio load -curl -loglevel $LOGL
 $CURL https://istio.io/robots.txt
 # Check that browse doesn't 404s
 $CURL ${BASE_FORTIO}browse
-# Check we can connect, and run a QPS test against ourselves through fetch
+# Check we can connect, and run a http QPS test against ourselves through fetch
 $CURL "${BASE_FORTIO}fetch/localhost:8080$FORTIO_UI_PREFIX?url=http://localhost:8080/debug&load=Start&qps=-1&json=on" | grep ActualQPS
+# Check we can connect, and run a grpc QPS test against ourselves through fetch
+$CURL "${BASE_FORTIO}fetch/localhost:8080$FORTIO_UI_PREFIX?url=localhost:8079&load=Start&qps=-1&json=on&runner=grpc" | grep ActualQPS
```
asadali: @danehans you raised a good point about the intended usage of grpc mode. I added an acceptance test here. let me know if this helps

ldemailly:
we could be nice and strip the http/https prefix for grpc
and I suppose even nicer would be to have https prefix imply --grpc-secure
but that can be in follow up pr

speaking of which can you add a checkbox for the Secure option of grpc ?

ldemailly:
@danehans the error is definitely obscure, so good find!

asadali:
added a grpc secure option

not totally convinced about the automatic translation of http/https to equivalent grpc secure/insecure mode. 
should we fail fast or build tolerations into the system?

danehans:
I removed `http://` from the url and it worked. __Note:__ After the run completes, the diagram adds the `http://` in front of `localhost:8079`.

danehans:
Here is a screenshot of the diagram showing http://localhost:8079"

screen shot 2018-02-23 at 3 47 44 pm

asadali:
tl:dr; fixing the title will require a larger refactoring of RunnerResults and its derived structs

the value for the diagram's title is determined [here](https://github.com/istio/fortio/blob/master/ui/static/js/fortio_chart.js#L90)

the GRPCRunnerResults do not contain a URL json field, instead they contain Destination. However, Destination isn't currently written out to the result file for a grpc result save so the value for the title should show up as undefined. I am not sure how the `http://localhost:8079` value got populated.

ldemailly:
fail fast is ok I guess - the message is obscure and delayed, I would expect the Dial to fail because http://x isn't a valid hostname

ui/uihandler.go
```diff
@@ -275,29 +279,37 @@ func Handler(w http.ResponseWriter, r *http.Request) {
...
 		if !JSONOnly {
 			flusher.Flush()
 		}
-		res, err := fhttp.RunHTTPTest(&o)
-		uiRunMapMutex.Lock()
-		delete(runs, runid)
-		uiRunMapMutex.Unlock()
+		var res periodic.HasRunnerResult
+		var err error
+		if runner == modegrpc {
```
ldemailly, line 296: 👍

Webtest.sh
ldemailly: maybe grep for something GRPC specific like `Health SERVING`

asadali:
understood. do you want the validation to be added as part of this PR?

ui/templates/main.html
```diff
@@ -52,6 +52,9 @@ <h1>Φορτίο (fortio) v{{.Version}}{{if not .DoLoad}} control UI{{end}}</h1>
 {{end}}{{end}} <!-- 2 extra header lines, TODO: add a JS 'more headers' button -->
 <input type="text" name="H" size=40 value="" /> <br />
 <input type="text" name="H" size=40 value="" /> <br />
+http: <input type="radio" name="runner" value="http" checked/>,
+grpc: <input type="radio" name="runner" value="grpc"/><br/>
+Use secure transport (tls) for GRPC <input type="checkbox" name="grpc-secure"/><br/>
```
ldemailly, line 57: thanks for having used the same url param name as the flag, as is attempted for most of the other options

ldemailly:
the title logic is here: https://github.com/istio/fortio/blob/master/ui/static/js/fortio_chart.js#L90

edit: you already found this :-)

Webtest.sh
asadali: the grep is on the result json object which is currently shared between the http/grpc modes. the result object does not have anything grpc-specific for now.
`Health SERVING` is printed to stdout. not to the json file. correct me if I understood this wrong.

ui/templates/main.html
asadali: yes, that was my intention. however, i did deviate a bit for the radio buttons. :|

ldemailly:
yeah let's just add something in fgrpc/ Dial() to bail early with a nice message (though maybe just dropping http:// and https:// again would be nicer

it's like when a compiler tells you "you forgot a ; on line 23" it would be nice to just assume it's there and continue with a warning instead (for compilers you can make the case it's ok to abort early to avoid doing completely wrong guesses but in this case it's pretty clear what to guess)

fgrpc/grpcrunner.go
```diff
@@ -29,6 +29,8 @@ import (
 	"istio.io/fortio/fnet"
 	"istio.io/fortio/log"
 	"istio.io/fortio/periodic"
+	"strings"
+	"regexp"
```
ldemailly: let’s not drag in regexp for a prefix match?

fgrpc/grpcrunner_test.go
```diff
@@ -95,6 +95,16 @@ func TestGRPCDestination(t *testing.T) {
 			"localhost:1234",
 			"localhost:1234",
 		},
+		{
+			"http hostname and port",
+			"http://localhost:1234",
+			"localhost:1234",
```
ldemailly, line 101: +1

fgrpc/grpcrunner_test.go
```diff
@@ -95,6 +95,16 @@ func TestGRPCDestination(t *testing.T) {
...
+		},
+		{
+			"https hostname and port",
+			"https://localhost:1234",
+			"localhost:1234",
```
ldemailly, line 106: can we check grpc secure state ?

fgrpc/grpcrunner.go
```diff
@@ -156,9 +160,16 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
 	return &total, nil
 }
 
-// GRPCDestination parses dest and returns dest:port based on dest type
+// parseGRPCDestination parses dest and returns dest:port based on dest type
```
ldemailly: why the name/visibility  change ?

asadali: since its not being used outside the package anymore and its being embedded into the Dial method

fgrpc/grpcrunner.go
asadali: ack. it felt like a hammer. thanks for calling it out. removed.

fgrpc/grpcrunner_test.go
asadali: that is not being set in the parseGRPCDestination function. It's tied in with the fgrpc.Dial function.

```
if tls || (strings.HasPrefix(serverAddr, "https://")) {
		opts = grpc.WithTransportCredentials(credentials.NewTLS(nil))
	}
```

not sure how to best test it.

asadali:
@ldemailly let me know if this needs any more work. rebased with current master

fgrpc/grpcrunner_test.go
ldemailly: you can test it (at least manually, and probably also in Webtest.sh, with
fortio grpcping https://fortio.istio.io/
(it will need to set the port to 443 too ideally if not supported)

maybe we should just bail if it has http[s]:// prefix

fgrpc/grpcrunner.go
ldemailly: go style would probably be to call this grpcDestination() then but it's ok

Webtest.sh
ldemailly: k can you file a follow up that the GRPC json is missing the equivalent of the Http code map that http has
(don't need to fix everything in this 1 pr if you don't have time)

fgrpc/grpcrunner.go
asadali: changed to `grpcDestination`

fgrpc/grpcrunner_test.go
asadali: sorry, was unclear earlier. i had tested it manually and the https state led to secureTLS internal state. In any case, I added an additional grpcping in webtest.sh

Webtest.sh
asadali: not sure that I understand this. the grpc object does have a retcode map object eg from a sample run of 3000 queries on `localhost:8079`
```
  "RetCodes": {
    "1": 3000
  }
```

however, the json doesn't contain a grpc-specific message eg `Health SERVING`, this is logged to stdout, hence can't be grep-ed for in the json result.

are you suggesting that the result.json should contain http/grpc specific messages?

ldemailly: so the code printing stdout is
		fmt.Printf("Health %s : %d\n", k.String(), total.RetCodes[k])

for some reason the Json has the numerical value "1" instead of "SERVING" - I wonder if that's fixable easily with some annotation on the struct

but we can postpone - I guess just grep for "1": 100 and use exactly 100 (-n 100)

Webtest.sh
```diff
@@ -38,10 +40,12 @@ if [ "$SIZE" -lt 50000 ]; then
...
 $CURL $BASE_FORTIO
-# Do a small load using std client
+# Do a small http load using std client
 docker exec fortio_server /usr/local/bin/fortio load -stdclient -qps 1 -t 2s -c 1 https://www.google.com/
 # and with normal and with custom headers
 docker exec fortio_server /usr/local/bin/fortio load -H Foo:Bar -H Blah:Blah -qps 1 -t 2s -c 2 http://www.google.com/
 # Do a grpcping
 docker exec fortio_server /usr/local/bin/fortio grpcping localhost
+# Do a grpcping to a scheme-prefixed destination
+docker exec fortio_server /usr/local/bin/fortio grpcping https://fortio.istio.io:443
```
ldemailly, line 50: 👍 can we make it work without :443 :-) ? [it's ok as, we can do that later]

fgrpc/grpcrunner.go
```diff
@@ -156,9 +160,15 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
...
 
-// GRPCDestination parses dest and returns dest:port based on dest type
+// grpcDestination parses dest and returns dest:port based on dest type
 // being a hostname, IP address or hostname/ip:port pair.
-func GRPCDestination(dest string) string {
+func grpcDestination(dest string) string {
+	// strip any unintentional http/https scheme prefixes from destination
+	if strings.HasPrefix(dest, "http://") || strings.HasPrefix(dest, "https://") {
+		dest = strings.Replace(strings.Replace(dest, "https://", "", 1), "http://", "", 1)
+		log.Warnf("stripping http/https scheme. grpc destination: %v", dest)
```
ldemailly: it's a feature so probably Infof or LogVf

fgrpc/grpcrunner.go
```diff
@@ -156,9 +160,15 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
...
 }
 
-// GRPCDestination parses dest and returns dest:port based on dest type
+// grpcDestination parses dest and returns dest:port based on dest type
 // being a hostname, IP address or hostname/ip:port pair.
-func GRPCDestination(dest string) string {
+func grpcDestination(dest string) string {
+	// strip any unintentional http/https scheme prefixes from destination
+	if strings.HasPrefix(dest, "http://") || strings.HasPrefix(dest, "https://") {
+		dest = strings.Replace(strings.Replace(dest, "https://", "", 1), "http://", "", 1)
```
ldemailly: this is a bit obscure, split the if ?

ui/uihandler.go
ldemailly: it's ok, though maybe add the input url and runner var

ldemailly (commented):
very nice, thanks for all of this, see below for minor things

Webtest.sh
asadali: ah, now i get it! will give it a shot

Webtest.sh
asadali: you mean, be intelligent and assume 443 on a https prefix?

Webtest.sh
asadali: #161

asadali: not sure how the automation works on this repo? will referencing the issue auto-close it if PR is merged?

fgrpc/grpcrunner.go
asadali: done

fgrpc/grpcrunner.go
asadali: changed to infof

ui/uihandler.go
```diff
@@ -275,29 +279,37 @@ func Handler(w http.ResponseWriter, r *http.Request) {
...
+			o := fhttp.HTTPRunnerOptions{
+				HTTPOptions:        *httpopts,
+				RunnerOptions:      ro,
+				AllowInitialErrors: true,
+			}
+			res, err = fhttp.RunHTTPTest(&o)
+		}
 		if err != nil {
-			log.Errf("Init error %+v : %v", o, err)
+			log.Errf("Init error for %s mode with url %s and options %+v : %v", runner, url, ro, err)
```
asadali, line 312: enhanced the error log message

Webtest.sh
ldemailly: yes if no :port is specified but http:// prefix is there, use 80, if https:// is there, use 443
but we can do that in followup

ldemailly:
pro tip:

please never force push a pr once reviewing has started

do a pull instead of a merge/rebase to be in sync

fgrpc/grpcrunner_test.go
```diff
@@ -95,6 +95,16 @@ func TestGRPCDestination(t *testing.T) {
...
+		},
 		{
 			"IPv4 address",
 			"1.2.3.4",
```
ldemailly, line 110: can you add tests for http:// and https://foo.bar:2567 -> foo.bar:2567 ?

ldemailly (commented):
just add 2 tests in existing grpcDestination and we'll merge this ! thanks for your patience!

fgrpc/grpcrunner_test.go
asadali: done

asadali:
sorry about the inconvenience with the force pushes. i guess i went overboard with trying to keep commit history clean

Webtest.sh
asadali: opened 162 to track that

ldemailly (approved):

ldemailly:
it gets squashed at the end on master so the PR can have 30 commits it's no problems

PR #153, opened 2018-02-22
Fix #143 : sort the headers alphabetically for `fortio server` /debug endpoint

### LINKED ISSUE #143: fortio /debug : sort the headers
would make it easier to find

SeriyBg:
fixes #143

### DIFF
--- a/fhttp/http.go
+++ b/fhttp/http.go
@@ -35,6 +35,8 @@ import (
 	// get /debug/pprof endpoints on default server - make sure report only doesn't have it
 	_ "net/http/pprof"
 
+	"sort"
+
 	"istio.io/fortio/fnet"
 	"istio.io/fortio/log"
 	"istio.io/fortio/stats"
@@ -1232,11 +1234,18 @@ func DebugHandler(w http.ResponseWriter, r *http.Request) {
 	// Host is removed from headers map and put here (!)
 	buf.WriteString("Host: ")
 	buf.WriteString(r.Host)
-	for name, headers := range r.Header {
+
+	var keys []string
+	for k := range r.Header {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, name := range keys {
 		buf.WriteByte('\n')
 		buf.WriteString(name)
 		buf.WriteString(": ")
 		first := true
+		headers := r.Header[name]
 		for _, h := range headers {
 			if !first {
 				buf.WriteByte(',')
--- a/fhttp/http.go
+++ b/fhttp/http.go
@@ -27,6 +27,7 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


fhttp/http.go
```diff
@@ -35,6 +35,8 @@ import (
 	// get /debug/pprof endpoints on default server - make sure report only doesn't have it
 	_ "net/http/pprof"
 
+	"sort"
```
ldemailly: please put that import with the others (should be sorted automatically bu go fmt)

ldemailly (commented):
Thanks ! small nit:

ldemailly:
also put “fixes #143” in the PR description

danehans:
I performed the following test:
```
$ docker run --name client --rm istio/fortio:webtest curl -loglevel verbose \
-H Zab:Baz -H Foo:Bar -H Baz:Barf -curl http://172.17.0.2:8080/debug

<SNIP>
headers:

Host: 172.17.0.2:8080
Baz: Barf
Foo: Bar
User-Agent: istio/fortio-0.7.2-pre
Zab: Baz

<SNIP>
```

@ldemailly are you looking for only the extra headers to be sorted alphabetically or all non-Host headers?

ldemailly:
Host is special so I don't mind we put it first

what do you mean "extra headers" vs "all non Host headers" ? what is the difference?

danehans:
I should have stated "additional headers", i.e. `-H` flag instead of extra headers.

fhttp/http.go
SeriyBg: Done

SeriyBg:
@ldemailly thanks for your comments! I have fixed it

ldemailly (dismissed):
thank you for your contribution

ideally there would be a unit test for this - do you want to add one ? (see http_test.go)
just add 2 headers A: x, B:y and make sure you get A before B ? maybe run the the check 3 times to ensure you don't get lucky, or add more headers to make it less likely to accidentally pass
(ideally comment out your new code, see the test fail, put back the fix, see the test pass)

SeriyBg:
@ldemailly sure will add the test. 
Just a quick question: is there any mock framework used to build the tests? I actually need to have a `mockResponseWriter` so I can test the `DebugHandler`. Or should I write the test on the upper level? If so, then maybe the test must be located in `httprunner_test`?

ldemailly:
You use a dynamichttp see the tests in http_test that start a server and connect to it. Early tests using the server were in httprunner test but new test should be in http test

So not a mock, a real request and response (kind of e2e test in unit for)

ldemailly:
an example of test/where to add:
https://github.com/istio/fortio/blob/master/fhttp/http_test.go#L564

SeriyBg:
@ldemailly thanx for your help. I have added the test for this issue, you can take a look when you have time.

ldemailly (approved):
awesome ! thanks !

PR #155, opened 2018-02-24
Port 0 support is added to fhttp serve

olimpias:
Todo is done "// TODO: make it work for port 0 and return the port found and also
// add a non blocking mode that makes sure the socket exists before returning"

### DIFF
--- a/fhttp/http.go
+++ b/fhttp/http.go
@@ -1268,17 +1268,24 @@ func DebugHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 // Serve starts a debug / echo http server on the given port.
-// TODO: make it work for port 0 and return the port found and also
-// add a non blocking mode that makes sure the socket exists before returning
-func Serve(port, debugPath string) {
+// Returns the port that opened for listing socket
+func Serve(port, debugPath string) string {
 	startTime = time.Now()
 	nPort := fnet.NormalizePort(port)
+	listener, err := net.Listen("tcp", nPort)
+	if err != nil {
+		log.Fatalf("Error occurred while listening %v", err)
+	}
+	nPort = strconv.Itoa(listener.Addr().(*net.TCPAddr).Port)
 	fmt.Printf("Fortio %s echo server listening on port %s\n", version.Short(), nPort)
-	if debugPath != "" {
-		http.HandleFunc(debugPath, DebugHandler)
-	}
-	http.HandleFunc("/", EchoHandler)
-	if err := http.ListenAndServe(nPort, nil); err != nil {
-		fmt.Println("Error starting server", err)
-	}
+	go func() {
+		if debugPath != "" {
+			http.HandleFunc(debugPath, DebugHandler)
+		}
+		http.HandleFunc("/", EchoHandler)
+		if err := http.Serve(listener, nil); err != nil {
+			fmt.Println("Error starting server", err)
+		}
+	}()
+	return nPort
 }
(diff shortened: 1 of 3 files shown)


fhttp/http.go
```diff
@@ -1268,17 +1268,24 @@ func DebugHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 // Serve starts a debug / echo http server on the given port.
-// TODO: make it work for port 0 and return the port found and also
-// add a non blocking mode that makes sure the socket exists before returning
-func Serve(port, debugPath string) {
+// Returns the port that opened for listing socket
+func Serve(port, debugPath string) string {
```
ldemailly: let's return the port as a number (why not?)

also maybe a follow up is to refactor common parts with  DynamicHTTPServer() above
https://github.com/istio/fortio/blob/master/fhttp/http.go#L1132 [but let's postpone that]

do add a test/exercise the return value though

fhttp/http.go
```diff
@@ -1268,17 +1268,24 @@ func DebugHandler(w http.ResponseWriter, r *http.Request) {
...
-	fmt.Printf("Fortio %s echo server listening on port %s\n", version.Short(), nPort)
-	if debugPath != "" {
-		http.HandleFunc(debugPath, DebugHandler)
-	}
-	http.HandleFunc("/", EchoHandler)
-	if err := http.ListenAndServe(nPort, nil); err != nil {
-		fmt.Println("Error starting server", err)
+	listener, err := net.Listen("tcp", nPort)
+	if err != nil {
+		log.Fatalf("Error occurred while listening %v", err)
```
ldemailly: while listening on %v: %v", nPort, err

ldemailly (commented):
this is great, 2 minor stuff below

fhttp/httprunner_test.go
```diff
@@ -116,17 +115,13 @@ func TestHTTPRunnerBadServer(t *testing.T) {
 // the error test for / url above fail:
 
 func TestServe(t *testing.T) {
-	listener, err := net.Listen("tcp", ":0") // nolint: gas
-	if err != nil {
-		log.Fatalf("Unable to listen to dynamic port: %v", err)
+	port := 0
+	outputPort := Serve(strconv.Itoa(port), "/debugx1")
```
ldemailly: nit:
```
port := Serve("0", "/debugx1")
```
would be simpler and also minimize the diff

fhttp/httprunner_test.go
```diff
@@ -116,17 +115,13 @@ func TestHTTPRunnerBadServer(t *testing.T) {
...
+	log.Infof("Using port: %d", outputPort)
+	url := fmt.Sprintf("http://localhost:%d/debugx1?env=dump", outputPort)
+	if port == outputPort {
```
ldemailly: port == 0

ldemailly (commented):
nice, almost done ! can you rebase and apply the following minor tweaks

ldemailly (approved):
awesome!

ldemailly:
hmm I expected you tested it with the command line... we broke what @danehans did earlier:

```
$ fortio server --http-port 0
Https redirector running on :8081
UI starting - visit:
http://localhost:0/fortio/   (or any host/ip reachable on this server)
Fortio 0.7.2-pre grpc ping server listening on port :8079
Fortio 0.7.2-pre echo server listening on port 52769
```

ldemailly:
can you follow up to fix it soon or I should ? (need to use the port returned to do the cosmetic line:
```
http://localhost:0/fortio/   (or any host/ip reachable on this server)
```

olimpias:
I can enter a fix for it.

ldemailly:
fixing most of it in https://github.com/istio/fortio/pull/156/files#diff-8180646bc61c297cbb3d32d9d58350b4

PR #169, opened 2018-03-01
addition of `latest_release`

ldemailly:
addition of `latest_release` git and docker hub tags to the release steps

### DIFF
--- a/release/README.md
+++ b/release/README.md
@@ -16,6 +16,8 @@
 
 - Increment the `patch` and commit that right away so the first point is true next time and so master/latest docker images have the correct next-pre version.
 
+- Once the release is deemed good/stable: move the git tag `latest_release` to the same as the release. Also push `latest_release` docker tag/image.
+
 ## How to change the build image
 
 Update [../Dockerfile.build](../Dockerfile.build)


mattdelco:
Based on how I've seen "latest" used for other images (mostly the 3rd party ones consumed by istio) I assumed that "latest" tended to already imply stable (at least when it comes to public hubs like docker.io -- I've seen how "docker build" and bazel default to "latest" though they're also just storing the images on the local system's registry rather than a hub on another system).

Anyway, I don't have a strong preference about whether latest doesn't imply stable (i.e., this change), vs the opposite assumption (e.g., initially use something like "latest-rc" or "latest-dev" and later retag with "latest:), but if we do go with what's proposed in this PR then has someone already signed up to update the various existing uses of "fortio:latest" in test apps/scripts with "fortio:latest_release"?

ldemailly:
what happens right now is every build on master that passes the tests automatically turns into an updated docker image (latest) it's convenient for dev/testing but getting a bit risky for wide use in istio

this matches https://github.com/istio/istio/pull/3859

(though I'm very open to a better way - I considered making a 0.7 floating tag but figured 0.8 will probably also be fine and not want to change it again)

also it was bugging me that I changed the version to have "clean" build detection but latest already has "-pre"

mattdelco:
I looked though the Dockefile's very quickly but that didn't help me to get an immediate sense of the breadth of the dependencies for building the images nor the time it typically takes.  I'm toying with the idea of whether fortio could be like proxy, whereby there's a prow postsubmit that generates an image and updates a sha in istio.VERSION.  I suppose this would only work if there was sufficient automated tests that could be run to verify a build was stable, otherwise istio's master would always get the latest fortio.  Offhand I'm not sure if there'd be any advantage to be like how proxy now builds as of this week (whereby a tar of binaries is consumed by istio/istio and a local docker image is created).

Anyway, is the overall motivation to not have to update FORTIO_TAG each time there's a monthly release of istio?  E.g., istio 0.8 still might essentially use fortio 0.7 if there's been no changes to fortio.  I'm not sure if the idea is that during the 0.8 release it might look bad or confusing to have a stale 0.7 for FORTIO_TAG in istio.VERSION.  Using "latest" or "latest_release" will hide the fact that the 0.7 might still be used for istio 0.8, but I'm not sure if that's an overall win.  I've also been hesitant to make additional bets on long-term compatibility.  I know it's a goal for certain parts of Istio, but what I've observed seems more like specific individuals being concerned about compatibility than a widespread effort (though my baseline for comparison is a product I worked on that basically had to keep checkpoints of variables in C code compatible for at least a decade).

ldemailly:
so far fortio has had about 5x the release rate than istio and many x less bugs, so it’s not about being behind

it’s just about not automatically getting master - and for all users of fortio, not just istio unit tests, to have a moving yet blessed tag to use in yamls that match a release (eg 0.7.2) instead of master (0.7.3-pre)

mattdelco (approved):
It seems like another instance in Istio where getting latest is desirable but, on the other hand (at least for some people) wanting some amount of reproducibility (e.g., I think we still have issues with some tests failing on patch builds of monthly release because some tests are pulling from current/latest source).  The update bot has been the most common so far in Istio.  I know it's not quite suitable here since the desire is latest stable, and designation of stable is a manual act.  So, it seems like it gets down to whether the manual act is with updating some sha reference in a file, or updating the tags on a docker hub.  I figured the former was better suited for auditing of current behavior and tracking how/when things changes.

Anyway, both schemes are workable. 
/lgtm
(mostly because the other changes went in so might as well update the docs to be consistent), though I realize there's probably no bots running on this repo.

ldemailly:
> Anyway, both schemes are workable.
/lgtm

thanks!

>  there's probably no bots running on this repo.

there aren't - trying to use just docker swarm auto builds - and keep it that way if at all possible
though I do want probably some automation on patch release auto bump

Issue #170, opened 2018-03-02
add a &close option to echo server


ldemailly:
to reproduce problems with http 1.0 like services or basic 1.1 without connection reuse

would allow to isolate if https://github.com/istio/istio/issues/3295 are related to 2 hops vs connection close

ldemailly:
done without %

Issue #168, opened 2018-03-01
fortio UI for grpc results is misleading/incomplete


ldemailly:
Follow up from #152 

- results are missing
- shows "100% error" all the time
- shows undefined as the url

![screen shot 2018-02-28 at 9 56 28 pm](https://user-images.githubusercontent.com/3664595/36829084-44647e5a-1cd2-11e8-9027-7cf348175364.png)

asadali:
ack. I had noticed the undefined string but failed to observe the 100% errors part. I will work on this.

what do you mean by results are missing?

ldemailly:
thanks. the codes map doesn’t show on the page during the run (probably goes to stdout instead of out)

ldemailly:
I'm fixing that last part in #173

Issue #165, opened 2018-02-28
add socket closed/opened count to fast http client


ldemailly:
so we can detect unexpected non reuse of socket

olimpias:
May I work on this?

ldemailly:
fixed in #173

PR #173, opened 2018-03-03
Multiple improvements/fixes

### LINKED ISSUE #172: unexpected init with empty url when using -H
$ fortio curl -H Host:httpbin  http://35.193.254.22/headers
17:58:00 E http.go:100> unexpected init with empty url

ldemailly:
- add both link to UI and Json data from run results and browse UI And a back to top link

- Fix #172 (error logged when using -H)

- Initial version of #170  &close option for echo server

- Build with both go1.8 and 1.10; put 1.10 binaries in docker images/release

- Add StdClient option for the web UI runs

- Misc cleanups to front page (grpc secure option)

- Add socket use cout (#165)

- don't consider early EOF or other an error (fix 50% errors for some call client closes - like /fortio/fetch/ does)

- Reminder of fix for #168 show grpc result codes in UI/text

### DIFF
--- a/fhttp/http.go
+++ b/fhttp/http.go
@@ -618,7 +618,7 @@ func (c *FastClient) connect() *net.TCPConn {
 
 // Extra error codes outside of the HTTP Status code ranges. ie negative.
 const (
-	// SocketError is return when a transport error occured: unexpected EOF, connection error, etc...
+	// SocketError is return when a transport error occurred: unexpected EOF, connection error, etc...
 	SocketError = -1
 	// RetryOnce is used internally as an error code to allow 1 retry for bad socket reuse.
 	RetryOnce = -2
--- a/fhttp/http.go
+++ b/fhttp/http.go
@@ -1113,6 +1113,9 @@ func EchoHandler(w http.ResponseWriter, r *http.Request) {
 			}
 		}
 	}
+	if r.FormValue("close") != "" {
+		w.Header().Set("Connection", "close")
+	}
 	size := generateSize(r.FormValue("size"))
 	if size >= 0 {
 		log.Errf("Writing %d size with %d status", size, status)
--- a/fhttp/http.go
+++ b/fhttp/http.go
@@ -616,9 +616,17 @@ func (c *FastClient) connect() *net.TCPConn {
 	return socket
 }
 
+// Extra error codes outside of the HTTP Status code ranges. ie negative.
+const (
+	// SocketError is return when a transport error occured: unexpected EOF, connection error, etc...
+	SocketError = -1
+	// RetryOnce is used internally as an error code to allow 1 retry for bad socket reuse.
+	RetryOnce = -2
+)
+
 // Fetch fetches the url content. Returns http code, data, offset of body.
 func (c *FastClient) Fetch() (int, []byte, int) {
-	c.code = -1
+	c.code = SocketError
 	c.size = 0
 	c.headerLen = 0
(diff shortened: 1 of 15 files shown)


fhttp/http.go
```diff
@@ -345,7 +356,7 @@ type FastClient struct {
 
 // Close cleans up any resources used by FastClient
 func (c *FastClient) Close() {
-	log.Debugf("Closing %+v", c)
+	log.Debugf("Closing %p %s", c, c.url)
```
ldemailly: fixes log flood where the buffer was logged in debug mode

fhttp/http.go
```diff
@@ -1102,6 +1113,9 @@ func EchoHandler(w http.ResponseWriter, r *http.Request) {
 			}
 		}
 	}
+	if r.FormValue("close") != "" {
+		w.Header().Set("Connection", "close")
```
ldemailly: this does close the connection, not just sets the header

ui/templates/browse.html
```diff
@@ -128,5 +128,6 @@ <h1>Φορτίο (fortio) v{{.Version}}{{.Extra}}</h1>
 fortio_load(files.value)
 </script>
 {{end}}
+<p>Go to <a href='./'>Top</a>.</p>
```
asadali, line 131: this link resolves to `<host-path>:8080/fortio/`

so, if used from the browse page, the link redirects to the fortio landing page rather than the top of the `/fortio/browse/` page.

if that is intentional, then rewording this to `Go Back` makes more sense.

else replace `href='./'` with `href='#'`

suggetion: the heading `<h1>Φορτίο (fortio) v{{.Version}}{{.Extra}}</h1>` can be enhanced as a link to the landing page

fhttp/httprunner.go
```diff
@@ -142,6 +143,7 @@ func RunHTTPTest(o *HTTPRunnerOptions) (*HTTPRunnerResults, error) {
 	}
 	sort.Ints(keys)
 	totalCount := float64(total.DurationHistogram.Count)
+	fmt.Fprintf(out, "Sockets used: %d (for perfect keepalive, would be %d)\n", total.SocketCount, r.Options().NumThreads)
```
asadali, line 146: +1

asadali (commented):
- the build v6 go1.10/go1.8 changes look good.
- minor change on "Go to Top" functionality.

ui/templates/browse.html
ldemailly: it’s a bit subtle but there are really more than 1 top depending on report only vs full server so I left the label ambiguous rather than add more variables and mode

fhttp/http.go
```diff
@@ -683,14 +703,22 @@ func (c *FastClient) readResponse(conn *net.TCPConn) {
...
 			if err != nil {
+				if reusedSocket && c.size == 0 {
+					// Ok for reused socket to be dead once (close by server)
+					log.Infof("Closing dead socket %v (err %v at first read)", *conn, err)
+					c.errorCount++
+					err = conn.Close() // close the previous one
+					if err != nil {
+						log.Warnf("Error closing dead socket %v: %v", *conn, err)
+					}
+					c.code = -2 // special "retry once" code
```
douglas-reid: worth making the special code a const ?

douglas-reid (commented):
LGTM

douglas-reid (approved):

PR #177, opened 2018-03-04
fnet cleanup part 1

ldemailly:
- report mode is now tested in Webtest
- use common code, now in fnet/ for listeners. add unit tests
- unified signature of various methods
- servers start are now all async
- fixes for localhost and port 0
- setup PPROF safely/explicitly and test in Webtest (that it's present in "server" and absent in "report" mode)

~based/stacked on top of #173 - might merge there but maybe it's easier to review separately~
now master based

### DIFF
--- a/fhttp/http.go
+++ b/fhttp/http.go
@@ -1211,10 +1211,10 @@ func HTTPServer(name string, port string) (*http.ServeMux, *net.TCPAddr) {
 // (when secure is true) server on it and returns the listening port and
 // mux to which one can attach handlers to.
 // TODO: make signature consistent ?
-func DynamicHTTPServer(closing bool) (int, *http.ServeMux) {
+func DynamicHTTPServer(closing bool) (*http.ServeMux, *net.TCPAddr) {
 	if !closing {
 		mux, addr := HTTPServer("dynamic", ":0")
-		return addr.Port, mux
+		return mux, addr
 	}
 	// Note: we actually use the fact it's not supported as an error server for tests - need to change that
 	log.Errf("Secure setup not yet supported. Will just close incoming connections for now")
@@ -1226,7 +1226,7 @@ func DynamicHTTPServer(closing bool) (int, *http.ServeMux) {
 			log.Fatalf("Unable to serve closing server on %s: %v", addr.String(), err)
 		}
 	}()
-	return addr.Port, nil
+	return nil, addr
 }
 
 /*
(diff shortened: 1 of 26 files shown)


fhttp/http.go
```diff
@@ -395,21 +395,12 @@ func NewFastClient(o *HTTPOptions) Fetcher {
...
-	}
-	if len(addrs) > 1 && log.LogDebug() {
-		log.Debugf("Using only the first of the addresses for %s : %v", bc.host, addrs)
-	}
-	log.Debugf("Will go to %s", addrs[0])
-	bc.dest.IP = addrs[0]
-	bc.dest.Port, err = net.LookupPort("tcp", bc.port)
-	if err != nil {
-		log.Errf("Unable to resolve port '%s' : %v", bc.port, err)
+	addr := fnet.Resolve(bc.hostname, bc.port)
```
rkpagadala, line 398: being pedantic :)
go idiom is to return err and check for it
```
- addr := fnet.Resolve(bc.hostname, bc.port)
- if addr == nil {
+ addr, err := fnet.Resolve(bc.hostname, bc.port)
+ if err != nil {
```

fhttp/http.go
```diff
@@ -1198,35 +1189,42 @@ func closingServer(listener net.Listener) error {
...
+		err := s.Serve(listener)
+		if err != nil {
+			log.Fatalf("Unable to serve %s on %s: %v", name, addr.String(), err)
 		}
+	}()
+	return m, addr
+}
+
+// DynamicHTTPServer listens on an available port, sets up an http or https
+// (when secure is true) server on it and returns the listening port and
```
rkpagadala: please update comment, no secure parameter

fhttp/http_test.go
```diff
@@ -493,11 +493,11 @@ func TestGenerateSize(t *testing.T) {
 // Many of the earlier http tests are through httprunner but new tests should go here
 
 func TestEchoBack(t *testing.T) {
-	p, m := DynamicHTTPServer(false)
+	m, a := DynamicHTTPServer(false)
```
rkpagadala, line 496:
```
-      m, a := DynamicHTTPServer(false)
+	mux, addr := DynamicHTTPServer(false)
```
here and elsewhere

ui/uihandler.go
```diff
@@ -963,28 +976,22 @@ func RedirectToHTTPSHandler(w http.ResponseWriter, r *http.Request) {
...
-// sets the urlHostPort variable and returns hostport unmodified.
+// sets the urlHostPort variable.
 func setHostAndPort(inputPort string, addr *net.TCPAddr) {
 	urlHostPort = inputPort
 	portStr := inputPort
 	if addr != nil {
 		urlHostPort = addr.String()
 		portStr = fmt.Sprintf(":%d", addr.Port)
 	}
+	if !strings.Contains(inputPort, ":") {
```
rkpagadala, line 992: call NormalizePort?

fhttp/http.go
ldemailly: I know; but I think nil is also a perfectly valid marker for failure - with the assumption (which I realize is 'bad' for a library, I have a comment to that effect in fnet about splitting the functions / error handling - specially where I have log.Fatalf...) that error logging already happened

fhttp/http_test.go
ldemailly: I thought go was all about short names ? (those are test files)

fhttp/http.go
ldemailly: good catch, fixing

ui/uihandler.go
ldemailly: yeah the logic is a bit in 2 places but I didn't want to re-call that method; I think I need to cleanup the whole uihandler some more in the future - right now just trying to get manual tests to work with localhost special case

fhttp/http_test.go
rkpagadala: not fixing is fine

ui/uihandler.go
rkpagadala: ok

rkpagadala (approved):

Issue #99, opened 2018-01-24
ui and/or command line to merge multiple json results (using the existing stats.Merge)


ldemailly:
assuming their start time is similar
the histograms can be merged with stats Merge/Transfer (but those methods are on Histogram not HistgramData at the moment) - if we do it at UI level only fortioResultToJsChartData could take multiple input (separately from fortioAddToMultiResult)

cc @fawadkhaliq who mentioned this first

asadali:
listing down a few assumptions here for the cli mode
- sample usage:
```
$ fortio merge file1.json file2.json ...
```
this should create a merged file (what should it be named?) with all the input result files. (file paths are relative and output file will be created in the same directory)

approach:
- parse os.Args for .json files
- iteratively merge each file
- how to merge:
  - Labels (append all? choose the first file's label as the only label)
  - merge StartTime (earliest wins)
  - merge RequestedQPS (average? how to handle differences?)
  - merge RequestedDuration (maximum)
  - ActualQPS (average)
  - ActualDuration (average)
  - NumThreads (sum)
  - Version (first file's version)
  - RetCodes (dependent on identifying the type of result)
  - HistogramData
    - convert to Histogram and call Merge? is the conversion possible?
    - create MergeHistogramData and aggregate values in common buckets

open questions:
1. how to determine if results are mergeable? (eg assuming similar start times)
2. should the type of results be specified in the saved json file? (http/grpc)

ldemailly:
Thanks for looking into this!

> this should create a merged file (what should it be named?) with all the input result files. (file paths are relative and output file will be created in the same directory)


if -a (current autosave) is specified it can create a new ID which then turns into a file name
otherwise the merged output could be stdout (like `-json -` does today)

or you have to specify `-json name`


probably it should warn if values are very different (like merging 1 requested qps with 100)
we could use "merged" as placeholder for string types or even null out some fields in merged json
(would have to check they are still graphable, potentially fixing bugs in graphing conversion)

> ActualDuration (average)

sum would be accurate here, not avg

> NumThreads (sum)

No, avg, or replaced by a list of numbers when they are not ==

> ActualQPS (average)

not really, should be recalculated - ie weighted avg - 1s run at 10kqps + 10s run at 5qps is not an 11s run at 7.5kqps

> RetCodes (dependent on identifying the type of result)

nope, just merged, it's a map

> should the type of results be specified in the saved json file? (http/grpc)

[...]

Issue #124, opened 2018-02-13
Inconsistent behavior noticed in latency across different qps


fawadkhaliq:
**Issue**
Fortio shows significantly increased latency at different qps for the same number of threads.

For example, 

400 qps with 1 Fortio instance and 48 threads,
   99.9% = ~100ms across several runs (consistent behavior)
500qps with 1 Fortio instance and 48 threads
  99.9% = ~10ms across several runs with occasional spike to ~50ms
1000qps with 1 Fortio instance and 48 threads
  99.9% = ~5ms across several runs

This behavior seems a bit inconsistent.

**Test setup**
k8s on aws
1 master, 2 nodes
t2.medium specs: 2 vcpus, 4 gb mem
No Istio involved in this.

Fortio as the client
lwan as the server

fortio.yaml
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: fortio
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: fortio
    spec:
      containers:
      - image: istio/fortio
        imagePullPolicy: IfNotPresent
        name: fortio
        resources:
          requests:
            memory: "128Mi"
            cpu: "0.1"
          limits:
            memory: "512Mi"
            cpu: "2"
```
lwan.yaml
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: lwan
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: lwan
    spec:
      containers:
      - image: jaxgeller/lwan
        imagePullPolicy: IfNotPresent
        name: lwan
        ports:
        - containerPort: 8080
---
apiVersion: v1
kind: Service
metadata:
  name: lwan
  labels:
    app: lwan
spec:
  ports:
  - port: 8080
    name: http
  selector:
    app: lwan    
```

**Sample run output**
400 QPS
---------
```
kubectl exec fortio-3574055260-gwlj3 -- fortio load -qps 400 -t 10s -c 48 -r 0.0001 http://lwan:8080
```
```
Fortio 0.6.9 running at 400 queries per second, 2->2 procs, for 10s: http://lwan:8080
23:48:20 I httprunner.go:72> Starting http test for http://lwan:8080 with 48 threads at 400.0 qps
Starting at 400 qps with 48 thread(s) [gomax 2] for 10s : 83 calls each (total 3984)
23:48:30 I periodic.go:523> T005 ended after 9.960805951s : 83 calls. qps=8.332659064768484
[...]

PR #182, opened 2018-03-07
[wip] merge results from multiple JSON files

asadali:
this is a very rough first cut for #99 
creating this to gather feedback

- refactor results operations to a results package. the logic can be shared between fortio_main and uihandler
- more assumptions outlined in comments on this PR

### DIFF
--- a/fortio_main.go
+++ b/fortio_main.go
@@ -17,7 +17,6 @@ package main
 // Do not add any external dependencies we want to keep fortio minimal.
 
 import (
-	"encoding/json"
 	"flag"
 	"fmt"
 	"net/http"
@@ -33,6 +32,7 @@ import (
 	"istio.io/fortio/stats"
 	"istio.io/fortio/ui"
 	"istio.io/fortio/version"
+	"istio.io/fortio/results"
 )
 
 var httpOpts fhttp.HTTPOptions
@@ -145,6 +145,10 @@ func main() {
 	if err != nil {
 		usage("Unable to extract percentiles from -p: ", err)
 	}
+	filesList, err := results.ParseArgsForResultFiles(os.Args)
+	if err != nil {
+		usage("Unable to parse files from args: ", err)
+	}
 	baseURL := strings.Trim(*baseURLFlag, " \t\n\r/") // remove trailing slash and other whitespace
 	sync := strings.TrimSpace(*syncFlag)
 	if sync != "" {
@@ -176,6 +180,12 @@ func main() {
 		ui.Serve(baseURL, *echoPortFlag, *echoDbgPathFlag, *uiPathFlag, *staticDirFlag, *dataDirFlag)
 	case "grpcping":
 		grpcClient()
+	case "merge":
+		res, err := results.LoadResultFiles(filesList)
+		if err != nil {
+			os.Exit(1) // error logged already
+		}
+		results.MergeRunnerResults(res)
 	default:
 		usage("Error: unknown command ", command)
 	}
@@ -261,7 +271,7 @@ func fortioLoad(justCurl bool, percList []float64) {
 		Labels:      labels,
 		Exactly:     *exactlyFlag,
 	}
-	var res periodic.HasRunnerResult
+	var res results.HasRunnerResult
 	var err error
 	if *grpcFlag {
 		o := fgrpc.GRPCRunnerOptions{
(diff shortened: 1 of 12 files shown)


fortio_main.go
```diff
@@ -145,6 +145,10 @@ func main() {
 	if err != nil {
 		usage("Unable to extract percentiles from -p: ", err)
 	}
+	filesList, err := results.ParseArgsForResultFiles(os.Args)
```
asadali: this will assume usage to be:

```
fortio merge file1.json file2.json
```

caveat, it won't honor the `-json` flag currently without some special casing.
i will add support for `-` and `-a` before marking PR ready

results/results.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// Result returns the common RunnerResults.
+func (r *RunnerResults) Result() *RunnerResults {
+	return r
+}
+
+// SaveJSON saves the result as json to the provided file
+//
+func SaveJSON(res HasRunnerResult, fileName string) (int, error) {
```
asadali: will add test for this and other functions introduced in this file

results/results.go
```diff
@@ -0,0 +1,171 @@
...
+			return nil, err
+		}
+		result := RunnerResults{}
+		json.Unmarshal(raw, &result)
+		res = append(res, &result)
+	}
+	return res, nil
+}
+
+func MergeRunnerResults(results []*RunnerResults) {
```
asadali: this is the part I am having trouble with :D will appreciate feedback hints

results/results.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		// merge RequestedQPS
+		// merge RequestedDuration
+		// ActualQPS
+		// ActualDuration
+		// merge NumThreads
+		mergedResults.NumThreads += result.NumThreads
+		// merge Version
+		// merge HistogramData
+		mergedResults.DurationHistogram = stats.MergeHistograms(mergedResults.DurationHistogram.Histogram(), result.DurationHistogram.Histogram()).Export()
```
asadali: this is a lot of redundant conversions between Histogram and HistogramData. However, merging HistogramData seems more complicated due to the possible differences in the percentiles or buckets?

stats/stats.go
```diff
@@ -464,9 +477,9 @@ func (h *Histogram) copyHDataFrom(src *Histogram) {
 	}
 }
 
-// Merge two different histogram with different scale parameters
+// MergeHistograms merges two different histogram with different scale parameters
 // Lowest offset and highest divider value will be selected on new Histogram as scale parameters
-func Merge(h1 *Histogram, h2 *Histogram) *Histogram {
+func MergeHistograms(h1 *Histogram, h2 *Histogram) *Histogram {
```
asadali, line 484: renaming this to be more specific

asadali (commented):
commented on the parts I need another pair of 👀 on

stats/stats.go
```diff
@@ -308,6 +308,19 @@ func (e *HistogramData) CalcPercentile(percentile float64) float64 {
...
 }
 
+func (hd *HistogramData) Histogram() *Histogram {
+	res := NewHistogram(0, 10)
+	res.Counter.Min = hd.Min
+	res.Counter.Max = hd.Max
+
+	for _, bucket := range hd.Data {
+		midpoint := (bucket.Start + bucket.End)/2
+		res.RecordN(midpoint, int(bucket.Count))
```
asadali: this is a very lazy attempt at reverse engineering the histogram data. ideas for a better way?

ldemailly:
I don't have time to review fully yet right now but:

why moving the runner results away from the runner - this should stay in periodic/ package - imo
(not to mention not mixing code movement diffs with feature diffs)

ldemailly:
ps: thanks a lot for working on this
(though maybe there are higher pri items...)

asadali:
this is turning out to be high priority for my team.
requirement: run fortio in distributed client mode and combine results (ideally on the UI, visuals communicate better)

currently, i can run fortio using multiple pods, however, inspecting the results individually isn't fun.
do you have another more effective workflow?

also, what's the best way to parse the issues? currently, there is only one high-pri item, the linter flakiness. which issues do you want to prioritize. send them my way :) i am learning quite a few things through this.

p.s: apologies about the last force push (linter changes). i didn't see your comment until after pushing and assumed that you may not have started reviewing.

asadali:
just realized that my comments on the PR are now auto-hidden cuz of the linter fixes. will appreciate if you can take a look at the collapsed comments too while reviewing

also, if you are busy, then this can wait. i have a fair idea for a localized hack to satisfy my use-case, i can unblock myself with that. (basically a very simple merge assuming that percentiles/buckets duration etc are all the same)

ldemailly:
re you use case (basically #134) yes I think it's useful

but in my experience so far, a single client can easily produce >> 100k qps which is way more than most ingress+backend can handle, so I'm not sure you really need more than 1 fortio client ?
(is it related to #124 which I need to look at soon)

will check the comments/questions too probably next week

results/results.go
ldemailly: maybe an initial version could only merge histograms with same settings; on the other hand the fun part is to merge the ones that are different. the function on histogram isn't perfect (and not used really anywhere but in tests)

stats/stats.go
ldemailly: we could export also missing pieces (offset and divider) in json to make it easy reconstruct the original histogram

ldemailly:
I added some comments inline, need to look more as well

asadali:
thanks for the review! sorry about the delay here, I will get back to this today

ldemailly:
I haven't had time to look in details sorry

asadali:
closing this for now, will reopen once i make some progress

ldemailly:
I was just going to have a look ;-)

we can schedule some time to brainstorm if you want - lmk

asadali:
yes! i can definitely use some help :)
i didn't want it to linger on as a WIP when i wasn't actually working on it. i will reopen it when I get back to this. hopefully sometime mid next week

PR #190, opened 2018-03-07
add MaxPayloadSizeKb parameter to set max payload size for echo call. eg. echo/?size=

kidiyoor:
wrt: https://github.com/istio/fortio/issues/183


README.md
```diff
@@ -113,6 +113,8 @@ and flags are:
       (default Info)
   -logprefix string
     	Prefix to log lines before logged messages (default "> ")
+  - maxpayloadsizekb
+       MaxPayloadSizeKb is the maximum size of payload to be generated by the EchoHandler size= argument (In kilobytes defaults to 256)
```
ldemailly: manually wrap to 80 cols (I need to find a tool to do that) - when doing it pay attention to tab vs space, copy the prefix from another existing wrapped line

fhttp/http.go
```diff
@@ -972,9 +970,9 @@ func generateSize(sizeInput string) (size int) {
 		}
 		size = s
 		log.Debugf("Parsed size %s -> %d", sizeInput, size)
-		if size > MaxPayloadSize {
-			log.Warnf("Requested size %d greater than max size %d, using max instead", size, MaxPayloadSize)
-			size = MaxPayloadSize
+		if size > MaxPayloadSizeKb*1024 {
```
ldemailly: I don't like the *1024 6 times :-)

fhttp/http.go
```diff
@@ -1151,7 +1149,7 @@ func EchoHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-var payload = make([]byte, MaxPayloadSize)
+var payload = make([]byte, MaxPayloadSizeKb*1024)
```
ldemailly: this is not going to work...
it's initialized during init() that probably need to be changed

you need a test that the change takes effect and that you don't get issues or non random data after the original size - you probably need a ChangePayloadSize() that translates from kb to bytes and sets the payload

fhttp/http_test.go
```diff
@@ -480,7 +480,7 @@ func TestGenerateSize(t *testing.T) {
 		{"512:45,512:55", 512},
 		{"0", 0}, // and not -1
 		{"262144", 262144},
-		{"262145", MaxPayloadSize}, // MaxSize test
+		{"262145", MaxPayloadSizeKb}, // MaxSize test
```
ldemailly: that should fail... wrong type

ldemailly (commented):
Thanks for the PR - but see below

ldemailly:
glad the tests work as I expect:
```
--- FAIL: TestGenerateSize (0.00s)
	http_test.go:488: Got 262144, expected 256 for generateSize("262145")
```

protip : as per readme, run "make test webtest lint" before submitting a PR

README.md
kidiyoor: ok. I'll try to copy from exiting ones

fhttp/http.go
kidiyoor: yep. writing a function to convert kb to bytes

fhttp/http_test.go
kidiyoor: ah. I need to change use bytes here not kb.

kidiyoor:
```var payload = make([]byte, MaxPayloadSizeKb*1024)``` issue is still not addressed.

**Comment by ldemailly** [comment]
you need function to change the buffer size

you can make the flag in fortio_main a regular variable and call fhttp.ChangeMaxPayloadSize() inside main after flag parsing

**Review by ldemailly — DISMISSED** [review]
need a function for changing the value to enlarge payload if needed (and initialize the new one)
ps: no need to do anything is the size is shrunk

and ideally a test for that (maybe in webtest.sh is easier)

**Comment by kidiyoor** [comment]
latest commit should fix that issue. However test not added yet.

File: fhttp/http.go [hunk 5]
```diff
@@ -1163,6 +1163,16 @@ func init() {
 	rand.Read(payload) // nolint:gas
 }
 
+// ChangeMaxPayloadSize is used to change max payload size and fill it with pseudorandom content
+func ChangeMaxPayloadSize(newMaxPayloadSizeKb int) {
```
ldemailly: change init() to call this as well or ideally to do it only if not already done by flags

also wtb test

Webtest.sh
```diff
@@ -41,6 +42,14 @@ if [ "$SIZE" -lt 50000 ]; then
   echo "Too small fetch for js: $SIZE"
   exit 1
 fi
+# Check if max payload set to value passed in cmd line parameter -maxpayloadsizekb
+SIZE=$($CURL "${BASE_URL}/echo?size=1048576" |wc -c)
+# It should return 8194, not sure why the difference
```
ldemailly: the size includes the headers, you'd have to subtract the headers

Webtest.sh
```diff
@@ -41,6 +42,14 @@ if [ "$SIZE" -lt 50000 ]; then
...
+if ! [ "$SIZE" -eq 8310 ]; then
```
ldemailly: make it [ "$SIZE" -gt 8200 ] . so it's not brittle to 1 byte change in headers - or get the body only using -stdclient and then it should be exactly 8k

fhttp/http.go
```diff
@@ -972,9 +977,9 @@ func generateSize(sizeInput string) (size int) {
 		}
 		size = s
 		log.Debugf("Parsed size %s -> %d", sizeInput, size)
-		if size > MaxPayloadSize {
-			log.Warnf("Requested size %d greater than max size %d, using max instead", size, MaxPayloadSize)
-			size = MaxPayloadSize
+		if size > changeKbToBytes(MaxPayloadSizeKb) {
```
ldemailly: please use an intermediate variable, don't call the same function 6 times

fhttp/http.go
```diff
@@ -1151,13 +1156,23 @@ func EchoHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-var payload = make([]byte, MaxPayloadSize)
-
 func init() {
+	if len(payload) == 0 {
```
ldemailly: that will always be true as init runs before flag parsing; it wasn't what I meant (even if I can see how it can be interpreted that way)
it was a) call the function from init - or even better and optional - not ChangeMaxPayloadSize twice

fhttp/http.go
```diff
@@ -674,6 +678,11 @@ func EscapeBytes(buf []byte) string {
 	return e[1 : len(e)-1]
 }
 
+// Change kb to bytes
+func changeKbToBytes(kb int) int {
```
ldemailly: the name is a bit long for go style... `kbToBytes()`?

fhttp/http.go
```diff
@@ -1151,13 +1156,23 @@ func EchoHandler(w http.ResponseWriter, r *http.Request) {
...
-
 func init() {
+	if len(payload) == 0 {
+		ChangeMaxPayloadSize(MaxPayloadSizeKb)
+	}
+}
+
+// ChangeMaxPayloadSize is used to change max payload size and fill it with pseudorandom content
+func ChangeMaxPayloadSize(newMaxPayloadSizeKb int) {
+	MaxPayloadSizeKb = newMaxPayloadSizeKb
```
ldemailly: we can either have 2 variables or just the bytes one or use len(payload) for the bytes value

fhttp/http.go
```diff
@@ -1151,13 +1156,23 @@ func EchoHandler(w http.ResponseWriter, r *http.Request) {
...
-var payload = make([]byte, MaxPayloadSize)
-
 func init() {
+	if len(payload) == 0 {
+		ChangeMaxPayloadSize(MaxPayloadSizeKb)
+	}
+}
+
+// ChangeMaxPayloadSize is used to change max payload size and fill it with pseudorandom content
+func ChangeMaxPayloadSize(newMaxPayloadSizeKb int) {
```
ldemailly: add a unit test for this

ldemailly (commented):
thx!

kidiyoor:
I'm sure something is still missing :?)

Webtest.sh
```diff
@@ -45,7 +45,7 @@ fi
 # Check if max payload set to value passed in cmd line parameter -maxpayloadsizekb
 SIZE=$($CURL "${BASE_URL}/echo?size=1048576" |wc -c)
 # It should return 8194, not sure why the difference
-if ! [ "$SIZE" -eq 8310 ]; then
+if [ "$SIZE" -lt 8194 ] || [ "$SIZE" -gt 8400 ]; then
```
ldemailly: 8*1024
8192

so make it 8191 for -lt ?

we should also check it doesn't compress but... some other time

Webtest.sh
ldemailly: update comment
```
# Payload is 8192 but between content chunking and headers fast client can return up to 8300 or so
```

Webtest.sh
```diff
@@ -41,6 +42,14 @@ if [ "$SIZE" -lt 50000 ]; then
   echo "Too small fetch for js: $SIZE"
   exit 1
 fi
+# Check if max payload set to value passed in cmd line parameter -maxpayloadsizekb
+SIZE=$($CURL "${BASE_URL}/echo?size=1048576" |wc -c)
+# It should return 8194, not sure why the difference
+if [ "$SIZE" -lt 8194 ] || [ "$SIZE" -gt 8400 ]; then
```
ldemailly: 8191

fhttp/http.go
```diff
@@ -53,8 +53,10 @@ type Fetcher interface {
 }
 
 var (
-	// BufferSizeKb size of the buffer (max data) for optimized client in kilobytes defaults to 32k.
+	// BufferSizeKb size of the buffer (max data) for optimized client in kilobytes defaults to 128k.
 	BufferSizeKb = 128
+	// MaxPayloadSize is the maximum size of payload to be generated by echo handler defaults to 256k
```
ldemailly: missing " in bytes" and "size= argument" from the original comment
In theory as there is a ChangeXxxx function to change it, the variable should be private
eg `maxPayloadSize` but we do need the default value for the flag... well or do we the flag default can be whichever

fortio_main.go
```diff
@@ -114,12 +114,15 @@ var (
...
 	baseURLFlag = flag.String("base-url", "",
 		"base URL used as prefix for data/index.tsv generation. (when empty, the url from the first request is used)")
+	newMaxPayloadSizeKb int
 )
 
 func main() {
 	flag.Var(&headersFlags, "H", "Additional Header(s)")
 	flag.IntVar(&fhttp.BufferSizeKb, "httpbufferkb", fhttp.BufferSizeKb,
 		"Size of the buffer (max data size) for the optimized http client in kbytes")
+	flag.IntVar(&newMaxPayloadSizeKb, "maxpayloadsizekb", fhttp.MaxPayloadSize/1024,
```
ldemailly: just put it with the other flags, as a flag.Int() doesn't need to use a variable

fortio_main.go
```diff
@@ -114,12 +114,15 @@ var (
 	syncFlag    = flag.String("sync", "", "index.tsv or s3/gcs bucket xml URL to fetch at startup for server modes.")
 	baseURLFlag = flag.String("base-url", "",
 		"base URL used as prefix for data/index.tsv generation. (when empty, the url from the first request is used)")
+	newMaxPayloadSizeKb int
```
ldemailly: delete/replace by flag directly (see goMaxProcsFlag as an example)

fhttp/http_test.go
```diff
@@ -705,3 +705,25 @@ func BenchmarkFoldFind(b *testing.B) {
...
+
+func TestChangeMaxPayloadSize(t *testing.T) {
+	var tests = []struct {
+		input    int
+		expected int
+	}{
+		// negative test cases
+		{-1, 0},
+		// lesser than current default
+		{0, 0},
```
ldemailly, line 812: (semi optional) add a test (spinning an echo server like other tests in this file) making such request, does it work with 0 and with 1 ?

[we may get race test failure on changing the size btw... might have to use the non race file to test that]

ldemailly (commented):
I am investing here, so I hope you're ok with all the changes and comment, let me know when you're running out of time :-)

ldemailly:
I split code into more manageable chunks/files - sorry you need to rebase/have a conflict

ldemailly:
ping :)

kidiyoor:
Pong. I had given a little pause on this.  
Expect few more commits soon enough.

fhttp/http_test.go
kidiyoor: it works for 0 and 1. Tested it manually.  (semi optional) => Next PR

fortio_main.go
kidiyoor: fixed.

fortio_main.go
kidiyoor: fixed.

Webtest.sh
kidiyoor: fixed.

Webtest.sh
kidiyoor: done.

fhttp/http_server.go
```diff
@@ -97,17 +102,25 @@ func EchoHandler(w http.ResponseWriter, r *http.Request) {
...
+		MaxPayloadSize = newMaxPayloadSize
+	} else {
+		MaxPayloadSize = 0
+	}
+	payload = make([]byte, MaxPayloadSize)
 	// One shared and 'constant' (over time) but pseudo random content for payload
 	// (to defeat compression). We don't need crypto strength here, just low cpu
 	// and speed:
-	rand.Read(payload) // nolint:gas
+	_, err := rand.Read(payload) // nolint:gas
```
ldemailly: (nit/optional) you can probably remove the nolint comment now that you check for err (also I gave up on gas anyway)

README.md
```diff
@@ -112,6 +112,8 @@ and flags are:
       (default Info)
   -logprefix string
     	Prefix to log lines before logged messages (default "> ")
+  -maxpayloadsizekb int
+        MaxPayloadSizeKb is the maximum size of payload to be generated by the EchoHandler size= argument (In kilobytes defaults to 256)
```
ldemailly: please manually wrap to 80 cols (or I can do it later too it's ok)

ldemailly (approved):
thx! just 2 minor stuff

fhttp/http.go
kidiyoor: Using one default in flag and another in http_server file doesn't sound good ? I have kept  MaxPayloadSize public for now

README.md
kidiyoor: done.

fhttp/http_server.go
kidiyoor: I'm not sure how gas works. I copied it from existing

kidiyoor: I think I used to get 'unsafe use of rand' error if I didn't use // nolint:gas this execption

kidiyoor: let me try again removing the - // nolint:gas

PR #195, opened 2018-03-10
Code move: split http into client,server,utils

ldemailly:
For better readability and more manageable chunk

While preserving history (will not squash merge)


ldemailly:
to review look at https://github.com/istio/fortio/pull/195/commits/52e9cc5ff31e1daba12d4e0a22a866196582c8c1 which shows only red in each 3 files
(modulo the move of payload in 1 section of the server code)

ldemailly:
given this is purely mechanical I may merge without review but a stamp is welcome

PR #197, opened 2018-03-11
Grpc cleanup and improve grpc test coverage

### LINKED ISSUE #196: cleanup grpc and increase test coverage
will allow to use it in pilot e2e

ldemailly:
so the grpc ping and health clients can be used as a library / from
istio tests (like pilot’s)

return errors instead of log fatal

Fixes #196

Dropped gas as linter


rkpagadala (approved):
/lgtm

ldemailly:
funny how adding  +79 lines covered and a new file 86% covered turns into "78.4% of diff hit"

PR #198, opened 2018-03-11
Raise http_client coverage

ldemailly:
test http 1.0 mode and default http/https ports
plus more tests
plus parallelized the build for faster builds/results


rkpagadala (approved):
/lgtm

PR #200, opened 2018-03-11
curl -L mode; move generic code out of uihandler; more tests; etc...

ldemailly:
- moved generic code from ui to fhttp/
- added `-L` and `FollowRedirects` options (meant for `fortio curl -L url > file`). -L is shortcut for both follow redirect and using the stdclient (decoding the data part, vs raw output from fast client). So `fortio curl -L ...` can be used for downloads.
- `FetchURL` and `Fetch` helper functions wrapping 1 time use client to help write 1 liner readiness
- more tests
- added go version in `fortio version` output (`version.Long()`)
- changed `-httpreqtimeout` flag to just `-timeout`
- expose `fhttp.RedirectToHTTPS`, `NewHTMLEscapeWriter`, `SetupPPROF`, `FetcherHandler` proxy, `LogRequest`, `OnBehalfOf` to the public fhttp library api.
- solved README.md wrapping of fortio command help text using new `release/updateFlags.sh`


ldemailly:
@mattdelco the `fortio curl -L ...` is for you (and the fhttp.FetchURL() for some of our liveness/readiness tests)

mattdelco:
Thanks, it looks alright to me (though I don't consider myself very knowledgeable about fortio or golang).  I have been a bit apprehensive about relying on stdout (vs an explicit file output parameter), but so far things seem okay (almost everything else is using stderr, either directly or indirectly via the logging library).

ldemailly:
any unix-y program worth its salt will not mix stdout and stderr :-) if it produces something to be saved it will be on stdout while logging goes to stderr

kyessenov (approved):

PR #203, opened 2018-03-14
remove logic to escape % in given urls

### LINKED ISSUE #181: % escaping makes loading URLs through fetch not work
http://X/fortio/fetch/Y/fortio1/fortio/?labels=Fortio&url=localhost%3A8080/echo&qps=100&t=3s&n=&c=8&p=50%2C+75%2C+90%2C+99%2C+99.9&r=0.0001&H=User-Agent%3A+istio%2Ffortio-0.8.0-pre&H=&H=&runner=http&save=on&load=Start

->

Fortio localhost%3A8080/echo

Aborting because unable to create client 0 for http://localhost%253A8080/echo

going to debug shows the issue:

GET /debug?labels=Fortio&url=localhost%253A8080/echo&qps=100&t=3s&n=&c=8&p=50%252C+75%252C+90%252C+99%252C+99.9&r=0.0001&H=User-Agent%253A+istio%252Ffortio-0.8.0-pre&H=&H=&runner=http&save=on&load=Start HTTP/1.1

%253A instead of %3A

asadali:
fixes #181 

@ldemailly this is a minor fix that I had on my local version as the escape logic failed a few of my perf test URLs. sending this in based on discussion on 181.

did i miss any documentation updates?

### DIFF
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -63,15 +63,12 @@ func NewHTTPOptions(url string) *HTTPOptions {
 }
 
 // Init initializes the headers in an HTTPOptions (User-Agent).
-// It replaces plain % to %25 in the url. If you already have properly
-// escaped URLs use o.URL = to set it.
 func (h *HTTPOptions) Init(url string) *HTTPOptions {
 	if h.initDone {
 		return h
 	}
 	h.initDone = true
-	// unescape then rescape % to %25 (so if it was already %25 it stays)
-	h.URL = strings.Replace(strings.Replace(url, "%25", "%", -1), "%", "%25", -1)
+	h.URL = url
 	h.NumConnections = 1
 	if h.HTTPReqTimeOut <= 0 {
 		h.HTTPReqTimeOut = HTTPReqTimeOutDefaultValue
--- a/fhttp/http_test.go
+++ b/fhttp/http_test.go
@@ -90,7 +90,8 @@ func TestNewHTTPRequest(t *testing.T) {
 func TestMultiInitAndEscape(t *testing.T) {
 	// one escaped already 2 not
 	o := NewHTTPOptions("localhost:8080/?delay=10ms:10%,0.5s:15%25,0.25s:5%")
-	expected := "http://localhost:8080/?delay=10ms:10%25,0.5s:15%25,0.25s:5%25"
+	// shouldn't perform any escapes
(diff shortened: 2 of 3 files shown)


ldemailly:
you need to change the default value on the UI I think - a test should fail (or maybe I don't have a test that simulates clicking the start button)

also probably add a test more explictly that a url with a %3F doesn't get mangled

asadali:
gotcha. was working on the tests. thanks for the clarifications

ldemailly:
basically, try

fortio server &
visit http://[::1]:8080/ and click start, it will probably fail to do the % with this PR

I'm still wondering if I could preserve the feature somehow but I guess it's rather low pri

ldemailly:
need UI test somehow as it's not caught (long term/future)

asadali:
surprisingly, i don't see an error for the default UI run with the % in the query. even with the changes in this PR

ui-remove-escape

ldemailly:
that's the point - you get "no error" while the sample url for instance asks for 0.5% of 503 and 1.5% of 429

ditto for the delay not showing up

ldemailly:
there should be some error in the server log too about not able to parse delay and status

ldemailly:
let's try to finish this for 0.8.0 which I want to do soon

asadali:
thanks for the insights. I hadn't fully understood the significance of escaping the url.

now, based on how http_utils -> generateStatus and generateDelay are programmed, they should work equally well without a % suffix. I am guessing the % was for aesthetic/intuitive reasons. (i am not making a change to removePercent)

also, removed % from the default URL on the UI and the parameters are being parsed correctly for configuring echo server's behavior

I didn't add an extra Test but modified MultiIinitAndEscape to test for the %3A case

asadali:
will you prefer that i add a test for the url mentioned in the issue.

something like 
```
# Check we can connect, and run a http QPS test against ourselves through fetch with an escaped URL
$CURL "${BASE_FORTIO}fetch/localhost:8080$FORTIO_UI_PREFIX?url=localhost%3A8080/echo&qps=100&t=3s&n=&c=8&p=50%2C+75%2C+90%2C+99%2C+99.9&json=on&load=Start" | grep ActualQPS
```

or should i edit the basic QPS test to match the URL being displayed in the UI

ldemailly (approved):
it's ok I think just harder to guess what :x does

Issue #160, opened 2018-02-26
add tee/fanout mode


ldemailly:
like fetch but to multiple preconfigured destinations

and high perf (client connection pool etc)

ldemailly:
status returned would be and of all leaf (if any leaf returns an error an error is returned)

ldemailly:
possible spec

```
[proto//][addr:]port[/path] destination
# For instance
4567 foo:7865 # tcp proxy from 4567 to foo.svc:7865
# http proxy/ingress/gateway on port 8085, sending /foo to foo:8080 and /bar to bar:9090;
# prefix dropped:
http://:8085/foo http://foo:8080
http://:8085/bar http://bar:9090
# later
https://cert@blah:443/ http://foo:8080 # sni snif blah; use cert; forward to foo svc
```

also configurable: connection pool; etc

ldemailly:
above maybe as configmap / file:   
`fortio server -proxy-spec filename` 
or command line directly 
`fortio proxy "4567 foo:7865" "http://:8085/foo http://foo:8080" ...`
or maybe `-proxy` to the `fortio server` command

ldemailly:
oh and for fan out:
```
http://:8085/both http://foo:8080,http://bar:9090,...
```
would send to both - tbd on results; probably just concatenate both data with maybe offset or in 2 chunks
 
possible cute idea:
```
http://:8085/or http://foo:8080|http://bar:9090|...
```
would make both requests but reply as soon as the first one returns

ldemailly:
cc @mandarjog @Tahler 

this is what I had in mind for the backend/server part

we can split the kubernetes specific from the generic part of @Tahler's project - wdyt ?

PR #206, opened 2018-03-16
Adding fnet.Proxy* for tcp proxying and -P multi flag for cmd line

### LINKED ISSUE #205: eliminate (most) log.Fatalf from library code
log.Critf instead and return nil or similar (original error) and let the called crash instead of the library

ldemailly:
Start of #160. Also Fixes #205 (less log.Fatalf)

This is to do ingress->svc1->svc2  without closing connections unlike the /fetch/ endpoint does

### DIFF
--- a/ui/uihandler.go
+++ b/ui/uihandler.go
@@ -790,13 +790,16 @@ func downloadOne(w http.ResponseWriter, client *fhttp.Client, name string, u str
 
 // Serve starts the fhttp.Serve() plus the UI server on the given port
 // and paths (empty disables the feature). uiPath should end with /
-// (be a 'directory' path)
-func Serve(baseurl, port, debugpath, uipath, staticRsrcDir string, datadir string, percentileList []float64) {
+// (be a 'directory' path). Returns true if server is started successfully.
+func Serve(baseurl, port, debugpath, uipath, staticRsrcDir string, datadir string, percentileList []float64) bool {
 	baseURL = baseurl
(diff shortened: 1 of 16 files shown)


ldemailly:
made istio/fortio:0.8.0-pre2 from this PR

I will run tests with it and if it looks fine will make 0.8.0 once approved
(will also add some more test to raise the coverage lost with combo of more if() instead of log.Fatalf and more code/features)

ui/uihandler.go
```diff
@@ -790,13 +790,16 @@ func downloadOne(w http.ResponseWriter, client *fhttp.Client, name string, u str
 
 // Serve starts the fhttp.Serve() plus the UI server on the given port
 // and paths (empty disables the feature). uiPath should end with /
-// (be a 'directory' path)
-func Serve(baseurl, port, debugpath, uipath, staticRsrcDir string, datadir string, percentileList []float64) {
+// (be a 'directory' path). Returns true if server is started successfully.
+func Serve(baseurl, port, debugpath, uipath, staticRsrcDir string, datadir string, percentileList []float64) bool {
```
rkpagadala, line 794: why not return error

rkpagadala (approved):
One meta comment. Not returning err and checking for addr != nil or other variables does not read easy for me.

ui/uihandler.go
ldemailly: yeah you're right - will propagate the error in future PR

PR #210, opened 2018-03-20
Bug fix with std client init, create a smaller executable with http client/curl only: fcurl

ldemailly:
Moved common code between the 2 (http client flags) into bincommon/

TODO: make a binary release (extend the tgz one) in addition to the
fortio.fcurl docker image

compressed image sizes:
https://hub.docker.com/r/istio/fortio.fcurl/tags/ 2mb
https://hub.docker.com/r/istio/fortio/tags/ 3mb

uncompressed:
```
REPOSITORY             TAG                 IMAGE ID            CREATED             SIZE
istio/fortio.fcurl     0.8.1-pre1          8d76b0e6bb26        41 minutes ago      6.19MB
istio/fortio.echosrv   0.8.1-pre1          9587d7af4c79        42 minutes ago      5.88MB
istio/fortio           0.8.1-pre1          9707ad2fdc2f        42 minutes ago      8.6MB
```

not tremendously smaller but maybe convenient for health checks etc/alternative to curl (3mb compressed) for scratch images or for go programs

also has less cognitive pressures on the flags:
```
$ fcurl
Φορτίο fortio-curl 0.8.1-pre usage:
	fcurl [flags] url
flags are:
  -H value
    	Additional Header(s)
  -L	Follow redirects (implies -std-client) - do not use for load test
  -compression
    	Enable http compression
  -halfclose
    	When not keepalive, whether to half close the connection (only for fast http)
  -http1.0
    	Use http1.0 (instead of http 1.1)
  -httpbufferkb int
    	Size of the buffer (max data size) for the optimized http client in kbytes (default 128)
  -httpccch
    	Check for Connection: Close Header
  -k	Do not verify certs in https connections
  -keepalive
    	Keep connection alive (only for fast http 1.1) (default true)
  -logcaller
    	Logs filename and line number of callers to log (default true)
  -loglevel value
    	loglevel, one of [Debug Verbose Info Warning Error Critical Fatal] (default Info)
  -logprefix string
    	Prefix to log lines before logged messages (default "> ")
  -quiet
    	Quiet mode: sets the loglevel to Error and reduces the output.
  -stdclient
    	Use the slower net/http standard client (works for TLS)
  -timeout duration
    	Connection and read timeout value (for http) (default 15s)
Error: need a url as parameter
```


ldemailly:
example:
```
$ docker run istio/fortio.fcurl:0.8.1-pre1 -L https://dl.google.com/go/go1.10.src.tar.gz | shasum -a 256
f3de49289405fda5fd1483a8fe6bd2fa5469e005fd567df64485c4fa000c7f24  -
```
matches:
screen shot 2018-03-19 at 11 05 03 pm


cc @mattdelco

ldemailly:
I found the bug because
https://github.com/istio/fortio/blob/master/fhttp/http_client.go#L200
errorred out (doing fcurl -loglevel debug -L localhost:8080 )
```
15:23:18 E http_client.go:200> Unable to dump request http: nil Request.Header
```

Issue #162, opened 2018-02-27
smart default ports for grpc mode with http/https scheme


asadali:
pointed out by @ldemailly 
> for the grpc mode
>  if no :port is specified but http:// prefix is there, use 80, if https:// is there, use 443

danehans:
@ldemailly I thought [this](https://github.com/istio/fortio/blob/master/fgrpc/grpcrunner.go#L47) code addresses this issue? If grpc-secure mode is specified or https prefixes the grpcping destination, then tls is used.

ldemailly:
the code doesn't set the port

danehans:
OK, let me know if you approve of this approach:

1. Create a func called `AppendPort` in the `fnet` pkg that parses a string for http/https in the prefix and appends :443/:80 accordingly. The func returns the original string if it's not prefixed with http/https or is unable to parse the string.
2. Update the fgrpc.Dial func to call `serverAddr := fnet.AppendPort(serverAddr string)`.

ldemailly:
the trick there is differentiate between `:` for port (http://grpchost:123) vs `:` from ip addresses (http://[::1]/) - for http fast client it's done that way indirectly: (through url parsing)
https://github.com/istio/fortio/blob/master/fhttp/http_client.go#L406-L407

I think the approach would be to parse as a url to not duplicate ipv6 escaping ?

PR #215, opened 2018-03-23
http/https prefix in grpc destination implies port 80/443 default port

danehans:
Implemented for Fortio http and grpc clients.

Fixes Issue #162


Webtest.sh
```diff
@@ -59,7 +59,7 @@ docker exec $DOCKERNAME /usr/local/bin/fortio load -H Foo:Bar -H Blah:Blah -qps
 # Do a grpcping
 docker exec $DOCKERNAME /usr/local/bin/fortio grpcping localhost
 # Do a grpcping to a scheme-prefixed destination
-docker exec $DOCKERNAME /usr/local/bin/fortio grpcping https://fortio.istio.io:443
+docker exec $DOCKERNAME /usr/local/bin/fortio grpcping https://fortio.istio.io
```
ldemailly: 🎉

bincommon/commonflags.go
```diff
@@ -110,7 +111,7 @@ func FetchURL(o *fhttp.HTTPOptions) {
 // fortio_main and fcurl.
 func SharedHTTPOptions() *fhttp.HTTPOptions {
 	url := strings.TrimLeft(flag.Arg(0), " \t\r\n")
-	httpOpts.URL = url
+	httpOpts.URL = fnet.AppendPort(url)
```
ldemailly: http/https already works so let's not change that part (as mentioned in https://github.com/istio/fortio/issues/162#issuecomment-375437361 too)

fnet/network_test.go
```diff
@@ -60,6 +60,105 @@ func TestNormalizePort(t *testing.T) {
...
+		input  string
+		output string
+	}{
+		{
+			"valid IPv4 address",
+			"1.2.3.4",
+			"1.2.3.4",
+		},
+		{
+			"valid IPv6 address",
```
ldemailly: add the already bracketized ones eg [::1] should stay [::1]

fnet/network.go
```diff
@@ -34,6 +35,91 @@ func NormalizePort(port string) string {
...
+	case u.Scheme == "http":
+		port = "80"
+	case u.Scheme == "https":
+		port = "443"
+	default:
+		// Unsupported scheme for valid url s
+		log.Errf("Unsupported URL scheme: %s", u.Scheme)
+		return s
+	}
+	ip := net.ParseIP(u.Host)
```
ldemailly:
```
$ git grep -n net.ParseIP
fgrpc/grpcrunner.go:188:        if ip := net.ParseIP(dest); ip != nil {
fnet/network.go:77:     isAddr := net.ParseIP(host)
```
let's clean up the duplicate if possible
(and if the  BracketizeIPv6Address was needed we should parse only once)

fnet/network.go
```diff
@@ -34,6 +35,91 @@ func NormalizePort(port string) string {
 	return ":" + port
 }
 
+// BracketizeIPv6Address returns s in brackets if s is an IPv6 address.
+func BracketizeIPv6Address(s string) string {
```
ldemailly: I don't think this function is needed, a url with ipv6 without [] is invalid (can't tell if last part is a port or part of the ip)

fnet/network.go
```diff
@@ -34,6 +35,91 @@ func NormalizePort(port string) string {
...
+	}
+}
+
+// AppendPort parses s as a url and returns s with the host-portion of s as s:port.
+// s is returned unmodified if s cannot be parsed or the host-portion of s is an
+// invalid domain name or invalid IP address.
+func AppendPort(s string) string {
+	u, err := url.Parse(s)
+	if err != nil {
+		log.Errf("Unable to Parse URL %s: %v", err)
```
ldemailly: this is going to generate errors for ipv6 in grpc 
https://play.golang.org/p/RKn_AaaGB39

fnet/network.go
```diff
@@ -34,6 +35,91 @@ func NormalizePort(port string) string {
...
+	}
+	var port string
+	switch {
+	case u.Scheme == "http":
+		port = "80"
+	case u.Scheme == "https":
+		port = "443"
+	default:
+		// Unsupported scheme for valid url s
+		log.Errf("Unsupported URL scheme: %s", u.Scheme)
```
ldemailly: it's perfectly valid to skip http:// in current fortio; doesn't want to loose that

fnet/network.go
```diff
@@ -34,6 +35,91 @@ func NormalizePort(port string) string {
...
+				u.Host += NormalizePort(port)
+				return u.String()
+			}
+			// The bracketed IPv6 address is invalid,
+			// return s unmodified.
+			log.Errf("Invalid IPv6 address wrapped in brackets: %v", u.Host)
+			return s
+		}
+		// Check if the host portion of s is a valid domain name.
+		_, err := net.LookupHost(u.Host)
```
ldemailly: I don't want to lookup multiple times; what for ?

fortio_main.go
```diff
@@ -323,9 +323,9 @@ func grpcClient() {
 	tls := *grpcSecureFlag
 	var err error
 	if *doHealthFlag {
-		_, err = fgrpc.GrpcHealthCheck(host, tls, *healthSvcFlag, count)
+		_, err = fgrpc.GrpcHealthCheck(fnet.AppendPort(host), tls, *healthSvcFlag, count)
```
ldemailly: this is breaking stuff, as the port is added later optionally

ldemailly (changes requested):

ldemailly:
I guess I need to add more test for stuff like

fortio load|curl localhost:8080

so those tests would warn you the PR has issues

bincommon/commonflags.go
danehans: OK. I didn't get that from #162. I'll remove.

fnet/network_test.go
danehans: OK, I'll update the func as to not append `:port` to v6 bracketized addrs.

fnet/network.go
danehans: OK. When I wrote this, I didn't realize it would be just for grpc url parsing. Let me try to have the url parsing func utilize the grpcDestination func.

bincommon/commonflags.go
ldemailly: it says grpc . also you should try/experiment: `fortio curl www.google.com` and `fortio curl http://www.google.com` and you'll see the url work without port (also see what happens to the Host: header when running with -loglevel debug)

fnet/network.go
ldemailly: even for grpc in case that's not clear - the http:// https:// prefix is a recent addition for grpc

fnet/network.go
danehans: That's originally what I expected too. I was surprised to find that `url.Parse` did not return an error when an IPv6 address was unbracketed within a url. See: https://play.golang.org/p/jqNv_1pnqvb

However, this is a moot point since I have refactored the code and dropped both func's that I originally added for this PR. Now that I have a better understanding of the requirements, I am simply updating the `grpcDestination` func with the necessary logic.

fgrpc/grpcrunner.go
```diff
@@ -168,32 +160,3 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
 	}
 	return &total, nil
 }
-
-// grpcDestination parses dest and returns dest:port based on dest type
-// being a hostname, IP address or hostname/ip:port pair.
-// TODO: change/fix this (NormalizePort and more)
-func grpcDestination(dest string) string {
```
ldemailly: can you please leave this code in fgrpc/ - it's specific to grpc no ?

danehans: It's also network specific and removes the need to import the net std lib. I'll go ahead and move it back if you prefer.

ldemailly: Yes I prefer to minimize the diff, it's easier to review, blame etc...
we can move the code later if there is value (but something named "grpcDestination" doesn't belong in fnet/)

fgrpc/grpcrunner.go
```diff
@@ -35,9 +35,11 @@ import (
 
 const (
 	// DefaultGRPCPort is the Fortio gRPC server default port number.
-	DefaultGRPCPort = "8079"
-	prefixHTTP      = "http://"
-	prefixHTTPS     = "https://"
+	DefaultGRPCPort  = "8079"
```
ldemailly, line 38: Did you check that
```
fortio load -grpc localhost
# and
fortio grpcping localhost
```
still works (and goes to localhost:8079)

fgrpc/grpcrunner.go
```diff
@@ -169,31 +171,59 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
...
+			parsedDest = "[" + ip.String() + "]" + fnet.NormalizePort(port)
+			log.Infof("grpc destination set to: %v", parsedDest)
+			return parsedDest
 		}
-
 	}
-	// dest must be in the form of hostname
-	return dest + fnet.NormalizePort(DefaultGRPCPort)
+	// Check if parsedDest is a valid domain name.
+	_, err := net.LookupHost(parsedDest)
```
ldemailly: why/what for ?

fgrpc/grpcrunner.go
```diff
@@ -169,31 +171,59 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
...
+		log.Infof("grpc destination set to: %v", parsedDest)
+		return parsedDest
 	}
-	if ip := net.ParseIP(dest); ip != nil {
+	if ip := net.ParseIP(parsedDest); ip != nil {
 		switch {
 		case ip.To4() != nil:
-			return ip.String() + fnet.NormalizePort(DefaultGRPCPort)
+			parsedDest = ip.String() + fnet.NormalizePort(port)
+			log.Infof("grpc destination set to: %v", parsedDest)
```
ldemailly: `log.LogVf`

I don't think it's useful as Info once it's working/debugged
(while developing run with -loglevel verbose to see those)

fgrpc/grpcrunner.go
```diff
@@ -169,31 +171,59 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
...
+			return parsedDest
 		case ip.To16() != nil:
-			return "[" + ip.String() + "]" + fnet.NormalizePort(DefaultGRPCPort)
+			parsedDest = "[" + ip.String() + "]" + fnet.NormalizePort(port)
+			log.Infof("grpc destination set to: %v", parsedDest)
```
ldemailly: same

fgrpc/grpcrunner_test.go
```diff
@@ -92,71 +92,111 @@ func TestGRPCDestination(t *testing.T) {
...
-			"hostname",
+			"valid hostname",
 			"localhost",
 			"localhost:8079",
 		},
 		{
-			"hostname and port",
+			"invalid hostname",
+			"lclhst",
+			"lclhst",
```
ldemailly: not useful/see above

ldemailly (commented):
almost there

fgrpc/grpcrunner.go
danehans: Yes. See the following details:

__Start Fortio Server__
```
$ docker run --rm --name fortio -p 8080:80 -p 8079:8079 istio/fortio:webtest server -loglevel debug -http-port 80
17:36:12 I logger.go:97> Log level is now 0 Debug (was 2 Info)
17:36:12 V stats.go:517> Will use [50 75 90 99 99.9] for percentiles
Fortio 0.8.1-pre grpc 'ping' server listening on [::]:8079
Fortio 0.8.1-pre https redirector server listening on [::]:8081
Fortio 0.8.1-pre echo server listening on [::]:80
UI started - visit:
http://localhost:80/fortio/   (or any host/ip reachable on this server)
17:36:12 V uihandler.go:96> Using resources directory set at link time: /usr/local/lib/fortio
17:36:12 I fortio_main.go:186> All fortio 0.8.1-pre 2018-03-26 22:14 6aa57194a7426326cdbc9c164753c96e67334242-dirty go1.10 servers started!
```

__Run `load -grpc localhost` command__
```
$ docker exec fortio /usr/local/bin/fortio grpcping localhost
17:37:07 I grpcrunner.go:199> grpc destination: localhost
17:37:07 I grpcrunner.go:227> grpc destination set to: localhost:8079
Clock skew histogram usec : count 1 avg -22.923 +/- 0 min -22.923 max -22.923 sum -22.923
>= -22.923 <= -22.923 , -22.923 , 100.00, 1
# target 50% -22.923
RTT histogram usec : count 3 avg 756.864 +/- 84.46 min 648.227 max 854.181 sum 2270.592
>= 648.227 <= 700 , 674.113 , 33.33, 1
> 700 <= 800 , 750 , 66.67, 1
> 800 <= 854.181 , 827.091 , 100.00, 1
# target 50% 750
17:37:07 I pingsrv.go:117> Ping RTT 756864 (avg of 648227, 768184, 854181 ns) clock skew -22923
```

__Run `grpcping localhost` command__
```
$ docker exec fortio /usr/local/bin/fortio load -grpc localhost
Fortio 0.8.1-pre running at 8 queries per second, 4->4 procs, for 5s: localhost
17:37:40 I grpcrunner.go:99> Starting grpc test for localhost with 4 threads at 8.0 qps
17:37:40 I grpcrunner.go:199> grpc destination: localhost
17:37:40 I grpcrunner.go:227> grpc destination set to: localhost:8079
17:37:40 I grpcrunner.go:199> grpc destination: localhost
17:37:40 I grpcrunner.go:227> grpc destination set to: localhost:8079
17:37:40 I grpcrunner.go:199> grpc destination: localhost
17:37:40 I grpcrunner.go:227> grpc destination set to: localhost:8079
17:37:40 I grpcrunner.go:199> grpc destination: localhost
17:37:40 I grpcrunner.go:227> grpc destination set to: localhost:8079
Starting at 8 qps with 4 thread(s) [gomax 4] for 5s : 10 calls each (total 40)
17:37:45 I periodic.go:530> T000 ended after 5.005335345s : 10 calls. qps=1.997868136845085
17:37:45 I periodic.go:530> T002 ended after 5.005375398s : 10 calls. qps=1.9978521499098159
17:37:45 I periodic.go:530> T001 ended after 5.005953009s : 10 calls. qps=1.9976216280938726
17:37:45 I periodic.go:530> T003 ended after 5.006542006s : 10 calls. qps=1.997386616953514
Ended after 5.006640388s : 40 calls. qps=7.9894
Sleep times : count 36 avg 0.55170978 +/- 0.001628 min 0.548772401 max 0.554760104 sum 19.8615522
Aggregated Function Time : count 40 avg 0.001034316 +/- 0.000333 min 0.000373654 max 0.001956965 sum 0.041372639
>= 0.000373654 <= 0.001 , 0.000686827 , 37.50, 15
> 0.001 <= 0.00195697 , 0.00147848 , 100.00, 25
# target 50% 0.00119139
# target 75% 0.00157418
# target 90% 0.00180385
# target 99% 0.00194165
# target 99.9% 0.00195543
Health SERVING : 40
All done 40 calls (plus 4 warmup) 1.034 ms avg, 8.0 qps
```

__Observe Server Logs__
```
17:36:51 V pingsrv.go:43> Ping called {Seq:0 Ts:0 Payload:} (ctx context.Background.WithCancel.WithCancel.WithValue(peer.peerKey{}, &peer.Peer{Addr:(*net.TCPAddr)(0xc420163950), AuthInfo:credentials.AuthInfo(nil)}).WithValue(transport.streamKey{}, <stream: 0xc4201d4500, /fgrpc.PingServer/Ping>).WithValue(metadata.mdIncomingKey{}, metadata.MD{":authority":[]string{"localhost:8079"}, "user-agent":[]string{"grpc-go/1.9.2"}}))
<SNIP>
17:37:07 V pingsrv.go:43> Ping called {Seq:0 Ts:0 Payload:} (ctx context.Background.WithCancel.WithCancel.WithValue(peer.peerKey{}, &peer.Peer{Addr:(*net.TCPAddr)(0xc420156780), AuthInfo:credentials.AuthInfo(nil)}).WithValue(transport.streamKey{}, <stream: 0xc4201d48c0, /fgrpc.PingServer/Ping>).WithValue(metadata.mdIncomingKey{}, metadata.MD{":authority":[]string{"localhost:8079"}, "user-agent":[]string{"grpc-go/1.9.2"}}))
<SNIP>
```

fgrpc/grpcrunner.go
danehans: For consistency. The code parses dest and if dest is an invalid IP, the unmodified dest is returned. I was thinking of following the same approach if dest is a domain/host name. Otherwise, we have 2 different return types for invalid dest inputs.

fgrpc/grpcrunner.go
danehans: Adding test cases to `Webtest.sh`.

ldemailly: as you can see
```
17:37:40 I grpcrunner.go:199> grpc destination: localhost
17:37:40 I grpcrunner.go:227> grpc destination set to: localhost:8079
17:37:40 I grpcrunner.go:199> grpc destination: localhost
17:37:40 I grpcrunner.go:227> grpc destination set to: localhost:8079
17:37:40 I grpcrunner.go:199> grpc destination: localhost
17:37:40 I grpcrunner.go:227> grpc destination set to: localhost:8079
17:37:40 I grpcrunner.go:199> grpc destination: localhost
17:37:40 I grpcrunner.go:227> grpc destination set to: localhost:8079
```
is way too much logging

it also shows this normalization should happen once and not once per thread

fgrpc/grpcrunner.go
ldemailly: if we have a lot of invalid replies it should return an error

but this code doesn't have to be that complicated, it's just a matter of adding :80 or :443 when it's missing / when http:// or https:// is provided as a prefix

fgrpc/grpcrunner.go
danehans: The normalization is happing once per thread b/c `grpcDestination` is being called by the `Dial` func, which gets called once per thread in `RunGRPCTest`. I think the log settings I added simply expose an existing issue. Can you create an Issue, assign me and I'll refactor in a separate PR?

fgrpc/grpcrunner.go
danehans: OK. I have further simplified the code and will submit an updated PR.

danehans:
@ldemailly I believe I addressed all your comments. Please let me know if anything else is needed to get this PR merged.

fgrpc/grpcrunner.go
```diff
@@ -35,9 +35,11 @@ import (
...
 const (
 	// DefaultGRPCPort is the Fortio gRPC server default port number.
-	DefaultGRPCPort = "8079"
-	prefixHTTP      = "http://"
-	prefixHTTPS     = "https://"
+	DefaultGRPCPort  = "8079"
+	defaultHTTPPort  = "80"
+	defaultHTTPSPort = "443"
+	prefixHTTP       = "http://"
+	prefixHTTPS      = "https://"
```
ldemailly, line 42: those must exist somewhere but I think it's fine to clone here if we haven't easily found those constants

ldemailly (approved):
thx!

the only thing would be to call grpcDestination earlier instead of dial but we can do that in follow up

fgrpc/grpcrunner.go
danehans: `prefixHTTP` and `prefixHTTP` were already part of the code. They show up in the diff b/c `=` is pushed over a couple spaces. I was thinking the same thing. I looked through the net, net/http and net/url pkgs and did not see any constants for the standard http/https ports.

danehans:
> the only thing would be to call grpcDestination earlier instead of dial but we can do that in follow up.

I'm not following. `grpcDestination` strips the http/https scheme. If we call it before `Dial`, then the logic in `Dial` used to establish a secure or insecure connection based on `https` in the `serverAddr` string would be broken.

ldemailly:
I mean that given all the client have the same logic, that logic should be moved out

like https://github.com/istio/fortio/blob/master/fhttp/http_client.go#L85
and http options init/new does normalization once before being copied in each thread

filled #222

fgrpc/grpcrunner.go
ldemailly: I meant somewhere in some go standard library

PR #219, opened 2018-03-27
Allow custom headers to be set on query for the echo server replies

### LINKED ISSUE #191: add ?header= to echosrv
@mandarjog request

ldemailly:
Fixes #191

You can specify multiple headers to be set/returned by the echo server
(including multiple for the same header key)

### DIFF
--- a/fhttp/http_test.go
+++ b/fhttp/http_test.go
@@ -745,7 +745,9 @@ func TestEchoHeaders(t *testing.T) {
 	}
 	// minimal manual encoding (only escape the space)
 	var urls []string
-	urls = append(urls, fmt.Sprintf("http://localhost:%d/echo?header=Foo:Bar1&header=Foo:Bar2&header=X:Y&header=Z:abc+def:xyz", a.Port))
+	urls = append(urls,
+		fmt.Sprintf("http://localhost:%d/echo?header=Foo:Bar1&header=Foo:Bar2&header=X:Y&header=Z:abc+def:xyz",
+			a.Port))
 	// proper encoding
 	urls = append(urls, fmt.Sprintf("http://localhost:%d/echo?%s", a.Port, v.Encode()))
 	for _, url := range urls {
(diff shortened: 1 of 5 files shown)


ldemailly:
Example:
```
$ curl -v "http://localhost:8080/?header=foo:bar1&header=foo:bar2&header=blah:xYz"
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /?header=foo:bar1&header=foo:bar2&header=blah:xYz HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Blah: xYz
< Foo: bar1
< Foo: bar2
< Date: Tue, 27 Mar 2018 00:19:26 GMT
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
```

mandarjog (approved):
lgtm

fhttp/http_test.go
```diff
@@ -872,7 +917,29 @@ func TestDefaultHeadersAndOptionsInit(t *testing.T) {
 	}
 }
 
-// --- for bench mark/comparaison
+func TestChangeMaxPayloadSize(t *testing.T) {
```
ldemailly, line 922: that part (first commit) is just a move of that test to the right section 
see https://github.com/istio/fortio/pull/219/commits/b22958f8c3ab0a8a87fef075ebb0d5f7bf7dfff8 for the interesting part

ldemailly:
I am wondering if I should make those propabilities too ? (ie Foo:Bar:2 would put Foo: Bar in 2% of replies - there would be an issue for multi header on whether the dice applies independently for each or not)

danehans:
> I am wondering if I should make those propabilities too ? (ie Foo:Bar:2 would put Foo: Bar in 2% of replies 

I think that is a cool sub feature, but is not required to address this issue. I would open a new issue for the sub feature and see if anyone weighs in and/or contributes code.

ldemailly:
will refactor in #224 though I am a bit concerned about potential break in api

PR #225, opened 2018-04-02
Adding grpc multiple stream support (-s) and max server side stream (-grpc-max-streams),...

### LINKED ISSUE #223: add a multiplexed / async / streaming mode to grpc load
To test limits such as 100 streams / connection

cc @costinm

ldemailly:
- `-s` (or `Streams` in GRPCRunner options) to multiply number of connections by given number of streams (multiplex)
- `-grpc-max-streams` option to the grpc ping server
- `-grpc-ping-delay` option to the ping client (and server) to delay the connection
- `-ping` to use ping instead of health as the grpc echo server (ie should be fortio on the backend)

Fixes #223; Manually tested number of sockets/connections (using lsof…) + unit tests

Also
- switches to go1.10.1 for build/release
- vendor update for 0.9
- Added UI for grpc-ping-delay and ping mode
- Improved JSON results with RunType (http/grpc differentiation and grpc mode details)


fgrpc/grpcrunner_test.go
```diff
@@ -55,9 +55,53 @@ func TestGRPCRunner(t *testing.T) {
 	}
 }
 
+//TODO : find something that actually fails based on MAXSTREAM
+func TestGRPCRunnerMaxStreams(t *testing.T) {
```
danehans: It would be nice to see this as a table driven test.

ldemailly: which table though ? there isn't much more than 1 thing to test

ldemailly:
k it's ready to go now, PTAL,
preview image: docker.io `istio/fortio:0.9.0-pre1`

ldemailly:
it's also live on https://fortio-stage.istio.io/
(through https://github.com/istio/admin-sites/pull/63 )

danehans (commented):
I'm having a difficult time understanding how to use the new flags. I can only imagine what it's like for a new user. It would be helpful to add example usages to the readme.

I think it would also be helpful to organize the cli options using https://github.com/spf13/cobra. I know you want to limit deps, but cobra is widely used, well supported and will greatly improve cli usability.

README.md
```diff
@@ -127,6 +132,8 @@ target is a url (http load tests) or host:port (grpc health test).  flags are:
 	List of pXX to calculate (default "50,75,90,99,99.9")
   -payload string
 	Payload string to send along
+  -ping
```
danehans, line 135: I thought the `load` command with the `-grpc` flag is responsible for performing a grpc load test?

danehans:
What is the expected behavior if `-grpc-max-streams` is set to a smaller number than `-S` on the client? I expected to see an error or failure, but the test case worked:
```
$ docker run -d --ulimit nofile=20 --name fortio_server istio/fortio:0.9.0-pre1 server -grpc-max-streams 5 -loglevel debug

$ docker logs -f fortio_server
18:03:04 I logger.go:97> Log level is now 0 Debug (was 2 Info)
18:03:04 V stats.go:517> Will use [50 75 90 99 99.9] for percentiles
18:03:04 I pingsrv.go:66> Setting grpc.MaxConcurrentStreams server to 5
<SNIP>

$ docker exec fortio_server /usr/local/bin/fortio load -grpc -ping -s 10 -c 8 -grpc-ping-delay 2s localhost:8079
Fortio 0.9.0-pre running at 8 queries per second, 4->4 procs, for 5s: localhost:8079
18:23:49 I grpcrunner.go:139> Starting GRPC Ping Delay=2s test for localhost:8079 with 10*8 threads at 8.0 qps
```

README.md
ldemailly: `fortio load -grpc` does health check grpc, `fortio load -grpc -ping` will now do ping instead of health grpc load

ldemailly:
> I'm having a difficult time understanding how to use the new flags. I can only imagine what it's like for a new user. It would be helpful to add example usages to the readme.

yes better docs are always welcome :-)

> What is the expected behavior if -grpc-max-streams is set to a smaller number than -S on the client? I expected to see an error or failure, but the test case worked:

as you can see in the unit test, the behavior is just it slows down the additional streams, it's a go client feature to wait for a stream - there may/should be an option to grpc to tell it to fail instead of wait but I didn't search for that yet - can you ?

README.md
danehans: ack

danehans:
@ldemailly why did the deps need to be upgraded to support this issue?

ldemailly:
> @ldemailly why did the deps need to be upgraded to support this issue?

it doesn't need to, I just want to rev up the vendor about once a month / as part of minor version bump and this PR has enough changes to warrant bumping to 0.9 (if we can get yours in too that'd be better)

danehans:
@ldemailly OK, I wanted to make sure I wasn't missing something in my review. I'm used to seeing separate dep bumps.

/lgtm

mandarjog (approved):
lgtm

PR #229, opened 2018-04-06
Trims any trailing slashes in grpcDest func

danehans:
Removes a trailing slash ("/") from an IPv4/IPv6 address or hostname in a url provided to the grpcDestination function.

Fixes Issue https://github.com/istio/fortio/issues/226

### DIFF
--- a/fgrpc/grpcrunner.go
+++ b/fgrpc/grpcrunner.go
@@ -264,36 +264,21 @@ func grpcDestination(dest string) (parsedDest string) {
 		parsedDest = dest
 		port = DefaultGRPCPort
 	}
-	host, sPort, err := net.SplitHostPort(parsedDest)
-	// parsedDest is in the form of host:port or ip:port
-	if err == nil {
-		if ip := net.ParseIP(host); ip != nil {
-			switch {
-			case ip.To4() != nil:
-				parsedDest = ip.String() + fnet.NormalizePort(sPort)
-				return parsedDest
-			case ip.To16() != nil:
-				parsedDest = "[" + ip.String() + "]" + fnet.NormalizePort(sPort)
-				return parsedDest
-			}
-		}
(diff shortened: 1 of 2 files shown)


fgrpc/grpcrunner.go
```diff
@@ -251,34 +251,49 @@ func grpcDestination(dest string) (parsedDest string) {
...
 		log.Infof("stripping https scheme. grpc destination: %v. grpc port: %s",
 			parsedDest, port)
 	default:
 		parsedDest = dest
 		port = DefaultGRPCPort
 	}
-	if _, _, err := net.SplitHostPort(parsedDest); err == nil {
+	host, sPort, err := net.SplitHostPort(parsedDest)
+	// parsedDest is in the form of host:port or ip:port
+	if err == nil {
```
ldemailly: why is all this coming back/changing - should just be a simple change with just the suffix trim above ?

fgrpc/grpcrunner_test.go
```diff
@@ -205,11 +200,6 @@ func TestGRPCDestination(t *testing.T) {
 			"https://2001:dba::1",
 			"[2001:dba::1]:443",
 		},
-		{
-			"IPv6 address with https prefix and trailing /",
-			"https://2001:dba::1/",
```
ldemailly: this should work ? (not a high pri but why wouldn't it ?)

ldemailly (approved):
please add back the "https://2001:dba::1/" test (or http or both whichever/one of those)

ldemailly:
and "http://1.2.3.4/" while at it (-> 1.2.3.4:80 )
basically any test that had `http://_...anything..._` should also pass with `http://_...anything..._/` and ditto with https. but given the code is so simple it's also ok to not have 100 tests testing the same thing (that / is removed)

PR #232, opened 2018-04-16
Allow disabled for grpc port too to skip without printing an error

ldemailly:
So this works without spurious output:
```
$ fortio server --quiet --redirect-port "disabled" --grpc-port "disabled" -http-port 0 -ui-path "" | awk '{print $7}'
[::]:64474
```

Manually tested


ldemailly:
no bots here :-) only use the github UI :-) 

I'm considering adding the server/daemon to this PR so not merging yet

PR #236, opened 2018-04-24
Show dest ip

ldemailly:
Eg.
```
15:25:37 V network.go:148> Proxy: Accepted proxy connection from
100.123.127.245:59765 -> 100.123.127.245:8088 (for [::1]:8080)
```


ldemailly:
@andraxylia this helps show which IP is used as destination as seen by receiving end
(and shows I was wrong about cluster ip vs pod ip! my bad!), ptal

andraxylia (approved):

PR #237, opened 2018-04-24
Add second QPS axis to multichart

### LINKED ISSUE #211: graph QPS on secondary Y axis
for multi graphs

Tahler:
Resolves #211.

image

### DIFF
--- a/ui/static/js/fortio_chart.js
+++ b/ui/static/js/fortio_chart.js
@@ -346,6 +346,7 @@ function fortioAddToMultiResult (i, res) {
   findData(5, i, res, '99')
   findData(6, i, res, '99.9')
   mchart.data.datasets[7].data[i] = 1000.0 * res.DurationHistogram.Max
+  mchart.data.datasets[8].data[i] = res.ActualQPS
 }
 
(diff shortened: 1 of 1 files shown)


ui/static/js/fortio_chart.js
```diff
@@ -372,7 +373,7 @@ function deleteSingleChart () {
   chart = {}
 }
 
-function makeMultiChart (data) {
```
ldemailly, line 375: I know data happens to also be global but isn't it cleaner to pass it anyway ?

ldemailly:
thx for this ! see comment/question

ui/static/js/fortio_chart.js
Tahler: `data` is not used in `makeMultiChart`. The data for the chart is added over time with `fortioAddToMultiResult`

ldemailly: oh ! :( my bad ! ok some day we should objectify this whole thing but separately from adding features

ldemailly (approved):

PR #240, opened 2018-04-26
Toggle QPS axis on QPS hide

### LINKED ISSUE #239: qps axis - any way to delete it when it's off (or blank out the data)
screen shot 2018-04-25 at 5 23 03 pm

when turning off (by clicking) the QPS data set, the right side changes to [0 - 1.0] which is a bit confusing - is it possible to change some setting in the axis itself (maybe there is a flag/option) or actively remove it if not ?

Tahler:
Resolves #239.

Found an undocumented, useful onClick function for the legend!

### DIFF
--- a/ui/static/js/fortio_chart.js
+++ b/ui/static/js/fortio_chart.js
@@ -492,6 +492,18 @@ function makeMultiChart () {
       }
     }
   })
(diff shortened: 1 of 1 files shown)


ldemailly (approved):
awesome

PR #243, opened 2018-05-01
Fix lag issue

ldemailly:
by postponing url update to after 0.75 s  any change

### DIFF
--- a/ui/static/js/fortio_chart.js
+++ b/ui/static/js/fortio_chart.js
@@ -446,9 +446,13 @@ function getCurrentChart () {
   return currentChart
 }
 
+var timeoutID = 0
 function updateChart (chart = getCurrentChart()) {
-  updateQueryString()
   updateChartOptions(chart)
+  if (timeoutID > 0) {
+    clearTimeout(timeoutID)
+  }
+  timeoutID = setTimeout("updateQueryString()", 750)
 }
 
 function multiLabel (res) {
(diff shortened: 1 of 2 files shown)


ldemailly:
turns out animation onComplete isn't called (it's only if you set a progress bar like animation)

ldemailly:
@Tahler I'm merging as I need a release for kubecon but please have a look anyway/ we can further refine later

ui/static/js/fortio_chart.js
```diff
@@ -446,9 +446,13 @@ function getCurrentChart () {
...
 }
 
+var timeoutID = 0
 function updateChart (chart = getCurrentChart()) {
-  updateQueryString()
   updateChartOptions(chart)
+  if (timeoutID > 0) {
+    clearTimeout(timeoutID)
+  }
+  timeoutID = setTimeout("updateQueryString()", 750)
```
Tahler, line 455: I think it'd be more kosher to hook into the afterAnimation function that we found but seems good enough for the state of things.

Tahler (commented):
LGTM

ui/static/js/fortio_chart.js
ldemailly: except I tried and it didn't work (because I think it's a different animation : https://github.com/istio/fortio/pull/243#issuecomment-385668354

PR #249, opened 2018-05-21
Danehans grpc certs

ldemailly:
Updates to #204 which can’t be loaded

* 2nd make test would error out because of cert-tmp

* webtest broken when interrupted and reran:
++ docker create -v /etc/ssl/certs/fortio --name certs istio/fortio.build:v7 /bin/true
Error response from daemon: Conflict. The container name "/certs" is already in use by container "e3dfe8780266c10e2b39af72ac23f02abb2109f8b626570e67c031e4ca227161". You have to remove (or rename) that container to be able to reuse that name.

* fix duplication and problem with test

* sync code flags with README (run `release/updateFlags.sh` and copy paste)

also while at it:

* switch to 0.11.0 as this quite a big change (and incompatible because of no more `-grpc-secure` flag)


ldemailly:
@danehans check the last (2 so far but will do a third one for the unit test) commits for your information

danehans (approved):
Thanks for tying up the loose ends.

/lgtm

Makefile
```diff
@@ -110,7 +128,7 @@ all: test install lint docker-version docker-push-internal
 
 # Makefile should be edited first
 FILES_WITH_IMAGE:= .circleci/config.yml Dockerfile Dockerfile.echosrv \
-	Dockerfile.test Dockerfile.fcurl release/Dockerfile.in
+	Dockerfile.test Dockerfile.fcurl release/Dockerfile.in /etc/ssl/certs/ca-certificates.crt
```
ldemailly, line 131: @danehans what is this change ? it's breaking the update tag for build image... I missed that change during review, will revert now (that list of files is all the files which reference the `fortio.build:vX` image

danehans: I'm not 100% sure. I probably read `# Makefile should be edited first` and thought it was required to run the following `Webtest.sh` tests added by this PR:
```
docker exec $DOCKERNAME /usr/local/bin/fortio grpcping -cacert $CERT fortio.istio.io:443
docker exec $DOCKERNAME /usr/local/bin/fortio grpcping -cacert $CERT https://fortio.istio.io
```

If `Webtest.sh` passes, then it's definitely not needed. It might be a good idea to add your explanation above before `FILES_WITH_IMAGE` in the Makefile to avoid potential confusion of future contributors. I apologize for causing any problems.

ldemailly: ic I'll improve the comment, it was all about changing the build image but it's not obvious

PR #254, opened 2018-06-02
Move some uihandler methods to another files and add their test.

olimpias:
Two uihandler methods are moved and their tests are added.

### DIFF
--- a/fhttp/http_test.go
+++ b/fhttp/http_test.go
@@ -941,6 +941,24 @@ func TestChangeMaxPayloadSize(t *testing.T) {
 	}
 }
 
+func TestAddHTTPS(t *testing.T) {
+	var tests = []struct {
+		input    string
+		expected string
+	}{
+		{"www.google.com", "https://www.google.com"},
+		{"http://www.google.com", "https://www.google.com"},
+		{"https://www.google.com", "https://www.google.com"},
+	}
+
+	for _, test := range tests {
+		output := AddHTTPS(test.input)
+		if output != test.expected {
+			t.Errorf("%s was expected but %s is received ", test.expected, output)
+		}
+	}
+}
+
 // --- for bench mark/comparison
 
 func asciiFold0(str string) []byte {
--- a/fhttp/http_test.go
+++ b/fhttp/http_test.go
@@ -951,8 +951,11 @@ func TestAddHTTPS(t *testing.T) {
 		{"https://foo.com", "https://foo.com"},
 		{"hTTps://foo.com", "https://foo.com"},
 		{"hTTp://foo.com", "https://foo.com"},
+		{"http://fOO.com", "https://fOO.com"},
 		{"hTTp://foo.com/about.html", "https://foo.com/about.html"},
 		{"hTTp://foo.com/ABOUT.html", "https://foo.com/ABOUT.html"},
+		{"hTTp://foo.com/BaR", "https://foo.com/BaR"},
+		{"https://foo.com/BaR", "https://foo.com/BaR"},
 	}
 
 	for _, test := range tests {
--- a/fhttp/http_test.go
+++ b/fhttp/http_test.go
@@ -946,15 +946,17 @@ func TestAddHTTPS(t *testing.T) {
 		input    string
 		expected string
 	}{
-		{"www.google.com", "https://www.google.com"},
-		{"http://www.google.com", "https://www.google.com"},
-		{"https://www.google.com", "https://www.google.com"},
+		{"foo.com", "https://foo.com"},
+		{"http://foo.com", "https://foo.com"},
+		{"https://foo.com", "https://foo.com"},
+		{"hTTps://foo.com", "https://foo.com"},
+		{"hTTp://foo.com", "https://foo.com"},
 	}
 
 	for _, test := range tests {
 		output := AddHTTPS(test.input)
 		if output != test.expected {
-			t.Errorf("%s was expected but %s is received ", test.expected, output)
+			t.Errorf("%s is received but %s was expected", output, test.expected)
 		}
 	}
 }
--- a/fhttp/http_test.go
+++ b/fhttp/http_test.go
@@ -946,6 +946,7 @@ func TestAddHTTPS(t *testing.T) {
 		input    string
 		expected string
 	}{
+		{"foo", "https://foo"},
 		{"foo.com", "https://foo.com"},
 		{"http://foo.com", "https://foo.com"},
 		{"https://foo.com", "https://foo.com"},
@@ -956,6 +957,7 @@ func TestAddHTTPS(t *testing.T) {
 		{"hTTp://foo.com/ABOUT.html", "https://foo.com/ABOUT.html"},
 		{"hTTp://foo.com/BaR", "https://foo.com/BaR"},
 		{"https://foo.com/BaR", "https://foo.com/BaR"},
+		{"http", "https://http"},
 	}
 
 	for _, test := range tests {
--- a/fhttp/http_test.go
+++ b/fhttp/http_test.go
@@ -951,6 +951,8 @@ func TestAddHTTPS(t *testing.T) {
 		{"https://foo.com", "https://foo.com"},
 		{"hTTps://foo.com", "https://foo.com"},
 		{"hTTp://foo.com", "https://foo.com"},
+		{"hTTp://foo.com/about.html", "https://foo.com/about.html"},
+		{"hTTp://foo.com/ABOUT.html", "https://foo.com/ABOUT.html"},
 	}
 
 	for _, test := range tests {
--- a/fhttp/http_utils.go
+++ b/fhttp/http_utils.go
@@ -455,7 +455,7 @@ const (
 	httpsPrefix = "https://"
 )
 
-// addHTTPS replaces "http://" in url with "https://" or prepends "https://"
+// AddHTTPS replaces "http://" in url with "https://" or prepends "https://"
 // if url does not contain prefix "http://".
 func AddHTTPS(url string) string {
 	if len(url) > len(httpPrefix) {
--- a/fhttp/http_utils.go
+++ b/fhttp/http_utils.go
@@ -453,6 +453,7 @@ func OnBehalfOf(o *HTTPOptions, r *http.Request) {
 // addHTTPS replaces "http://" in url with "https://" or prepends "https://"
 // if url does not contain prefix "http://".
 func AddHTTPS(url string) (pURL string) {
+	url = strings.ToLower(url)
 	if strings.HasPrefix(url, "http://") {
 		log.Infof("Replacing http scheme with https for url: %s", url)
 		pURL = strings.TrimPrefix(url, "http://")
--- a/fhttp/http_utils.go
+++ b/fhttp/http_utils.go
@@ -451,27 +451,25 @@ func OnBehalfOf(o *HTTPOptions, r *http.Request) {
 }
 
 const (
-	httpText  = "http://"
-	httpsText = "https://"
+	httpPrefix  = "http://"
+	httpsPrefix = "https://"
 )
 
 // addHTTPS replaces "http://" in url with "https://" or prepends "https://"
 // if url does not contain prefix "http://".
-func AddHTTPS(url string) (pURL string) {
-	if len(url) > len(httpText) {
-		if strings.EqualFold(url[0:len(httpText)], httpText) {
+func AddHTTPS(url string) string {
+	if len(url) > len(httpPrefix) {
+		if strings.EqualFold(url[:len(httpPrefix)], httpPrefix) {
 			log.Infof("Replacing http scheme with https for url: %s", url)
-			pURL = url[len(httpText):]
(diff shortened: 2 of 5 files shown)


olimpias:
Hi @ldemailly . I moved AddHttps to http_utils file and SetHostAndPort method to network file. Also added their test methods. I think there is a bug on SetHostPort. You have asked it in closed PR. The problem, even I add ip address to its input, the result returns with localhost. For example, I input 8080, 192.168.31.50:8080 to SetHostAndPort method. It returns localhost:8080.  Please let me know, which methods should I move to http_utils in the uihandler.go file. Thanks

fhttp/http_test.go
```diff
@@ -941,6 +941,24 @@ func TestChangeMaxPayloadSize(t *testing.T) {
...
+	}{
+		{"www.google.com", "https://www.google.com"},
+		{"http://www.google.com", "https://www.google.com"},
+		{"https://www.google.com", "https://www.google.com"},
+	}
+
+	for _, test := range tests {
+		output := AddHTTPS(test.input)
+		if output != test.expected {
+			t.Errorf("%s was expected but %s is received ", test.expected, output)
```
ldemailly: go convention is to print actual first, before expected (I don't think it's very good or important but so far it's been consistently applied)

fhttp/http_test.go
```diff
@@ -941,6 +941,24 @@ func TestChangeMaxPayloadSize(t *testing.T) {
...
 
+func TestAddHTTPS(t *testing.T) {
+	var tests = []struct {
+		input    string
+		expected string
+	}{
+		{"www.google.com", "https://www.google.com"},
+		{"http://www.google.com", "https://www.google.com"},
+		{"https://www.google.com", "https://www.google.com"},
+	}
```
ldemailly: in theory we'd add mixed case e.g `hTTPs://foo.com/` but I don't think it'll pass in that case so not yet

ldemailly (commented):
will check hostandport and your question later, but for starter:

fhttp/http_test.go
olimpias: To pass it, I added 	pURL = strings.ToLower(pURL) in the beginning of the AddHTTPS method.

ldemailly: that would break things (url params are case sensitive)

olimpias: While I was doing that I didn't think the param cases ;). I have added a method and test cases for it as well.

fhttp/http_test.go
```diff
@@ -941,6 +941,28 @@ func TestChangeMaxPayloadSize(t *testing.T) {
...
 
+func TestAddHTTPS(t *testing.T) {
+	var tests = []struct {
+		input    string
+		expected string
+	}{
+		{"foo.com", "https://foo.com"},
+		{"http://foo.com", "https://foo.com"},
+		{"https://foo.com", "https://foo.com"},
+		{"hTTps://foo.com", "https://foo.com"},
```
ldemailly: add /BaR there to confirm when it's already https it's also preserving the case
(and/or fOO.com)

fhttp/http_utils.go
```diff
@@ -449,3 +451,39 @@ func NewHTMLEscapeWriter(w io.Writer) io.Writer {
...
+	}
+	// url must not contain any prefix, so add https prefix
+	log.Infof("Prepending https:// to url: %s", url)
+	pURL = "https://" + url
+	return pURL
+}
+
+// toLowerCasePrefix makes prefix lower case in s. If prefix is not found in s
+// returns input s and false otherwise, returns s with lowered case prefix
+func toLowerCasePrefix(s string, prefix string) (string, bool) {
```
ldemailly: this is unnecessarily expensive/complex

use strings.EqualFold() instead

fnet/network.go
```diff
@@ -160,3 +160,20 @@ func Proxy(port string, dest *net.TCPAddr) *net.TCPAddr {
 func ProxyToDestination(listenPort string, destination string) *net.TCPAddr {
 	return Proxy(listenPort, ResolveDestination(destination))
 }
+
+// Join the input Host And Port in the form of hostname:port, ip:port or :port. And returns
+func JoinHostAndPort(inputPort string, addr *net.TCPAddr) string {
+	urlHostPort := inputPort
+	portStr := ""
```
ldemailly: this used to be
```
portStr := inputPort
```
in the original code, I doubt it works the same without it ?

fnet/network_test.go
```diff
@@ -169,6 +169,37 @@ func TestResolveIpV6(t *testing.T) {
...
+	}{
+		{":8080", &net.TCPAddr{
+			IP:   []byte{192, 168, 2, 3},
+			Port: 8081,
+		}, "192.168.2.3:8081"},
+		{":8081", &net.TCPAddr{
+			IP:   []byte{192, 168, 30, 14},
+			Port: 8080,
+		}, "192.168.30.14:8080"},
+		{":8080",
```
ldemailly: missing the case for say "8080" (no : ) and nil address
(which will fail to do localhost:8080 )

ldemailly (changes requested):

fnet/network.go
olimpias: I used it for different purpose. portStr indicates the port that gathered from net.TCPAddr

fhttp/http_utils.go
```diff
@@ -449,3 +449,29 @@ func NewHTMLEscapeWriter(w io.Writer) io.Writer {
 func OnBehalfOf(o *HTTPOptions, r *http.Request) {
 	_ = o.AddAndValidateExtraHeader("X-On-Behalf-Of: " + r.RemoteAddr)
 }
+
+const (
```
ldemailly, line 453: we have those constants a few places already, notably fgrpc/grpcrunner.go - I guess we can clean that up later / put them in a common place without adding dependencies (maybe fnet/)

fnet/network.go
```diff
@@ -160,3 +160,20 @@ func Proxy(port string, dest *net.TCPAddr) *net.TCPAddr {
 func ProxyToDestination(listenPort string, destination string) *net.TCPAddr {
 	return Proxy(listenPort, ResolveDestination(destination))
 }
+
+// Join the input Host And Port in the form of hostname:port, ip:port or :port. And returns
+func JoinHostAndPort(inputPort string, addr *net.TCPAddr) string {
```
ldemailly: I checked the caller in uihandler; addr is never nil so the code simplifies to:
```golang
	urlHostPort := addr.String()
	if strings.HasPrefix(inputPort, ":") || !strings.Contains(inputPort, ":") {
            urlHostPort = fmt.Sprintf("localhost:%d", addr.Port)
	}
	return urlHostPort
```

let's rename this function to `NormalizeHostPort` and the comment to
```
// NormalizeHostPort generates host:port string for the address or uses localhost instead of [::] 
// when the original port binding input didn't specify an address
```

ps: the linter should have complained that the comment doesn't match the function

fnet/network_test.go
ldemailly: remove the nil case but do keep 8080 :8080 and [::1]:8080 as inputPort tests

fhttp/http_test.go
```diff
@@ -941,6 +941,31 @@ func TestChangeMaxPayloadSize(t *testing.T) {
 	}
 }
 
+func TestAddHTTPS(t *testing.T) {
+	var tests = []struct {
+		input    string
+		expected string
+	}{
+		{"foo.com", "https://foo.com"},
```
ldemailly: add a test for `foo` (shorter than prefix) and maybe `http` which should turn into https://http (if a host name was http)

fhttp/http_utils.go
```diff
@@ -449,3 +449,29 @@ func NewHTMLEscapeWriter(w io.Writer) io.Writer {
...
+	httpText  = "http://"
+	httpsText = "https://"
+)
+
+// addHTTPS replaces "http://" in url with "https://" or prepends "https://"
+// if url does not contain prefix "http://".
+func AddHTTPS(url string) (pURL string) {
+	if len(url) > len(httpText) {
+		if strings.EqualFold(url[0:len(httpText)], httpText) {
```
ldemailly: remove `0` it's the default (afaik)

fhttp/http_utils.go
```diff
@@ -449,3 +449,29 @@ func NewHTMLEscapeWriter(w io.Writer) io.Writer {
 func OnBehalfOf(o *HTTPOptions, r *http.Request) {
 	_ = o.AddAndValidateExtraHeader("X-On-Behalf-Of: " + r.RemoteAddr)
 }
+
+const (
+	httpText  = "http://"
```
ldemailly: do rename to `httpPrefix` and `httpsPrefix` respectively

fhttp/http_utils.go
```diff
@@ -449,3 +449,29 @@ func NewHTMLEscapeWriter(w io.Writer) io.Writer {
...
+	httpsText = "https://"
+)
+
+// addHTTPS replaces "http://" in url with "https://" or prepends "https://"
+// if url does not contain prefix "http://".
+func AddHTTPS(url string) (pURL string) {
```
ldemailly: remove pURL just string return, no need to name it specially given it's not always what is returned

fhttp/http_utils.go
```diff
@@ -449,3 +449,29 @@ func NewHTMLEscapeWriter(w io.Writer) io.Writer {
...
+	if len(url) > len(httpText) {
+		if strings.EqualFold(url[0:len(httpText)], httpText) {
+			log.Infof("Replacing http scheme with https for url: %s", url)
+			pURL = url[len(httpText):]
+			return httpsText + pURL
```
ldemailly: just directly
```
return httpsPrefix + url[len(httpText):]
```

fhttp/http_utils.go
```diff
@@ -449,3 +449,29 @@ func NewHTMLEscapeWriter(w io.Writer) io.Writer {
...
+		}
+		// return url unchanged since it already has "https://"
+		if strings.EqualFold(url[0:len(httpsText)], httpsText) {
+			pURL = url[len(httpsText):]
+			return httpsText + pURL
```
ldemailly: same and update the comment that we return normalized lowercase https prefix

fhttp/http_utils.go
```diff
@@ -449,3 +449,29 @@ func NewHTMLEscapeWriter(w io.Writer) io.Writer {
...
+		}
+	}
+	// url must not contain any prefix, so add https prefix
+	log.Infof("Prepending https:// to url: %s", url)
+	pURL = httpsText + url
+	return pURL
```
ldemailly: same, just 1 line instead of 2:
```
return httpsPrefix + url
```

though again I thought the linters would catch/not like that and force it to be HTTPSPrefix ?

fhttp/http_test.go
olimpias: I might miss understand your comment here about shorter than prefix :).

fnet/network.go
```diff
@@ -161,19 +161,12 @@ func ProxyToDestination(listenPort string, destination string) *net.TCPAddr {
...
-	if !strings.Contains(inputPort, ":") {
-		inputPort = ":" + inputPort
-	}
-	if !strings.HasPrefix(portStr, ":") {
-		urlHostPort = "localhost" + inputPort
+// NormalizeHostPort generates host:port string for the address or uses localhost instead of [::]
+// when the original port binding input didn't specify an address
+func NormalizeHostPort(inputPort string, addr *net.TCPAddr) string {
+	urlHostPort := addr.String()
+	if strings.HasPrefix(inputPort, ":") || !strings.Contains(urlHostPort, ":") {
```
ldemailly: both should be inputPort; basically unless the input had a addr:port it will be changed to localhost:port

ldemailly: see
https://github.com/istio/fortio/pull/257/files#diff-61a4e729b116ffe8a0303004555c636c

fnet/network_test.go
```diff
@@ -169,6 +169,40 @@ func TestResolveIpV6(t *testing.T) {
...
+func TestJoinHostAndPort(t *testing.T) {
+	var tests = []struct {
+		inputPort string
+		addr      *net.TCPAddr
+		expected  string
+	}{
+		{"8080", &net.TCPAddr{
+			IP:   []byte{192, 168, 2, 3},
+			Port: 8081,
+		}, "192.168.2.3:8081"},
```
ldemailly: this would be localhost unless you put 192.168.2.3:8081 as input (the tests need to include using addr:port)

fnet/network.go
olimpias: I see. I think, I got the idea :)

ldemailly:
linters work now:

fhttp/http_utils.go:458:1:warning: comment on exported function AddHTTPS should be of the form "AddHTTPS ..." (golint)

ldemailly (approved):

olimpias:
Thanks for your comments @ldemailly :)

ldemailly:
it's in 1.0 !

PR #258, opened 2018-06-16
Prefix headers are moved to common file

olimpias:
In #254 PR, decided to move prefix http to common file. They are moved to network.

### DIFF
--- a/fnet/network.go
+++ b/fnet/network.go
@@ -28,10 +28,10 @@ import (
 const (
 	// DefaultGRPCPort is the Fortio gRPC server default port number.
 	DefaultGRPCPort = "8079"
-	// DefaultHTTPPort is the Standard http port number.
-	DefaultHTTPPort = "80"
-	// DefaultHTTPSPort is the Standard https port number.
-	DefaultHTTPSPort = "443"
+	// StandardHTTPPort is the Standard http port number.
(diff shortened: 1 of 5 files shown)


ldemailly:
fix the compile failures?
usually it is best to build and test before PR

/ok-to-test

olimpias:
Yes normally I do that but I don't know why this time I jumped to build state. I thought there wont be a problem. I shouldn't have rushed. that is a lesson to me :).

fnet/network.go
```diff
@@ -25,6 +25,19 @@ import (
 	"istio.io/fortio/version"
 )
 
+const (
+	// DefaultGRPCPort is the Fortio gRPC server default port number.
+	DefaultGRPCPort = "8079"
+	// DefaultHTTPPort is the Fortio http server default port number.
```
ldemailly: that comment isnt correct 
call 80 and 443 StandardHTTP(S)Port and the comment is it is the standard

unlike 8079 which is the default port the grpc server listens to (the default http is 8080)

ldemailly (commented):
can you change the variable name too, grpc is “default” but http/https are standard ports (they are not the same meaning so they shouldn’t be called the same)

ldemailly (approved):
thanks

Issue #256, opened 2018-06-11
Support authentication


ymesika:
Something like a new `-user username:password` for http load tests.

ldemailly:
it's quite involved if supporting all the possible modes (and for the super basic one it can be done using -H)

ymesika:
Correct, your could but then you need to put there b64 encoding of that pair and google the correct header key and value format. Or, you could simply pass in an arg ;-)

ldemailly:
`fortio curl -H "Authorization: Basic $(echo user:password|base64)" host:port` should work for now

ldemailly:
I don't mind looking into adding better support - just not super high pri

olimpias:
After 258 PR, may I work on this?

ldemailly:
Sure but let’s discuss how before you write a lot of code or add any dependencies

olimpias:
Okay. Give me a day to look at the code. I will give my comment about how I can write and add dependency. Then I will wait for your feedback comment to start impl. or continue to search.

olimpias:
Here is my proposal, I'm going to add a userFlag to commonflags.go where expected input will be username:password format. I will carry userFlag on HTTPOptions by calling user as property name. In the InitHeaders method, I will check HTTPOptions's user property. If the value is not empty, I will add authentication header. I'm planning to use base64.StdEncoding.EncodeToString dependency to encode user:password. What do you think?

ymesika:
@olimpias can we have it a curl-like flag?
I.e. `-u, --user <user:password> Server user and password`

olimpias:
I haven't seen any example with double dash but `-user <user:password>` probably be okay. I hope its work for you.

ldemailly:
yes I sort of try to follow some of the curl command line option choices so `-user USER[:PASSWORD]` like curl is good; I don't think that option is so frequent/useful as to use the short `-u` for it (unlike what I did for -L for follow redirects for instance)

ps: basic auth is really not that useful but maybe it's a start

ps2: @olimpias accept the invite at https://github.com/istio/fortio/invitations so I can assign the issue to you

olimpias:
@ldemailly I accepted. As you say, I'm going to start with basic for a start. I don't know much about other possible authentication types in http except digest auth. Which type of authentication are you suggesting for http to implement?

ldemailly:
there are a bunch of schemes https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml 
[...]

PR #261, opened 2018-06-20
Basic Authentication Support

olimpias:
Basic authentication support is added.(#256)

### DIFF
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -147,10 +147,24 @@ func (h *HTTPOptions) ResetHeaders() {
 func (h *HTTPOptions) InitHeaders() {
 	h.ResetHeaders()
 	h.extraHeaders.Add("User-Agent", userAgent)
-	err := ValidateAndAddBasicAuthentication(h)
+	err := h.ValidateAndAddBasicAuthentication()
 	if err != nil {
 		log.Debugf("User credential is not valid. %v", err)
 	}
+}
+
+// ValidateAndAddBasicAuthentication validates user credentials and adds basic authentication to http header,
+// if user credentials are valid.
+func (h *HTTPOptions) ValidateAndAddBasicAuthentication() error {
+	if len(h.UserCredentials) <= 0 {
+		return nil // user credential is not entered
+	}
+	s := strings.SplitN(h.UserCredentials, ":", 2)
+	if len(s) != 2 {
+		return fmt.Errorf("invalid user credentials are used %s. Expected format user:password", h.UserCredentials)
+	}
+	h.extraHeaders.Add("Authorization", generateBase64UserCredentials(h.UserCredentials))
+	return nil
 }
 
 // GetHeaders returns the current set of headers.
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -133,6 +133,8 @@ type HTTPOptions struct {
 	// Host is treated specially, remember that one separately.
 	hostOverride   string
 	HTTPReqTimeOut time.Duration // timeout value for http request
+
+	UserCredentials string // user credentials for authorization
 }
 
 // ResetHeaders resets all the headers, including the User-Agent one.
@@ -145,6 +147,10 @@ func (h *HTTPOptions) ResetHeaders() {
 func (h *HTTPOptions) InitHeaders() {
 	h.ResetHeaders()
 	h.extraHeaders.Add("User-Agent", userAgent)
+	err := ValidateAndAddBasicAuthentication(h)
+	if err != nil {
+		log.Debugf("User credential is not valid. %v", err)
+	}
 }
 
 // GetHeaders returns the current set of headers.
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -149,7 +149,7 @@ func (h *HTTPOptions) InitHeaders() {
 	h.extraHeaders.Add("User-Agent", userAgent)
 	err := h.ValidateAndAddBasicAuthentication()
 	if err != nil {
-		log.Debugf("User credential is not valid. %v", err)
+		log.Errf("User credential is not valid. %v", err)
 	}
 }
 
--- a/fhttp/http_test.go
+++ b/fhttp/http_test.go
@@ -974,15 +974,15 @@ func TestValidateAndAddBasicAuthentication(t *testing.T) {
 		isCredentialsValid bool
 		isAuthHeaderAdded  bool
 	}{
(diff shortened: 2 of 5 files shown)


olimpias:
I couldnt figure out why shifting the sentence isn't worked as expected in README.md file. Any idea? Nevermind, I solve it from github edit. Somehow GoLand idea was showing differently

fhttp/http_client.go
```diff
@@ -145,6 +147,10 @@ func (h *HTTPOptions) ResetHeaders() {
 func (h *HTTPOptions) InitHeaders() {
 	h.ResetHeaders()
 	h.extraHeaders.Add("User-Agent", userAgent)
+	err := ValidateAndAddBasicAuthentication(h)
+	if err != nil {
+		log.Debugf("User credential is not valid. %v", err)
```
ldemailly: warning or error level

fhttp/http_client.go
```diff
@@ -145,6 +147,24 @@ func (h *HTTPOptions) ResetHeaders() {
 func (h *HTTPOptions) InitHeaders() {
 	h.ResetHeaders()
 	h.extraHeaders.Add("User-Agent", userAgent)
+	err := h.ValidateAndAddBasicAuthentication()
```
ldemailly, line 150: I'm not sure this is the right place for this, the init is the baseline for the headers
(but then again the whole header part isn't super clean at the moment)

fhttp/http_client.go
```diff
@@ -145,6 +147,24 @@ func (h *HTTPOptions) ResetHeaders() {
...
+	if err != nil {
+		log.Debugf("User credential is not valid. %v", err)
```
ldemailly: log error or warning not debug

fhttp/http_client.go
```diff
@@ -145,6 +147,24 @@ func (h *HTTPOptions) ResetHeaders() {
...
+
+// ValidateAndAddBasicAuthentication validates user credentials and adds basic authentication to http header,
+// if user credentials are valid.
+func (h *HTTPOptions) ValidateAndAddBasicAuthentication() error {
+	if len(h.UserCredentials) <= 0 {
+		return nil // user credential is not entered
+	}
+	s := strings.SplitN(h.UserCredentials, ":", 2)
+	if len(s) != 2 {
+		return fmt.Errorf("invalid user credentials are used %s. Expected format user:password", h.UserCredentials)
```
ldemailly, line 164: or log the error here

but right now it's not logged if I pass -user foo is it ? please paste the full output of
```
fortio load -user foo www.google.com
```

fhttp/http_test.go
```diff
@@ -968,6 +968,30 @@ func TestAddHTTPS(t *testing.T) {
 	}
 }
 
+func TestValidateAndAddBasicAuthentication(t *testing.T) {
+	var tests = []struct {
+		o                  HTTPOptions
+		isCredentialsValid bool
+		isAuthHeaderAdded  bool
+	}{
+		{HTTPOptions{UserCredentials: "foo:foo"}, false, true},
```
ldemailly: this should be true (valid credentials) and true (auth header added)

or your variable isCredentialsValid is wrong

ldemailly (commented):
thanks, see below:

fhttp/http_test.go
olimpias: I should have written that isCredentialsNotValid. But, I'm keeping variable name same and changing boolean values to their not

fhttp/http_client.go
olimpias:
```
Fortio 1.0.0-pre running at 8 queries per second, 4->4 procs, for 5s: www.google.com
07:53:18 I httprunner.go:82> Starting http test for www.google.com with 4 threads at 8.0 qps
07:53:18 E http_client.go:152> User credential is not valid. invalid user credentials are used foo. Expected format user:password
07:53:18 W http_client.go:105> assuming http:// on missing scheme for 'www.google.com'
Starting at 8 qps with 4 thread(s) [gomax 4] for 5s : 10 calls each (total 40)
07:53:24 I periodic.go:533> T003 ended after 5.102056123s : 10 calls. qps=1.9599941198059614
07:53:24 I periodic.go:533> T002 ended after 5.10392474s : 10 calls. qps=1.9592765390188727
07:53:24 I periodic.go:533> T001 ended after 5.104550597s : 10 calls. qps=1.959036316708685
07:53:24 I periodic.go:533> T000 ended after 5.109756345s : 10 calls. qps=1.957040478023028
Ended after 5.109886411s : 40 calls. qps=7.828
Sleep times : count 36 avg 0.44596145 +/- 0.009171 min 0.413167981 max 0.45571464 sum 16.0546124
Aggregated Function Time : count 40 avg 0.10895143 +/- 0.008826 min 0.099699288 max 0.142015374 sum 4.35805737
>= 0.0996993 <= 0.1 , 0.0998496 , 2.50, 1
> 0.1 <= 0.12 , 0.11 , 85.00, 33
> 0.12 <= 0.14 , 0.13 , 97.50, 5
> 0.14 <= 0.142015 , 0.141008 , 100.00, 1
# target 50% 0.111515
# target 75% 0.117576
# target 90% 0.128
# target 99% 0.141209
# target 99.9% 0.141935
Code 200 : 40 (100.0 %)
Response Header Sizes : count 40 avg 693 +/- 0 min 693 max 693 sum 27720
Response Body/Total Sizes : count 40 avg 11884.25 +/- 18.59 min 11853 max 11926 sum 475370
All done 40 calls (plus 4 warmup) 108.951 ms avg, 7.8 qps
```
(laurent: edited/added the backticks)

olimpias: In this output your above comment, which is changing log to err or warring from debug,  is applied.

fhttp/http_client.go
olimpias: I put header in this method because it solves the fast client as well.  We can move the method, if the header part becomes clean in the future.

fhttp/http_client.go
ldemailly: pro tip bracket command output with 3 backticks : ```
to get a pre formattted block like
```
this 
is 
an example
```

ldemailly (approved):
thank you

fhttp/http_client.go
olimpias: ``` test ```

ldemailly:
this btw doesn't work when combined with -H :/

fhttp/http_client.go
ldemailly: I should have insisted... see #282

PR #268, opened 2018-07-01
add Dockerfile.fpm that will generate .deb files using fpm

### LINKED ISSUE #246: .deb/.rpm files for easier installation
Hi all,

while you provide a zip file, I think a nice addition would be to provide .deb/.rpm files. If there is enough interest, I can hack a script around fpm (for a start) that can provide these formats. Additionally, if my issue with man page gets accepted, it can be packaged there as well.

Please let me know your thoughts.

thanasisk:
Fixes #246 

this is my 1st attempt - please provide some feedback and I can adjust it, as needed


package.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+set -e
+set -u
+
+VERSION=`/source/usr/local/bin/fortio version | cut -f 1 -d ' '`
+echo $VERSION
+# how that we have our version, let's build DEB
+cd /source
+fpm -s dir -t deb -n fortio -v $VERSION .
+cp fortio*.deb /packages
```
ldemailly, line 10: how does that get outside of the image ?

package.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
ldemailly, line 1: move (this source file) to the release/ directory

Dockerfile.fpm
```diff
@@ -0,0 +1,30 @@
+# Build the binaries in larger image
```
ldemailly, line 1: this is a dup of Dockerfile - have a look at
https://github.com/istio/fortio/blob/master/release/Dockerfile.in
and "make release" to extend/use the same pattern ?

ldemailly (changes requested):
ping ;-) ^

thanasisk:
pong - will address comments this weekend

package.sh
thanasisk: you mount packages outside of the container, in the "host"

ldemailly:
more ping :)

thanasisk:
Hi @ldemailly - I need more time, $IRL came calling - most likely I will have a 2nd look in the coming weekend. Another solution is for someone else to take over this, if time is of the essence. Let me know your thoughts ...

ldemailly:
take your time

ldemailly:
friendly ping :-)

thanasisk:
on it!

package.sh
ldemailly: is there an example / end to end of how that works, or a script for that part ? (see for instance how I get the tgz out of the image for the other binary build)

thanasisk:
I started a refactoring of the PR. Specifically:
- I moved all relevant scripts under release/
- there are 2 scripts that run inside container, one is generating RPM the other one DEB
- there are 2 scripts called from the (modified) Makefile that work similarly to the .tgz generator

(the 2 scripts above can be consolidated into one, expecting a proper command line argument - currently evaluating this to minimize code duplication)

The issue that I am facing is which container image to use - a random Ruby enabled debian/ubuntu one for .deb and a random Ruby enabled Centos/Fedora for .rpm - is there any guidelines on this topic?

ldemailly:
we could add ruby to https://github.com/istio/fortio/blob/master/Dockerfile.build if needed

but can you start with a standard image ? we must not be the first people trying to generate packages without making a custom image ?

see also what I wrote on email last week

thanasisk:
what do you mean by standard image?

ldemailly:
any "ruby" image like https://hub.docker.com/_/ruby/ (or fpm ...)

thanasisk:
on it!

package.sh
```diff
@@ -0,0 +1,10 @@
...
+set -e
+set -u
+
+VERSION=`/source/usr/local/bin/fortio version | cut -f 1 -d ' '`
```
ldemailly, line 5: `fortio version -s` prints the short version directly

package.sh
```diff
@@ -0,0 +1,10 @@
...
+echo $VERSION
+# how that we have our version, let's build DEB
+cd /source
+fpm -s dir -t deb -n fortio -v $VERSION .
```
ldemailly, line 9: how does that handle the `fortio.org/fortio/ui.resourcesDir=$(LIB_DIR)` files?

ldemailly:
going to close this as it went stale, I did it in a more integrated with the existing `make release` in #302 but thanks for the pointer to fpm

PR #269, opened 2018-07-02
Merge vendor submodule

Tahler:
- Preserves the history from the vendor repository
- Matches the setup on istio/istio
- Removes now irrelevant Makefile recipes (e.g. `make submodule-sync`)


ldemailly:
Uh, no/why?

Tahler:
I've just found the sub-module process more than annoying the couple times I've added dependencies in my fork where I'm adding the mock service stuff. I did this in my own fork and thought since I'd done the work I'd share it as a PR.

From a git perspective, keeping them as one makes the history cleaner. For example, when I add the Prometheus client library as a dependency and the initial application code that uses it, the intention is clear, rather than this three step dance where the history is split.

Not a huge deal but its definitely a headache to work with and it's pretty unusual. Dep doesn't make any recommendations for this type of structure (also see https://github.com/golang/dep/blob/master/docs/FAQ.md#will-dep-let-me-use-git-submodules-to-store-dependencies-in-vendor) and I don't know of any other Go repositories that use vendors as submodules.

ldemailly:
I'm am really philosophically opposed to having foreign, unrelated and duplicated code checked in in the same repo. it's a transient "youth error" for golang to not have a better system to manage dependencies (imo/hopefully).

Also I don't want to add any dependencies to fortio :-) so making it easy to do so is a non goal for me :) or even I would say making it hard is a benefit.

I do agree the red part of your PR is good / simpler - but at the cost of checking in more code in fortio than fortio itself

Let's find a creative way for fortio to be extended with additional - optional - dependencies ?
(in the meantime just to get cranking / for proof of concept you can use your fork but I would like to make it a requirement it's an extension/plugin/optional component if at all possible before bloating fortio with new dependencies)

Tahler:
I completely agree with your comments on Go's immaturity towards versioning / dependency management and the whole silliness of checking in dependencies and putting the burden on version control and bloating the repository. It's all horrible but submoduling just increases headache. The real gate-keeping happens at the PR, where the reviewer would notice dependencies are being added to vendor and could refuse the merge.

As far as keeping binary/image size down, it can only increase when dependencies are imported for the compiler. I'm adding a separate binary called "mocksrv" which imports some of my own code, some of Fortio's libraries, and the Prometheus client library. Its size and dependencies do not influence other binary sizes.

Later on when I'm adding Prometheus to the load testing application there's going to have to be something like a metrics interface to keep Prometheus optional. One binary will pass an implementation like the current behavior and the other binary will pass an implementation using Prometheus.

ldemailly:
By the way fcurl is an example of smaller (even) binary/image with less dependencies (no grpc, it's about 1/2 the size) and built differently - though not quite a module/plugin/extension

Tahler:
Right. And same to echosrv and my WIP mocksrv, since I'm not adding it to fortio_main.go

ldemailly:
going to close this one if you don’t mind

PR #270, opened 2018-07-05
Payload size flag is added for GRPC.

olimpias:
Using payload size flag, payload field will be generated randomly as a string for grpc ping message.
Random payload generator method is moved to network from http_server for reusing.


fgrpc/pingsrv.go
```diff
@@ -92,11 +94,14 @@ func PingServer(port, cert, key, healthServiceName string, maxConcurrentStreams
 
 // PingClientCall calls the ping service (presumably running as PingServer on
 // the destination). returns the average round trip in seconds.
-func PingClientCall(serverAddr, cacert string, n int, payload string, delay time.Duration) (float64, error) {
+func PingClientCall(serverAddr, cacert string, n int, payload string, payloadSize int, delay time.Duration) (float64, error) {
 	conn, err := Dial(serverAddr, cacert, "") // somehow this never seem to error out, error comes later
 	if err != nil {
 		return -1, err // error already logged
 	}
+	if payloadSize != DefaultPayloadSize {
```
ldemailly: change this to `if payloadSize > 0 {`
it will also avoid a crash

but move it out into the caller

fhttp/http_utils.go
```diff
@@ -287,9 +287,9 @@ func generateSize(sizeInput string) (size int) {
 		}
 		size = s
 		log.Debugf("Parsed size %s -> %d", sizeInput, size)
-		if size > MaxPayloadSize {
-			log.Warnf("Requested size %d greater than max size %d, using max instead", size, MaxPayloadSize)
-			size = MaxPayloadSize
+		if size > fnet.MaxPayloadSize {
```
ldemailly: move to a function so it can also be used in fortio_main to validate/set the size

fgrpc/pingsrv.go
```diff
@@ -92,11 +94,14 @@ func PingServer(port, cert, key, healthServiceName string, maxConcurrentStreams
 
 // PingClientCall calls the ping service (presumably running as PingServer on
 // the destination). returns the average round trip in seconds.
-func PingClientCall(serverAddr, cacert string, n int, payload string, delay time.Duration) (float64, error) {
+func PingClientCall(serverAddr, cacert string, n int, payload string, payloadSize int, delay time.Duration) (float64, error) {
```
ldemailly: no, this is weird to pass 2 payloads to this function. don't change this signature (will also avoid changing a bunch of tests), instead generate from the caller

ldemailly (changes requested):
thanks, see below

ldemailly:
also let's increase/keep coverage

fhttp/http_utils.go
olimpias: I didn't understand what you mean in this comment. The size is set according to input `sizeInput`

olimpias:
I'm not sure how I can increase test coverage in network class. The only not covered places are about log.Err.

fhttp/http_utils.go
ldemailly: The if is needed 3 times (not exceed max size(

olimpias: I got it now. I implemented the method in network instead of fortio_main. I can not check the size before `generateSize` method because, the input size is not flat. String checks are applied on `InputSize`.

fnet/network.go
```diff
@@ -183,3 +213,12 @@ func NormalizeHostPort(inputPort string, addr *net.TCPAddr) string {
 	}
 	return urlHostPort
 }
+
+// ValidatePayloadSize compares input size with MaxPayLoadSize. If size exceeds the MaxPayloadSize
+// size will set to MaxPayLoadSize
+func ValidatePayloadSize(size *int) {
```
ldemailly, line 219: (optional but would be better) : could also validate size is >= 0

fortio_main.go
```diff
@@ -141,6 +141,8 @@ var (
 
 	maxStreamsFlag = flag.Uint("grpc-max-streams", 0,
 		"MaxConcurrentStreams for the grpc server. Default (0) is to leave the option unset.")
+
+	payloadSizeFlag = flag.Int("payload-size", fgrpc.DefaultPayloadSize, "Payload size to create random payload")
```
ldemailly: you didn't add "replaces -payload when set" and didn't update README.md to match output of release/updateFlags.sh

let's make the help string
"Additional random payload size, replaces -payload when set > 0, must be smaller than -maxpayloadsizekb"

note/optional: it may be more interesting for users to automatically set maxpayloadsize

fgrpc/pingsrv.go
```diff
@@ -35,6 +35,8 @@ import (
 const (
 	// DefaultHealthServiceName is the default health service name used by fortio.
 	DefaultHealthServiceName = "ping"
+	// DefaultPayloadSize indicates the message size of the grpc message for ping
+	DefaultPayloadSize = 0
```
ldemailly: let's not have a constant for 0, delete

ldemailly (commented):
almost there but:

fnet/network.go
```diff
@@ -217,7 +217,7 @@ func NormalizeHostPort(inputPort string, addr *net.TCPAddr) string {
 // ValidatePayloadSize compares input size with MaxPayLoadSize. If size exceeds the MaxPayloadSize
 // size will set to MaxPayLoadSize
 func ValidatePayloadSize(size *int) {
-	if *size > MaxPayloadSize {
+	if *size > MaxPayloadSize || *size < 0 {
```
ldemailly: probably should be a different if and basically use 0 for <0 not max

ldemailly (changes requested):
almost, just that negative -> max issue to fix

fnet/network.go
olimpias: In the `TestGenerateSize` test, 0 is expected to be 0. If I convert `*size<=0`', it will cause to fail. Also above comment, you told me to use s >= 0 Aren't we going to handle negative size and update to `MaxPayloadSize`?

ldemailly: negative payload should be no payload - not the max payload

ldemailly: and the test is wrong, -1 should yield 0 not max

fnet/network.go
```diff
@@ -217,8 +217,11 @@ func NormalizeHostPort(inputPort string, addr *net.TCPAddr) string {
 // ValidatePayloadSize compares input size with MaxPayLoadSize. If size exceeds the MaxPayloadSize
 // size will set to MaxPayLoadSize
 func ValidatePayloadSize(size *int) {
-	if *size > MaxPayloadSize || *size < 0 {
+	if *size > MaxPayloadSize && *size > 0 {
 		log.Warnf("Requested size %d greater than max size %d, using max instead", *size, MaxPayloadSize)
 		*size = MaxPayloadSize
+	} else if *size < 0 {
+		log.Warnf("Requested size %d is smaller than 0. Size is set 0", *size)
```
ldemailly: “Requested size %d is negative, using 0 (no additional payload) instead.”

ldemailly (approved):
Thanks!

olimpias:
Thank you for your comments :)

Issue #260, opened 2018-06-20
support http1 post / client side payload


ldemailly:
We can already control server side reply size (size= parameter to echo server), and for GRPC load test pass a -payload string (though that's inconvenient for large payload/limited by the command line length)

We should support POST outside of GRPC

a) generated payload
b) raw file (which could be key/value pair encoded)

olimpias:
Does h1 mean http 1.1/1.0 right?

ldemailly:
yes vs h2 used in grpc

olimpias:
May I work on this? Also shouldn't be there one more input flag for identifying content-type like json, xml.. etc right?

ldemailly:
I was thinking to give it a shot, let's see if I manage to find the time this week otherwise sure

alex-koturanov:
Looking forward to see HTTP POST support in Fortio to pass JSON data in the body of the request.

olimpias:
Hi @ldemailly, if you are still busy, I can propose a solution and write my solution way to you for your confirmation before I start to implement.

ldemailly:
Thanks, go ahead / give it a shot

olimpias:
At the beginning, I was thinking to rename `Fetcher interface` because we were going to post a request, however the point our post again to fetch response(Not interested in input data or response details). So that I think  `Fetcher interface` and fetch method could be reused. Also I'm thinking that we can reuse flag payload for http as well, but we need to add raw type(json/xml) as a flag. For the request part, if a payload is set for http request, we could use POST method also set to content-type to header. I'm assuming that content-type will be set from raw type flag. Raw type will be mandatory for http payload. I mean if a user wants to use http payload, user has to add raw-type from flags as well . I would like to hear your opinions about my comment @ldemailly

ldemailly:
I see 3 modes, 2 of which we have already for grpc: 

-payload-size random payload (though the content type is to be determined)
-payload string 
-payload-file -  for stdin or filename for a file to send as is - we do need content type option

PR #273, opened 2018-07-28
Support For Http Post

olimpias:
*Http post support is added for the client and also for the fast client.
*Payload can be upload from a file with -payload-file.
*Content type is added as a flag for http.
Ref task #260


olimpias:
Hi @ldemailly , I hope you are okay :). This is my ping time.

ldemailly:
just back from europe and my leave tonight, will check tomorrow (hopefully), ping again if I don’t !

fnet/mockfile
```diff
@@ -0,0 +1 @@
+{"test":"test"}
```
ldemailly: maybe use a new testdata/ directory it's more idiomatic and call the file payloadTest1.txt

fhttp/http_client.go
```diff
@@ -135,6 +135,10 @@ type HTTPOptions struct {
 	HTTPReqTimeOut time.Duration // timeout value for http request
 
 	UserCredentials string // user credentials for authorization
+
+	ContentType string // indicates request body type
+	Payload     string // body for http request
```
ldemailly: this should probably be `byte[]` and also test that indeed binary gets sent cleanly (and echo'ed back) something like AB\0CD (ie bytes 41 42 00 43 44  don't get truncated at the NUL; another test would be invalid utf8 sequence)

fhttp/http_client.go
```diff
@@ -404,6 +419,13 @@ func (c *FastClient) Close() int {
 // This function itself doesn't need to be super efficient as it is created at
 // the beginning and then reused many times.
 func NewFastClient(o *HTTPOptions) Fetcher {
+	var method string
+	payloadLen := len(o.Payload)
+	if payloadLen > 0 {
```
ldemailly: duplicated from above, share the code/do it once

fnet/network_test.go
```diff
@@ -255,6 +255,73 @@ func TestValidatePayloadSize(t *testing.T) {
 	}
 }
 
+func TestGenerateRandomPayload(t *testing.T) {
```
ldemailly, line 311: you can add a test for sending all the 2 bytes from 0x0000 to 0xFFFF - but not sure the current echo handler can handle that raw binary as it expects I believe url-encoded / http standard - tbd

ldemailly (commented):
very nice start!

fhttp/http_client.go
olimpias: Added `TestPayloadWithEchoBack` method.

fnet/mockfile
olimpias: I made .testdata/ as a folder name because `make test` was giving error due to go test ../... It is looking for files that includes .go . testdata is not including any .go

fhttp/http_client.go
```diff
@@ -151,6 +154,17 @@ func (h *HTTPOptions) InitHeaders() {
 	if err != nil {
 		log.Errf("User credential is not valid. %v", err)
 	}
+	if len(h.ContentType) > 0 {
+		h.extraHeaders.Add("Content-Type", h.ContentType)
+	}
+}
+
+// GetPayloadString returns the payload as a string. If paylaod is null return empty string
```
ldemailly: paylaod->payload

also add a comment this is only needed because grpc ping proto takes a string instead of bytes
(my initial mistake and we're a bit stuck for backward compatibility)

fhttp/http_test.go
```diff
@@ -492,6 +492,37 @@ func TestGenerateSize(t *testing.T) {
...
+	}
+	m, a := DynamicHTTPServer(false)
+	m.HandleFunc("/", EchoHandler)
+	url := fmt.Sprintf("http://localhost:%d/", a.Port)
+	for _, test := range tests {
+		opts := NewHTTPOptions(url)
+		opts.HTTP10 = test.http10
+		opts.Payload = test.payload
+		cli := NewClient(opts)
+		code, body, _ := cli.Fetch()
```
ldemailly: let's use the header length argument here so you can use that instead of len(body)-len(test.payload)

fhttp/http_test.go
```diff
@@ -678,6 +709,93 @@ func TestPayloadSizeSmall(t *testing.T) {
...
+		{"application/xml",
+			[]byte("<test test=\"test\">"),
+			"POST"},
+		{"",
+			nil,
+			"GET"},
+	}
+	for _, test := range tests {
+		hOptions := HTTPOptions{}
+		hOptions.URL = "http://foo.com"
```
ldemailly: it's a bit dangerous to use a real site, either use doesntexist.istio.io or use google.com / a real site that won't mind getting connections

fhttp/http_test.go
```diff
@@ -678,6 +709,93 @@ func TestPayloadSizeSmall(t *testing.T) {
...
+		hOptions.ContentType = test.contentType
+		hOptions.Payload = test.payload
+		hOptions.DisableFastClient = true
+		fetcher := NewClient(&hOptions)
```
ldemailly: you could call NewStdClient directly (and NewFastClient for the other)

but why do you need the actual client ?

olimpias: You are right, instead of working with casting to client, I can just use `NewStdClient`

olimpias: For `NewFastClient` method I change return type from `Client` to `*FastClient`

ldemailly: yes but why, can't the test be generic for both clients and just toggle the DisableFastClient bit (run it twice once with true, once with false; but same "e2e" test)

fhttp/http_client.go
```diff
@@ -159,7 +159,8 @@ func (h *HTTPOptions) InitHeaders() {
 	}
 }
 
-// GetPayloadString returns the payload as a string. If paylaod is null return empty string
+// GetPayloadString returns the payload as a string. If payload is null return empty string
+// This is only needed due to grpc ping proto. It takes string stead of byte array.
```
ldemailly: stead -> instead

fhttp/http_client.go
```diff
@@ -430,7 +431,7 @@ func (c *FastClient) Close() int {
 // NewFastClient makes a basic, efficient http 1.0/1.1 client.
 // This function itself doesn't need to be super efficient as it is created at
 // the beginning and then reused many times.
-func NewFastClient(o *HTTPOptions) Fetcher {
+func NewFastClient(o *HTTPOptions) *FastClient {
```
ldemailly: The idea there was to hide implementation details, that you get just a Fetcher either way but maybe this is ok - depending on the answer to my other question (why do you need the details)

fhttp/http_test.go
```diff
@@ -523,6 +524,27 @@ func TestPayloadWithEchoBack(t *testing.T) {
 	}
 }
 
+func extractContentLength(body string) int {
```
ldemailly: sorry I wasn't clear, I meant to use
https://github.com/istio/fortio/blob/master/fhttp/http_client.go#L38-L40
the 3rd returned parameter which is where does the body start

README.md
```diff
@@ -109,6 +109,8 @@ target is a url (http load tests) or host:port (grpc health test).  flags are:
 	Path to the certificate file to be used for GRPC server TLS
   -compression
 	Enable http compression
+  -content-type
```
ldemailly: I noticed in my other PR that the flags section of the README is actually slightly out of sync, from a white space point of view, with the source, did you run the full 

go install .
release/updateFlags.sh

and copy paste the whole section into the readme ?

fhttp/http_client.go
olimpias: In the `TestPayloadForFastClient` test, I'm looking for the `client.req` field to check If the body is applied on FastClient as I expected. So I needed to receive `*FastClient` typed object to look 'req' field.

fhttp/http_test.go
olimpias: In `TestPayloadWithEchoBack` while I was testing for 3rd parameter, I received a empty body for http 1.0 in fast client. Even I sent body it returns 0. Am I missing flag or is there a bug? Btw I was accually looking for testing fastClient or stdclient

README.md
olimpias: I did

fhttp/http_test.go
olimpias: The problem that I mention above, I could take a look after this PR

fhttp/http_client.go
ldemailly: you could just check it is echoed back?

this being said we can make that signature change as the other std client already returns the concrete type

ldemailly:
can you rebase/resolve conflicts ?

fhttp/http_client.go
olimpias: Okay, I take `Fetcher` back. But at `TestPayloadForFastClient` test, my point is to see do I send body to server as I expected. I mean that I'm checking the headers like `content type`,`content length` and `method`. If I use echo back for `TestPayloadForFastClient` I won't able to cover the method type.

olimpias: Btw if you are saying that instead of `TestPayloadForClient` and `TestPayloadForFastClient` tests, we could use e2e, its okay for me as well. But in the above comment, I want to mention my concerns.

fhttp/http_test.go
ldemailly: do you have something reproducible, I'm not sure what you mean exactly

README.md
```diff
@@ -81,34 +81,36 @@ Full list of command line flags (`fortio help`):
...
-where command is one of: load (load testing), server (starts grpc ping and http
-echo/ui/redirect/proxy servers), grpcping (grpc client), report (report only UI
-server), redirect (redirect only server), or curl (single URL debug).  where
-target is a url (http load tests) or host:port (grpc health test).  flags are:
-  -H value
+where command is one of: load (load testing), server (starts grpc ping and
+http echo/ui/redirect/proxy servers), grpcping (grpc client), report (report
+only UI server), redirect (redirect only server), or curl (single URL debug).
+where target is a url (http load tests) or host:port (grpc health test).
+flags are: .H value
```
ldemailly, line 88: interesting, this is wrong but looks like the script works only on my mac and not in a docker image for instance

it's because mac is bsd fmt:
https://www.freebsd.org/cgi/man.cgi?query=fmt
which doesn't format lines starting with a .
vs
http://linuxcommand.org/lc3_man_pages/fmt1.html

fhttp/http_test.go
olimpias: Try this test in http_test in my branch. You will see that `header` will return 0 from `cli.Fetch()`, but it is needed to be 5. This error happens when http1.0 is used for fastclient
```
func TestFetchError(t *testing.T) {
	var tests = []struct {
		payload           []byte
		http10 bool
	}{
		{[]byte{44, 45, 00, 46, 47}, true},
		{[]byte("groß"), true},
	}
	m, a := DynamicHTTPServer(false)
	m.HandleFunc("/", EchoHandler)
	url := fmt.Sprintf("http://localhost:%d/", a.Port)
	for _, test := range tests {
		opts := NewHTTPOptions(url)
		opts.HTTP10 = test.http10
		opts.Payload = test.payload
		cli := NewClient(opts)
		code, body, header := cli.Fetch()
		if code != 200 {
			t.Errorf("Unexpected error %d", code)
		}
		fmt.Printf("Header size %d\n",header)
		if !bytes.Equal(body[header:], test.payload) {
			t.Errorf("Got %s, expected %s from echo", string(body), string(test.payload))
		}
		cli.Close()
	}
}
```

fhttp/http_client.go
olimpias: Any comment for here :)

fhttp/http_test.go
ldemailly: oh http1.0 on purpose doesn't bother parsing headers (to be faster) just don't use that to check POST

olimpias: Ah okay. I released that while I was wrongly using http10 instead of disablefastclient for e2e.

ldemailly:
merge conflict solved in #280 and code review / follow up change in https://github.com/istio/fortio/pull/280/commits/a3689789917fae7766cd38818db0d9e403471c8b
(please check those changes)

PR #276, opened 2018-08-07
Add unix domain support

### LINKED ISSUE #275: UDS (unix domain socket) support
Will help when testing envoy's UDS support

cc @costinm

ldemailly:
Fixes #275

Allows -unix-socket option for client side and use a path for server 'ports',
eg. 
`fortio server -http-port /tmp/fortio/http-uds1`
and
`fortio load -unix-socket /tmp/fortio/http-uds1 http://localhost/echo`

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,8 +229,9 @@ Fortio 1.1.0 grpc 'ping' server listening on [::]:8079
 Fortio 1.1.0 https redirector server listening on [::]:8081
 Fortio 1.1.0 echo server listening on [::]:8080
 UI started - visit:
-http://localhost:8080/fortio/   (or any host/ip reachable on this server)
-21:45:23 I fortio_main.go:195> All fortio 1.1.0 buildinfo go1.10 servers started!
+http://localhost:8080/fortio/
+(or any host/ip reachable on this server)
+14:57:12 I fortio_main.go:217> All fortio 1.1.0 unknown go1.10.3 servers started!
 ```
 
 * By default, Fortio's web/echo servers listen on port 8080 on all interfaces.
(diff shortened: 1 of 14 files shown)


ldemailly:
I'll improve coverage tomorrow but please start looking at the diff

fnet/network.go
```diff
@@ -111,7 +135,8 @@ func ResolveDestination(dest string) *net.TCPAddr {
 
 // Resolve returns the TCP address of the host,port suitable for net.Dial.
 // nil in case of errors.
-func Resolve(host string, port string) *net.TCPAddr {
+func Resolve(host string, port string) net.Addr {
```
olimpias, line 139: Do we really need to change here to `net.Addr` because if I'm not thinking wrong, we are using `resolve` method for TCP ADDR. It might return again '*net.TCPAddr'

ldemailly: You are right, I made that change when I was going down the path of trying to hack into the host/port part of the url some kind of scheme to parse UDS. golang's url parse rejects encoding inside the host/part so I gave up and went for UnixDomainSocket option instead

This being said it doesn't hurt to have the option to return other type of addresses and most of the code now does take the generic Addr instead of TCPAddr

README.md
```diff
@@ -127,7 +127,8 @@ target is a url (http load tests) or host:port (grpc health test).  flags are:
 	grpc ping delay in response
   -grpc-port string
 	grpc server port. Can be in the form of host:port, ip:port or port or
-	"disabled" to not start the grpc server. (default "8079")
+	/unix/domain/path or "disabled" to not start the grpc server. (default
```
costinm, line 130: Do you also support namespaced UDS ? (no path - just name ) I believe that's what envoy uses.

costinm (approved):

PR #280, opened 2018-08-10
http post feature (merge/update of #273)

### LINKED ISSUE #260: support http1 post / client side payload

### LINKED ISSUE #282: headers cleanup. -u doesn't work with -H and other similar issue
The initialization of headers is overly complicated and has created bugs in recent diff
#261 for instance adds -user but it doesn't work if -H is also provided:
```
ldemailly-macbookpro:fortio ldemailly$ fortio curl -user foo:bar  http://localhost:8080/debug
HTTP/1.1 200 OK
Content-Type: text/plain; charset=UTF-8
Date: Fri, 10 Aug 2018 22:14:26 GMT
Content-Length: 284

Φορτίο version 1.1.0-pre unknown go1.10.3 echo debug server up for 59m4.5s on ldemailly-macbookpro.roam.corp.google.com - request from [::1]:56408

GET /debug HTTP/1.1

headers:

Host: localhost:8080
Authorization: Basic Zm9vOmJhcg==
User-Agent: istio/fortio-1.1.0-pre

body:


ldemailly-macbookpro:fortio ldemailly$ fortio curl -user foo:bar -H blah:blah  http://localhost:8080/debug
HTTP/1.1 200 OK
Content-Type: text/plain; charset=UTF-8
Date: Fri, 10 Aug 2018 22:14:45 GMT
Content-Length: 262

Φορτίο version 1.1.0-pre unknown go1.10.3 echo debug server up for 59m23.7s on ldemailly-macbookpro.roam.corp.google.com - request from [::1]:56413

GET /debug HTTP/1.1

headers:

Host: localhost:8080
Blah: blah
User-Agent: istio/fortio-1.1.0-pre

body:


```

found while working on/testing #280 - cc @olimpias

ldemailly:
@olimpias's #273 (fixes #260) + merge + review changes

additionally, fixes #282


olimpias (approved):
Looking very good to me :). I like to way of  `GenerateHeaders` method

PR #286, opened 2018-08-21
Use https instead of http

ldemailly:
Avoids security error on https://istio.io/


ldemailly:
#287 will cleanup to copy pasta/duplication in those 3 files

olimpias (approved):

Issue #161, opened 2018-02-27
json result for grpc should display RetCode enum String rather than its ordinal value


asadali:
@ldemailly pointed out:

> so the code printing stdout is
> `fmt.Printf("Health %s : %d\n", k.String(), total.RetCodes[k])`
> for some reason the Json has the numerical value "1" instead of "SERVING" - I wonder if that's fixable easily with some annotation on the struct

current json output:
```
  "RetCodes": {
    "1": 3000
  }
```

expected json output:
```
  "RetCodes": {
    "SERVING": 3000
  }
```

PR #290, opened 2018-08-23
json result for grpc should display RetCode enum String rather than its ordinal value #161

olimpias:
*To reach string type instead of ordinary value.

*Solves #161


ldemailly (approved):
thanks
there is maybe a bit too many .String() but it's fine!

PR #292, opened 2018-08-27
fixing duplicate headers, because generate is called more than once, …

### LINKED ISSUE #291: Duplication of Content-Type header
**Description**
There are duplications of `Content-Type` header in the requests that are generated by Fortio. The number of duplicated headers is corresponding to the defined number of connections.

**How to reproduce**
1. Run `fortio load -content-type 'application/json' -payload '{"query":"{__schema { types { kind } }}"}' -c 3 -qps 0 -n 1 -loglevel Debug https://31zrkwkkv.lp.gql.zone/graphql`
2. In the debug output will be three requests with different set of headers:
```
Content-Type: application/json
```
```
Content-Type: application/json
Content-Type: application/json
```
```
Content-Type: application/json
Content-Type: application/json
Content-Type: application/json
```

**What expected**
Only one `Content-Type: application/json` header per request

ldemailly:
…use Set instead of Add. Fixes #291. (would need a test and maybe better refactoring)

repro is
```
fortio load -user foo:bar -content-type foo/bar -loglevel debug -c 2 -n 2 www.google.com
```

before:
```
POST / HTTP/1.1
Host: www.google.com
Authorization: Basic Zm9vOmJhcg==
Authorization: Basic Zm9vOmJhcg==
Content-Length: 0
Content-Type: foo/bar
Content-Type: foo/bar
User-Agent: fortio.org/fortio-1.2.0-pre
```
on the 2nd connection,
after
```
POST / HTTP/1.1
Host: www.google.com
Authorization: Basic Zm9vOmJhcg==
Content-Length: 0
Content-Type: foo/bar
User-Agent: fortio.org/fortio-1.2.0-pre
```

### DIFF
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -87,6 +87,7 @@ const (
 // GenerateHeaders completes the header generation, including Content-Type/Length
 // and user credential coming from the http options in addition to extra headers
 // coming from flags and AddAndValidateExtraHeader().
+// Warning this gets called more than once, do not generate duplicate headers.
 func (h *HTTPOptions) GenerateHeaders() http.Header {
 	if h.extraHeaders == nil { // not already initialized from flags.
 		h.InitHeaders()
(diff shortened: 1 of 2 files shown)


fhttp/http_client.go
```diff
@@ -87,6 +87,7 @@ const (
 // GenerateHeaders completes the header generation, including Content-Type/Length
 // and user credential coming from the http options in addition to extra headers
 // coming from flags and AddAndValidateExtraHeader().
+// Warning this gets called more than once, do not generate duplicate headers.
```
ldemailly, line 90: TODO: this is a bit of a landmine, the header setting from flag vs other httpoptions is a mess that needs refactoring but this PR fixes the known bugs for now - ptal

mandarjog (approved):
lgtm

PR #297, opened 2018-08-28
switch to v11 build with go1.11 (keep 1.10.3 in build image for now t…

ldemailly:
…oo), also future proof a bit so no need change so many lines when go1.12 comes out


ldemailly:
this was done using https://github.com/fortio/fortio/tree/master/release#how-to-change-the-build-image
and it seems go1.11 is faster (on mac+docker) than go1.10

ldemailly:
looks like linters have changed in 1.11 ... will need some update

Issue #312, opened 2018-11-30
Help Wanted: GRPC results always say "100% error"


ihcsim:
The legend in my grpc result charts always seem to show a _100% error_ label. There are no errors in the report log. Is the label correct?

<img width="1403" alt="screen shot 2018-11-30 at 11 54 54 am" src="https://user-images.githubusercontent.com/1330522/49311963-6ebe3b80-f497-11e8-9ee9-a8d088e4bb36.png">

Here's the corresponding report log.

```
Fortio 1.3.1-pre running at 2113 queries per second, 1->1 procs, for 5m0s: fortio-echo-1.benchmark-istio:8079
13:46:51 I grpcrunner.go:152> Starting GRPC Ping test for fortio-echo-1.benchmark-istio:8079 with 1*32 threads at 2113.0 qps
Starting at 2113 qps with 32 thread(s) [gomax 1] for 5m0s : 19809 calls each (total 633888)
13:51:51 I periodic.go:533> T025 ended after 4m59.9970303s : 19809 calls. qps=66.03065363744035
13:51:51 I periodic.go:533> T005 ended after 5m0.006889453s : 19809 calls. qps=66.02848366621707
13:51:51 I periodic.go:533> T000 ended after 5m0.009846806s : 19809 calls. qps=66.02783278913309
13:51:51 I periodic.go:533> T016 ended after 5m0.010833158s : 19809 calls. qps=66.02761570802224
13:51:51 I periodic.go:533> T006 ended after 5m0.010860616s : 19809 calls. qps=66.02760966495343
13:51:51 I periodic.go:533> T021 ended after 5m0.010896479s : 19809 calls. qps=66.02760177207956
13:51:51 I periodic.go:533> T003 ended after 5m0.010909818s : 19809 calls. qps=66.02759883637906
13:51:51 I periodic.go:533> T011 ended after 5m0.010922416s : 19809 calls. qps=66.02759606376104
13:51:51 I periodic.go:533> T018 ended after 5m0.010934967s : 19809 calls. qps=66.02759330148719
13:51:51 I periodic.go:533> T015 ended after 5m0.010946929s : 19809 calls. qps=66.02759066884302
13:51:51 I periodic.go:533> T027 ended after 5m0.010983224s : 19809 calls. qps=66.02758268089747
13:51:51 I periodic.go:533> T019 ended after 5m0.010994844s : 19809 calls. qps=66.02758012352282
[...]

PR #313, opened 2018-12-01
Error display for grpc is fixed

olimpias:
In #290 task, grpc status display values have changed, but some side effects are missed for UI.

Ref:#312


ldemailly (approved):
thanks

PR #316, opened 2019-01-24
Use proxy from env if stdclient is used

costinm:


ldemailly (commented):
? what is this for/about ?

ldemailly:
will close this as it seeme erroneous but please explain if there is an issue

PR #317, opened 2019-01-31
Use proxy from environment

costinm:
Sorry for the previous PR, got mixed into vendor problems...

I'm using the 'std' http client, and want to test using HTTP_PROXY environment variable - istio has supported this for a while, but now we have a new 'Whitebox' mode, without any iptables - and outgoing
http traffic will all go trough HTTP_PROXY.  (TCP will go to localhost:port) 

The line just tells go to read the environment variable if set and use it when connecting.

### DIFF
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -374,6 +374,7 @@ func NewStdClient(o *HTTPOptions) *Client {
 		MaxIdleConnsPerHost: o.NumConnections,
 		DisableCompression:  !o.Compression,
 		DisableKeepAlives:   o.DisableKeepAlive,
+		Proxy:               http.ProxyFromEnvironment,
 		Dial: (&net.Dialer{
 			Timeout: o.HTTPReqTimeOut,
 		}).Dial,


ldemailly (approved):
Thanks, This uses/matches https://golang.org/pkg/net/http/#ProxyFromEnvironment and DefaultTransport

It would be nice to have a test and support in the fast client too but that can be later

costinm:
Thanks. It can be added to the fast client - but a bit more complicated and would really need a good test for fast client. 
Not sure if there is any library to start a quick http proxy server - can be tested manually or e2e easily ( I have a test in in istio using this, with a custom build of fortio with the PR merged )

ldemailly:
I am traveling and my travel laptop crashed and I have to reinstall linux subsystem for windows10, ubuntu, go, git, vscode etc... on it so it's taking me a bit more time but I'll make a 1.3.1 soonish (I want to get in the debian PR too - btw if you can help turning that debian/ setup into an actual accepted package that would be nice)

PR #328, opened 2019-02-18
Refactor generateStatus, generateSize and generateDelay

### LINKED ISSUE #145: factor out/refactor fhttp generateSize generateDelay generateStatus
they are a lot of common code (but also differences)

thekuwayama:
Resolves #145 

This PR adds function to extract common code of generateStatus, generateSize and generateDelay.

### DIFF
--- a/fhttp/http_utils.go
+++ b/fhttp/http_utils.go
@@ -208,7 +208,30 @@ func removeTrailingPercent(s string) string {
 	return s
 }
 
-func parseFromFormattedString(input, about string, parseFunc func(string) (int64, error)) (int64, error) {
+func parseEntry(entry, about, input string, parseFunc func(input string) (int64, error)) (int64, float64, error) {
+	l2 := strings.Split(entry, ":")
+	if len(l2) != 2 {
+		log.Warnf("Should have exactly 1 : in %s list %s -> %v", about, input, entry)
+		return -1, -1, errors.New("invalid format")
+	}
+	s, err := parseFunc(l2[0])
+	if err != nil {
+		log.Warnf("Bad input %s %v -> %v, not a number before colon", input, about, l2[0])
+		return -1, -1, errors.New("invalid format")
+	}
+	percStr := removeTrailingPercent(l2[1])
+	p, err := strconv.ParseFloat(percStr, 32)
+	if err != nil || p < 0 || p > 100 {
+		log.Warnf("Percentage is not a [0. - 100.] number in %v -> %v : %v %f", input, percStr, err, p)
+		return -1, -1, errors.New("invalid format")
+	}
+	return s, p, nil
+}
+
+// input="XX:20,YY:10,ZZ:0.5" for 20% parsed_XX, 10% parsed_YY, 0.5% parsed_ZZ 69.5% -1(means default value)
+// entry parsed by parseFunc
+// when parseFromFormattedString raise error, return -1, error(not nil)
+func parseFormattedString(input, about string, parseFunc func(string) (int64, error)) (int64, error) {
 	// Input is empty case:
 	if len(input) == 0 {
 		return -1, errors.New("input is empty")
@@ -219,30 +242,19 @@ func parseFromFormattedString(input, about string, parseFunc func(string) (int64
 	if len(lst) == 1 && !strings.ContainsRune(input, ':') {
 		s, err := parseFunc(input)
 		if err != nil {
-			return -1, err
+			s = -1
 		}
-		return s, nil
+		return s, err
 	}
+	// Parse each entry
 	weights := make([]float32, len(lst))
 	values := make([]int64, len(lst))
 	lastPercent := float64(0)
 	i := 0
 	for _, entry := range lst {
-		l2 := strings.Split(entry, ":")
-		if len(l2) != 2 {
-			log.Warnf("Should have exactly 1 : in %s list %s -> %v", about, input, entry)
-			return -1, errors.New("invalid format")
-		}
-		s, err := parseFunc(l2[0])
+		s, p, err := parseEntry(entry, about, input, parseFunc)
 		if err != nil {
-			log.Warnf("Bad input %s %v -> %v, not a number before colon", input, about, l2[0])
-			return -1, errors.New("invalid format")
-		}
-		percStr := removeTrailingPercent(l2[1])
-		p, err := strconv.ParseFloat(percStr, 32)
-		if err != nil || p < 0 || p > 100 {
-			log.Warnf("Percentage is not a [0. - 100.] number in %v -> %v : %v %f", input, percStr, err, p)
-			return -1, errors.New("invalid format")
+			return -1, err
 		}
 		lastPercent += p
 		// Round() needed to cover 'exactly' 100% and not more or less because of rounding errors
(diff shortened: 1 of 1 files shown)


fhttp/http_utils.go
```diff
@@ -207,131 +208,106 @@ func removeTrailingPercent(s string) string {
 	return s
 }
 
-// generateStatus from string, format: status="503" for 100% 503s
-// status="503:20,404:10,403:0.5" for 20% 503s, 10% 404s, 0.5% 403s 69.5% 200s
-func generateStatus(status string) int {
-	lst := strings.Split(status, ",")
-	log.Debugf("Parsing status %s -> %v", status, lst)
+func parseFromFormattedString(input, about string, parseFunc func(string) (int64, error)) (int64, error) {
```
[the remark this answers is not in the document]
thekuwayama: 👍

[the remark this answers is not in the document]
thekuwayama: resolved 👍

fhttp/http_utils.go
```diff
@@ -207,131 +208,118 @@ func removeTrailingPercent(s string) string {
...
+		log.Warnf("Percentage is not a [0. - 100.] number in %v -> %v : %v %f", input, percStr, err, p)
+		return -1, -1, errors.New("invalid format")
+	}
+	return s, p, nil
+}
+
+// input="XX:20,YY:10,ZZ:0.5" for 20% parsed_XX, 10% parsed_YY, 0.5% parsed_ZZ 69.5% -1(means default value)
+// entry parsed by parseFunc
+// when parseFromFormattedString raise error, return -1, error(not nil)
+func parseFormattedString(input, about string, parseFunc func(string) (int64, error)) (int64, error) {
```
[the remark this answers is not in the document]
thekuwayama: resolved 👍

fhttp/http_utils.go
```diff
@@ -207,131 +208,132 @@ func removeTrailingPercent(s string) string {
...
```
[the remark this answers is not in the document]
ldemailly: ignore this, it's fine

fhttp/http_utils.go
```diff
@@ -232,48 +232,60 @@ func parseEntry(entry, about, input string, parseFunc func(input string) (int64,
...
-		if err != nil {
-			return -1, err
+	weights = make([]float32, len(lst))
+	values = make([]int64, len(lst))
+	lastPercent = float64(0)
+	for i, entry := range lst {
+		s, p, e := parseEntry(entry, about, input, parseFunc)
+		if e != nil {
+			err = e
+			goto ParseError
```
ldemailly: I'm not a big fan of goto but maybe in this case it's ok (usually return/break is enough- sometime adding an intermediate function helps too)

fhttp/http_utils.go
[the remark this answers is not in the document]
thekuwayama: resolved 👍

fhttp/http_utils.go
thekuwayama: @ldemailly Thanks for the feedback !!
I rewrote.
Please let me know if you have any concerns.

fhttp/http_utils.go
```diff
@@ -207,131 +208,127 @@ func removeTrailingPercent(s string) string {
...
-			log.Warnf("Should have exactly 1 : in size list %s -> %v", sizeInput, entry)
-			return size
-		}
-		s, err := strconv.Atoi(l2[0])
+	parseSizeFunc := func(input string) (int64, error) {
+		s, err := strconv.Atoi(input)
 		if err != nil {
-			log.Warnf("Bad input size %v -> %v, not a number before colon", sizeInput, l2[0])
-			return size
+			log.Warnf("Bad input status %v -> %v, not a number before colon", input, s)
```
ldemailly: should be size, not status here

fhttp/http_utils.go
```diff
@@ -207,131 +208,127 @@ func removeTrailingPercent(s string) string {
...
+		return -1, -1, errors.New("invalid format")
+	}
+	return s, p, nil
+}
+
+func getWeightedValue(input string, weights []float32, values []int64) int64 {
+	res := 100. * rand.Float32()
+	for i, v := range weights {
+		if res <= v {
+			if i >= len(values) {
```
ldemailly: this test/if is new, why / when does this case happen ? (comment needed at least)
you could refactor to have one vector of pairs / struct instead of 2 arrays - or assert the 2 arrays have the same size instead of silently swallowing what must be a bug of usage)

fhttp/http_utils.go
```diff
@@ -207,131 +208,127 @@ func removeTrailingPercent(s string) string {
...
+	}
+	lst := strings.Split(input, ",")
+	log.Debugf("Parsing %s %s -> %v", about, input, lst)
 	// Simple non probabilistic status case:
-	if len(lst) == 1 && !strings.ContainsRune(status, ':') {
-		s, err := strconv.Atoi(status)
+	if len(lst) == 1 && !strings.ContainsRune(input, ':') {
+		s, err := parseFunc(input)
 		if err != nil {
-			log.Warnf("Bad input status %v, not a number nor comma and colon separated %% list", status)
```
ldemailly, line 219: that warning is gone/lost ? (the case about being just a value that doesn't parse)

fhttp/http_utils.go
```diff
@@ -207,131 +208,127 @@ func removeTrailingPercent(s string) string {
...
-			return http.StatusBadRequest
+			s = -1
```
ldemailly: in this generic function you shouldn't need both an error return and a special value ?

fhttp/http_utils.go
```diff
@@ -207,131 +208,127 @@ func removeTrailingPercent(s string) string {
...
+		if res <= v {
+			if i >= len(values) {
+				return -1
+			}
+			log.Debugf("[0. - 100.] for %s roll %f got #%d -> %d", input, res, i, values[i])
+			return values[i]
+		}
+	}
+	log.Debugf("[0. - 100.] for %s roll %f no hit, return default value", input, res)
+	return -1 // default/reminder of probability table
```
ldemailly: this could be used for types other than int (the http status code being an example even if it maps fine to an int), and for cases where -1 is a valid value maybe nil would be a better return for this case
ps: for status the default is ok not -1 and for delay it was 0 but I guess if you document (the function `getWeightedValue` is missing comments/doc even for internal it's useful) clearly maybe it's ok as is (but moves the testing for -1 in the callers)

ldemailly (commented):
thanks for this work

it's interesting we only save 1/3rd of the 3 copies (-149 lines, +103 lines)

see below

ldemailly:
friendly ping :)

thekuwayama:
Thank you for ping.
Just a second.

fhttp/http_utils.go
thekuwayama: Thank you!
Done.

fhttp/http_utils.go
thekuwayama: added WeightValuePair struct instead of 2 arrays.

fhttp/http_utils.go
thekuwayama: sorry, my mistake. add warning.

thekuwayama:
Thank you for waiting.

I've tried your suggest, modified `getWeightedValue` to returns string.
`parseFormattedString` should check entry that is convertible before got value from weighted values, so I added `checkConvertible`.

thekuwayama:
@ldemailly Sorry reply late.


I think, `parseFromFormattedString` was required to return string if it returned generic value.
`generateStatus` calls strconv.Atoi twice if `parseFromFormattedString` returns string value.
I worry about it.  🤔

thekuwayama:
@ldemailly  Thank you very much for reviewing.
It would be difficult for me to refactor so I close this PR.

PR #334, opened 2019-04-23
Support `curl --resolve` behavior. 

p4ali:
The provided ip will be used to call the target host, and url host will be used for sni answer.


ldemailly:
Thanks for the PR, I will look shortly but a preliminary question though:
Isn't the same feature achieved (at least for fast client, I have to verify for std go and ssl) by putting the IP in the url and passing the Host header with whichever you want, ie if you want to go to "google.com/foo/bar" but have google.com resolve to ip 1.2.3.4:

```
$ fortio curl -loglevel debug -H Host:google.com http://1.2.3.4/foo/bar 
09:10:26 I logger.go:97> Log level is now 0 Debug (was 2 Info)
09:10:26 V http_client.go:250> Will be setting special Host header to google.com
09:10:26 V stats.go:516> Will use [50 75 90 99 99.9] for percentiles
09:10:26 V http_client.go:123> URLSchemeCheck &{URL:http://1.2.3.4/foo/bar NumConnections:1 Compression:false DisableFastClient:false HTTP10:false DisableKeepAlive:false AllowHalfClose:false Insecure:false FollowRedirects:false initDone:true https:false extraHeaders:map[User-Agent:[fortio.org/fortio-1.3.0]] hostOverride:google.com HTTPReqTimeOut:15s UserCredentials: ContentType: Payload:[] UnixDomainSocket:}
09:10:26 V http_client.go:123> URLSchemeCheck &{URL:http://1.2.3.4/foo/bar NumConnections:1 Compression:false DisableFastClient:false HTTP10:false DisableKeepAlive:false AllowHalfClose:false Insecure:false FollowRedirects:false initDone:true https:false extraHeaders:map[User-Agent:[fortio.org/fortio-1.3.0]] hostOverride:google.com HTTPReqTimeOut:15s UserCredentials: ContentType: Payload:[] UnixDomainSocket:}
09:10:26 V http_client.go:485> No port specified, using http
09:10:26 D network.go:145> Resolve() called with host=1.2.3.4 port=http
09:10:26 D network.go:154> Host already an IP, will go to 1.2.3.4
09:10:26 D http_client.go:527> Created client:
1.2.3.4:80
GET /foo/bar HTTP/1.1
Host: google.com
User-Agent: fortio.org/fortio-1.3.0

^C
```

p4ali:
@ldemailly Thanks for the quick feedback. Is this feature available for load testing also?

 I had a test case like this:

- services A, B, C running on cluster X, behind LB (ip: 1.2.3.4)
- clients a, b, c running on cluster Y, which calling A, B, C separately with TLS through X's LB
- a vm to provision X and Y, and then trigger the clients to start load test
- All services and clients are `fortio server` 

To trigger test from **a->A**, I make http call from **vm** to **a** with following url:

`http://fortio-1:8080/fortio/?labels=2&url=https://test-1.example.com:443/echo/?size=10240&qps=-1&runner=http&https-insecure=on&save=on&load=Start&json=on"`

This will trigger load testing from **a** to **A**.

But the above failed because I have no way to pass the LB ip (1.2.3.4), ideally, it would be perfect if the the sni-name can be resolved to LB's IP - that's what this PR for.

I've thought the curl command you mentioned is just one time, is that also for load testing?

Or if you have other suggestion (I am new to fortio, but I like it very much), please let me know. Thanks!

ldemailly:
it will work for loadtesting too, put your service name in &H=Host:service and the IP in the url ?

ldemailly:
also you need to go fmt your patch (set your editor to auto does that on save is best)

.looper.yml
```diff
@@ -0,0 +1,30 @@
...
+  default:
+    - call: buildOneDockerImage
+
+  # Build docker image and publish to docker hub
+  buildOneDockerImage:
+  - node(docker-daemon):
+    - dockerImage:
+        baseDir: './'
+        tags:
+          - docker.prod.walmart.com/ali/fortio:1.3.2-sni
```
ldemailly: What is this (what is this file btw? curious why you need it ?)

Please delete it makes the build fail

ldemailly (changes requested):

Dockerfile
```diff
@@ -1,18 +1,18 @@
 # Build the binaries in larger image
-FROM docker.io/fortio/fortio.build:v13 as build
+FROM hub.docker.prod.walmart.com/fortio/fortio.build:v13 as build
```
ldemailly: please don't make those changes

I am curious though, which change did you make to the build image ?

p4ali: There's no change. But in walmart we only have internal docker proxy. I built a docker image for my change.
Sorry for not cleaning up this.

p4ali:
> it will work for loadtesting too, put your service name in &H=Host:service and the IP in the url ?

I've tried your suggest, but can not work. Attached the snapshot: One with Host header, but not work. Another use the changes, and it works.

The test-1.example.com is resolved to 1.2.3.4 (LB). And it is cert's cn. 

Screen Shot 2019-04-23 at 4 24 12 PM
Screen Shot 2019-04-23 at 4 12 43 PM

ldemailly:
Headers when using the web form go to the "Headers:" section of the UI, please try that
(they then get turned into &H= by the form)

p4ali:
Tried with the Host header in the Headers form, but still not work.

Screen Shot 2019-04-23 at 5 03 40 PM

ldemailly:
probably that’s because of https and standard go client - I am not sure if you are able to test with http ?

p4ali:
Yes, the http works. But https not. PR will make https work.

Screen Shot 2019-04-23 at 9 39 46 PM

ldemailly:
I agree the dialer is the oroper way to solve this for std go client

and I suppose for the fast client, having 2 ways to achieve the same thing is ok-ish too

thx for doing the tests!

ldemailly:
it'd be great to add a unit test to get coverage of the new feature

p4ali:
@ldemailly Sorry reply late. I am not available this week. Will do next week.

fhttp/http_client.go
```diff
@@ -155,14 +156,15 @@ const (
...
-	Insecure          bool // do not verify certs for https
-	FollowRedirects   bool // For the Std Client only: follow redirects.
+	NumConnections    int    // num connections (for std client)
+	Compression       bool   // defaults to no compression, only used by std client
+	DisableFastClient bool   // defaults to fast client
+	HTTP10            bool   // defaults to http1.1
+	DisableKeepAlive  bool   // so default is keep alive
+	AllowHalfClose    bool   // if not keepalive, whether to half close after request
+	Insecure          bool   // do not verify certs for https
+	ResovledIP        string // resolve Common Name to this ip, when use CN as target url
```
ldemailly: you can minimise the diff by moving the new variable after initDone

ui/templates/main.html
```diff
@@ -58,7 +58,8 @@ <h1>Φορτίο (fortio) v{{.Version}}{{if not .DoLoad}} control UI{{end}}</h1>
     Load using:<br />
     http: <input type="radio" name="runner" value="http" checked/>
     (https insecure:<input type="checkbox" name="https-insecure" />,
-    standard go client instead of fastclient:<input type="checkbox" name="stdclient"/>)
+    standard go client instead of fastclient:<input type="checkbox" name="stdclient"/>,
+    resolvedIP: <input type="text" name="resolvedIP" size="12" value="" />)
```
ldemailly: call it resolve, I try so far to have the url param name and the flag name be the same so it’s self documented

ui/uihandler.go
```diff
@@ -158,6 +158,7 @@ func Handler(w http.ResponseWriter, r *http.Request) {
 	grpcPingDelay, _ := time.ParseDuration(r.FormValue("grpc-ping-delay"))
 	stdClient := (r.FormValue("stdclient") == "on")
 	httpsInsecure := (r.FormValue("https-insecure") == "on")
+	resolvedIP := r.FormValue("resolvedIP")
```
ldemailly: should be « resolve »

p4ali: @ldemailly Thank you so much for being patient with me, I am relative new to go, and feel honored to work gurus like you. 

I guess you will smash all the commit before merge.  Otherwise, I will do a rebase.

ldemailly:
yes we always squish commits, no need for you to change anything and thanks for the kind words and the PR

ldemailly:
the test doesn't cover resolve for fastclient/non https but let's get this is in anyway for now (a follow up PR is welcome)

PR #341, opened 2019-06-06
Changing size of docker image

truongnh1992:
The size of docker image is 3MB instead of 3Mb

Signed-off-by: Nguyen Hai Truong <truongnh@vn.fujitsu.com>


ldemailly:
well a size “Mb” is obviously mega bytes and not mega bits but if you
really think you want to change it, make it

“3 Mbytes”

btw, technically the correct SI name is MiB but nobody really uses that

ldemailly:
going to close this, reopen if you fix it

PR #345, opened 2019-07-15
Added Merging Functionality

saimsalman36:
Addresses Issues (https://github.com/fortio/fortio/issues/99, https://github.com/fortio/fortio/issues/134). Adds the functionality to merge multiple fortio json outputs. 

Example: `fortio merge -json "MergedResults.json" -files Result1.json,Result2.json,Result3.json`

1. To enable the merging functionality I had to include the `Offset` and `Divider` variables in the json output to recover the original Histogram from the result.
2. To calculate the Duration I add up all the Durations in the individual files. It might be better to take the average.

### DIFF
--- a/fhttp/httprunner.go
+++ b/fhttp/httprunner.go
@@ -49,6 +49,51 @@ type HTTPRunnerResults struct {
 	aborter *periodic.Aborter
 }
 
+// Merge two different HTTPRunnerResults
+func Merge(rr1 *HTTPRunnerResults,
+	       rr2 *HTTPRunnerResults,
+	       percList []float64) (*HTTPRunnerResults, error) {
+
+	ret := HTTPRunnerResults{}
+
+	if rr1.URL != rr2.URL {
+		return nil, fmt.Errorf("error: %s and %s are different URLs", rr1.URL, rr2.URL)
+	}
+
+	ret.RunnerResults = *periodic.Merge(rr1.Result(),
+                                        rr2.Result(),
+                                        percList)
+
+	for k, v := range rr2.RetCodes {
+    	_, ok := rr1.RetCodes[k]
+
+    	if ok {
+    		rr1.RetCodes[k] += rr2.RetCodes[k]
+    	} else {
+    		rr1.RetCodes[k] = v
+    	}
+	}
+
+	ret.RetCodes = rr1.RetCodes
+
+	if rr1.sizes != nil && rr2.sizes != nil {
+		ret.sizes = stats.Merge(rr1.sizes, rr2.sizes)
+	}
+
+	if rr1.headerSizes != nil && rr2.headerSizes != nil {
+		ret.headerSizes = stats.Merge(rr1.headerSizes, rr2.headerSizes)
+	}
+
+	ret.Sizes = stats.MergeHistData(rr1.Sizes, rr2.Sizes, percList)
+	ret.HeaderSizes = stats.MergeHistData(rr1.HeaderSizes, rr2.HeaderSizes, percList)
+
+	ret.URL = rr1.URL
+	ret.SocketCount = rr1.SocketCount + rr2.SocketCount
+
+	return &ret, nil
+
+}
+
 // Run tests http request fetching. Main call being run at the target QPS.
 // To be set as the Function in RunnerOptions.
 func (httpstate *HTTPRunnerResults) Run(t int) {
--- a/fhttp/httprunner.go
+++ b/fhttp/httprunner.go
@@ -51,8 +51,8 @@ type HTTPRunnerResults struct {
 
 // Merge two different HTTPRunnerResults
 func Merge(rr1 *HTTPRunnerResults,
-	       rr2 *HTTPRunnerResults,
-	       percList []float64) (*HTTPRunnerResults, error) {
+	rr2 *HTTPRunnerResults,
+	percList []float64) (*HTTPRunnerResults, error) {
 
 	ret := HTTPRunnerResults{}
 
@@ -61,17 +61,17 @@ func Merge(rr1 *HTTPRunnerResults,
 	}
 
 	ret.RunnerResults = *periodic.Merge(rr1.Result(),
-                                        rr2.Result(),
-                                        percList)
+		rr2.Result(),
+		percList)
 
 	for k, v := range rr2.RetCodes {
-    	_, ok := rr1.RetCodes[k]
+		_, ok := rr1.RetCodes[k]
 
-    	if ok {
-    		rr1.RetCodes[k] += rr2.RetCodes[k]
-    	} else {
-    		rr1.RetCodes[k] = v
-    	}
+		if ok {
+			rr1.RetCodes[k] += rr2.RetCodes[k]
+		} else {
+			rr1.RetCodes[k] = v
+		}
 	}
 
 	ret.RetCodes = rr1.RetCodes
--- a/fortio_main.go
+++ b/fortio_main.go
@@ -86,7 +86,6 @@ var (
 	numThreadsFlag    = flag.Int("c", defaults.NumThreads, "Number of connections/goroutine/threads")
 	durationFlag      = flag.Duration("t", defaults.Duration, "How long to run the test or 0 to run until ^C")
 	percentilesFlag   = flag.String("p", "50,75,90,99,99.9", "List of pXX to calculate")
-	mergeFilesFlag    = flag.String("files", "", "List of json files to merge")
 	resolutionFlag    = flag.Float64("r", defaults.Resolution, "Resolution of the histogram lowest buckets in seconds")
 	goMaxProcsFlag    = flag.Int("gomaxprocs", 0, "Setting for runtime.GOMAXPROCS, <1 doesn't change the default")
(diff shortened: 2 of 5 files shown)


ldemailly:
Thanks a lot for the PR ! I haven't looked yet but

can you make sure to address the linter failure and code climate doc requests

also need tests

saimsalman36:
Sure.

fortio_main.go
```diff
@@ -85,6 +86,7 @@ var (
 	numThreadsFlag    = flag.Int("c", defaults.NumThreads, "Number of connections/goroutine/threads")
 	durationFlag      = flag.Duration("t", defaults.Duration, "How long to run the test or 0 to run until ^C")
 	percentilesFlag   = flag.String("p", "50,75,90,99,99.9", "List of pXX to calculate")
+	mergeFilesFlag    = flag.String("files", "", "List of json files to merge")
```
ldemailly: how is the list specified ? (what is the separator) or does it work like proxiesFlags where you can have many -P ?

files is also a bit vague.

As this is just going to be a new command, like `fortio merge ...` then no need for flags just have that command and the rest of the args are the files (a flag would be some -include-file or such to read files from a file/stdin instead of command line)

You also probably want to make an "fmerge" like there is an fcurl and echosrv standalone options

fortio_main.go
```diff
@@ -280,7 +349,7 @@ func fortioLoad(justCurl bool, percList []float64) {
 		Out:         out,
 		Labels:      labels,
 		Exactly:     *exactlyFlag,
-		Jitter:      *jitterFlag,
+		// Jitter:      *jitterFlag,
```
ldemailly: oversight ?

stats/stats.go
```diff
@@ -308,6 +311,50 @@ func (e *HistogramData) CalcPercentile(percentile float64) float64 {
...
+		if v == value {
+			return p
+		}
+	}
+	return -1
+}
+
+// Import translate the external representation of the histogram data in
+// an internally usable one.
+func (e *HistogramData) Import() *Histogram {
```
ldemailly, line 340: definitively need some tests that you get indeed the same histogram back in/out

stats/stats.go
```diff
@@ -308,6 +311,50 @@ func (e *HistogramData) CalcPercentile(percentile float64) float64 {
...
+	res := NewHistogram(e.Offset,
+		e.Divider)
+	res.Counter.Count = e.Count
+	res.Counter.Min = e.Min
+	res.Counter.Max = e.Max
+	res.Counter.Sum = e.Sum
+	res.Counter.sumOfSquares = sumOfSquares(e.StdDev, e.Sum, e.Count)
+
+	for idx, bucket := range e.Data {
```
ldemailly: to be considered you can probably use transfer-like code too

stats/stats.go
```diff
@@ -308,6 +311,50 @@ func (e *HistogramData) CalcPercentile(percentile float64) float64 {
...
+		if idx < len(e.Data)-1 {
```
ldemailly: I don't think you can use the idx ? the exported data can have holes (it doesn't export 0s) while the internal data structure doesn't

ldemailly (dismissed):
thanks for the contribution !

see above and below

stats/stats.go
saimsalman36: But this idx is looping over the external data structure and using the offset and divider filling in the histogram at the respective places. So even if there is a hole it shouldn't affect it.

ldemailly:
the dockercloud failure is
```
--- FAIL: TestStartMaxQps (0.16s)
periodic_test.go:139: MaxQpsTest executed unexpected number of times 10 instead 12
periodic_test.go:142: MaxQpsTest executed unexpected number of times 10 instead 12
```
I retried it, let's see and will check tomorrow, it also prevented the coverage from re-running and I'd like to know what it does (maybe you can add a dummy change to retrigger if that doesn't happen)

saimsalman36:
Its passing on my end.

```
=== RUN   TestStartMaxQps
11:48:07 V periodic.go:195> Negative qps -1.000000 means max speed mode/no wait between calls
11:48:07 I periodic.go:61> Making 4 clone of &{count:0xc000110bc8 lock:0xc000110be0}
Starting at max qps with 4 thread(s) [gomax 12] for 140ms
11:48:07 V periodic.go:228> WATCHER 1 First outstanding run starting, catching signal
11:48:07 V periodic.go:234> WATCHER 1 starting new watcher for signal! chan  g 0xc0001b2cc0 r 0xc0001f0720 (8)
11:48:07 I periodic.go:578> T003 ended after 150.538892ms : 3 calls. qps=19.928404946676505
11:48:07 I periodic.go:578> T001 ended after 150.547231ms : 3 calls. qps=19.927301087324548
11:48:07 I periodic.go:578> T002 ended after 150.542456ms : 3 calls. qps=19.927933153953592
11:48:07 I periodic.go:578> T000 ended after 150.610454ms : 3 calls. qps=19.91893603879582
Ended after 150.668421ms : 12 calls. qps=79.645
Aggregated Function Time : count 12 avg 0.050157137 +/- 6.362e-05 min 0.050091917 max 0.050296697 sum 0.601885645
>= 0.0500919 <= 0.0502967 , 0.0501943 , 100.00, 12
# target 90% 0.0502744
11:48:07 V periodic.go:500> RUNNER r.Stop not already closed, closing
11:48:07 V periodic.go:271> Abort called for 0xc00022c780 &{RunType: Runners:[0xc000134bc0 0xc000134bc0 0xc000134bc0 0xc000134bc0] QPS:-1 Duration:140ms NumThreads:4 Percentiles:[90] Resolution:0.001 Out:0xc00000e010 Labels: Stop:0xc000134c10 Exactly:0 Jitter:false}
11:48:07 V periodic.go:90> Closing 0xc0001f0720
--- PASS: TestStartMaxQps (0.15s)
```

ldemailly:
it passed, the coverage numbers seem wrong though, can you just add/remove some white space somewhere and update the PR to retrigger the builds

stats/stats_test.go
```diff
@@ -136,6 +136,76 @@ func TestZeroDivider(t *testing.T) {
...
+	}
+
+	for _, test := range tests {
+		if !reflect.DeepEqual(test.actual, test.expected) {
+			t.Errorf("%s: got %+v, not as expected %+v", test.msg, test.actual, test.expected)
+		}
+	}
+}
+
+func TestImportFunction3(t *testing.T) {
```
ldemailly, line 182: maybe one more test where you generate random points (and do the check in a loop, like 100 random histograms with 100 points in each)

fhttp/httprunner.go
```diff
@@ -49,6 +49,51 @@ type HTTPRunnerResults struct {
 	aborter *periodic.Aborter
 }
 
+// Merge two different HTTPRunnerResults
+func Merge(rr1 *HTTPRunnerResults,
```
ldemailly, line 53: it's great you have tests for the lower level (stats) one but this one also needs tests

fortio_main.go
```diff
@@ -226,6 +231,92 @@ func main() {
 	}
 }
 
+func generateJSONFile(res periodic.HasRunnerResult, out io.Writer) {
```
ldemailly, line 234: while that code was already somewhat present and already somewhat untested (or at least not tested through unit tests; do check WebTest.sh for e2e tests...) it'd be nice to ideally move that function in one of the util pkg and test it there

stats/stats_test.go
saimsalman36: Hey Laurent,

So while writing this test case I realized there is a bug in my code. If all the values are mapped to one bucket (the JSON file contains one bucket), I sometimes incorrectly generate the original Histogram.

The reason being that in the Export function both the Start and End of the bucket are over-written by the min and max hence I lose the information about which slot in the histogram contains the actual count. This causes there to be an off-by-one error.

One solution might be to modify the Export function but I wanted to discuss this before I modified it.

ldemailly: I don't get the jump between the values set to min and max (which helps making the interval more useful/increases the information available directly) and not being able to add that data correctly later ? if it was 1 bucket it'll be still one bucket... as you already export the offset and divisor, you can just make the same histogram and add the data back to it ?

ldemailly:
see also https://github.com/fortio/fortio/blob/master/stats/stats.go#L462

stats/stats_test.go
saimsalman36:
```
val := int32(math.Round((e.BucketMax - res.Offset) / res.Divider))
res.Hdata[indexSlice(histogramBucketValues, val)] = int32(bucket.Count)
```

To calculate the ```val``` I have to round because the ```histogramBucketValues``` contain exact integers. But when the bucket start or end value is replaced by the min or max this round might cause the data to be filled in the incorrect index (the next index). 

As I don't have the original values the Round might give the incorrect value.

This case will only occur if all the values are mapped to the same bucket (as then both the start and end values are replaces) and the bucket is one of the initial indices in the histogram [as initially the values are close together 0,1,2]

ldemailly:
ping?

ldemailly:
ping ping :)

ldemailly:
closing this for now but do reopen if you want

PR #349, opened 2019-11-06
support mTLS, passing Cert and Key to http client

hao1939:
fortio has the options '-cert' and '-key', but it's not used by HTTP client.
The commit passing the options to HTTP client and setup the client side certificate.


fhttp/http_client.go
```diff
@@ -388,9 +391,20 @@ func NewStdClient(o *HTTPOptions) *Client {
...
+	if o.https {
+		tr.TLSClientConfig = &tls.Config{}
+		if o.Insecure {
+			log.LogVf("using insecure https")
+			tr.TLSClientConfig.InsecureSkipVerify = true
+		}
+		if len(o.Cert) > 0 && len(o.Key) > 0 {
+			cert, err := tls.LoadX509KeyPair(o.Cert, o.Key)
+			if err != nil {
+				log.LogVf(fmt.Sprintf("%v", err))
```
ldemailly: this probably should be a log.Err (maybe even fatal)

ldemailly:
ping

fhttp/http_client.go
hao1939: got you, changed to log.Err.

ldemailly:
one issue is this isn't plugged into/exercisable from the command line

ldemailly:
actually I was wrong it's fine for client side, will merge (would be nice to have client side too but I will get to that one)

thanks for the contribution! (sorry for the delay, though ideally a test would have been great too)

ldemailly (approved):

fhttp/http_client.go
```diff
@@ -167,6 +167,9 @@ type HTTPOptions struct {
 	FollowRedirects   bool // For the Std Client only: follow redirects.
 	initDone          bool
 	https             bool   // whether URLSchemeCheck determined this was an https:// call or not
+	Cacert            string // `Path` to a custom CA certificate file to be used
```
ldemailly, line 170: this isn't used by the code below :(

PR #352, opened 2020-01-24
fix a typo

ptux:
fix a typo for `-t` option


ldemailly:
30min doesnt work?

let me fix the tests

ldemailly (approved):
you are right: https://golang.org/pkg/time/#ParseDuration only "m" can be used for minutes

ldemailly:
thx for that fix!

PR #355, opened 2020-02-21
make the fortio json filename aka ID to be longer for being able to include much more meaningful test info

carolynhu:
Originally I was solving this issue filed here:https://github.com/istio/istio/issues/21289

While digging into the code, I found it was originated from fortio side. Since in the fortio jsonfilename aka ID() function sets the length limit to be 64. 

Offline discussed with @mandarjog and xinnan, we can either use the unix epoch time format to get rid of this https://github.com/fortio/fortio/blob/master/periodic/periodic.go#L573 formatDate() func, or make the length limit to be larger.

While, after comparing with the unix time with the time generated from formatDate() func, I found that there is not so much decrease in length. Like unix time: `1582245603` is 10 chars. The time generated from formatDate(): `2020-02-20-114432` is 17 chars. While, the drawback of using unix time is that we have to do a conversion to know the exact date and time, which is inconvenience.

Therefore, I decide to only change the length limit.


mandarjog:
@ldemailly can we get a review on this?

mandarjog (approved):
lgtm

costinm (approved):

ldemailly:
sorry I was fixing the tests earlier but they are failing because fortio isn’t running anymore on fortio.istio.io. I was looking at a free tier option or some other way to run fortio somewhere on the internet.

ldemailly:
@carolynhu can you give me an example of when you need between 64 and 96 long ids? (the timestamp alone as you noted and as the comment shows is 17)

ldemailly:
also a test falling in between would be good

ldemailly:
I signed up for a free GCP trial and let's see how quickly the $300 get consumed

carolynhu:
@ldemailly Yeah, sure, here is an example:
Screen Shot 2020-02-26 at 4 23 16 PM

This is the fortio json data generated through istio perf pipeline. As you can see the the `serveronly` and `clientonly` are being truncated. That's why we want to enlarge the maximum size of the filename.

ldemailly:
ic, I guess the "nullvm" part isn't too interesting but it doesn't hurt I suppose... maybe we can use max file path or such

ldemailly (changes requested):
you need to update the failing test:
--- FAIL: TestID (0.00s)
    periodic_test.go:249: id: got 2001-01-02-030405_123456789012345678901234567890123456789012345678901234567890, not as expected 2001-01-02-030405_1234567890123456789012345678901234567890123456

mandarjog:
@carolynhu Please address the test failure when you get a chance.

carolynhu:
@ldemailly @mandarjog test failure is fixed, please take a look, thanks

ldemailly:
thanks! not sure why circleci isn’t reporting the last green build but will merge as it did pass

PR #363, opened 2020-06-19
Switching to go1.14 and go modules and update all dependencies

ldemailly:
No more vendor/ toml or dep
Switched version to 2.0.0 to indicate breaking compatibility with go1.8 which was maintained up to 1.3.1
updated to the latest grpc (1.29.1) and other dependencies


ldemailly:
docker cloud build seems to be failing because it's using vendor/ for no reason (there is no more submodule but it seems to keep it anyway)

PR #371, opened 2020-07-21
new option SetFlag handler to change dynamic flags from urls. moved e…

ldemailly:
…ndpoint to its own package so it can use fortio log and fhttp


ldemailly:
- [x] note: json needs multi line edit

PR #379, opened 2020-07-30
update build image to v21 and use golangci-lint

### LINKED ISSUE #378: switch linters to  golangci-lint

ldemailly:
fix #378


ldemailly:
we'll follow up on #380 for remaining lint fixes, it's good already to catch any new issue and build from newer image and golang 1.14.6

PR #383, opened 2020-08-03
prepare for next tag and update readme (and was testing lint)

ldemailly:


ldemailly:
works:
Screen Shot 2020-08-03 at 11 31 37 AM

PR #410, opened 2020-10-07
Adding fan out forwarder http proxy mode (-M ...)

### LINKED ISSUE #160: add tee/fanout mode

ldemailly:
Fixes #160


ldemailly:
functional but the connections aren't reused

PR #471, opened 2021-06-03
Support -cacert for http calls

### LINKED ISSUE #329: add support for custom CA for https requests (in addition to current grpc support)
Related to #315 but we have -cacert for grpc but that doesn't work/apply when making https requests for which the only option is to change the image/system CAs or switch to insecure

Code sample:
https://forfuncsake.github.io/post/2017/08/trust-extra-ca-cert-in-go-app/

ldemailly:
fixes #329

- move ca,cert,key flags to common (http) flags
- improve flags doc
- fixed cert,key error message
- use ca if provided for std https client

manually tested (should have auto test too... later...)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ version: 2
 defaultEnv: &defaultEnv
     docker:
       # specify the version
-      - image: docker.io/fortio/fortio.build:v33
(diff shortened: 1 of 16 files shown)


bseenu:
LGTM 👍

PR #483, opened 2021-07-21
allow utilizing gRPC's proxyless xDS support

stevenctl:
I'm currently implementing control-plane support in Istio for [proxyless gRPC ](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md). As part of this effort I'm using Fortio to load-test, comparing plaintext gRPC, Envoy+mTLS and proxyless+mTLS performance. 

The necessary steps to enable a gRPC client application are:
* Importing `"google.golang.org/grpc/xds"` to install the `xds:///` resolver
* To allow the client to use security configuration provided over xDS:
  * GRPC_XDS_EXPERIMENTAL_SECURITY_SUPPORT must be on
  * `xds.NewClientCredentials` must be provided (adding the insecure fallback means we fallback only if the xDS configuration doesn't provide security configuration)
* The app must be run with the GRPC_XDS_BOOTSTRAP environment variable set (we will do this on the istio side)
* The `--profile` flag wasn't actually passed to grpcrunner
* The random payload generator must generate valid UTF-8 bytes or gRPC will fail. 

I'm trying this out in my own tests now, but it could be useful for upstream.


stevenctl:
Tested with `kubectl --namespace twopods-istio exec fortioclient-5d4db69df6-5sxdv  -- fortio load  -jitter=False -c 2 -qps 1000 -t 240s -a -r 0.001  -grpc -ping -httpbufferkb=128 -labels 01565ec6_qps_1000_c_2_1024_v2-stats-nullvm_both -payload-size 1024 xds:///fortioserver.twopods-istio.svc.cluster.local:8079`

ldemailly:
is it backward compatible; as in without the env it just doesn't use xds?

stevenctl:
Everything is still fully backwards compatible. You need both the env and the scheme to actually utilize XDS.

fgrpc/grpcrunner.go
```diff
@@ -61,6 +76,7 @@ func Dial(o *GRPCRunnerOptions) (conn *grpc.ClientConn, err error) {
 			return net.Dial(fnet.UnixDomainSocket, o.UnixDomainSocket)
 		}))
 	}
+	log.Infof("Normalized server address")
```
ldemailly: leftover? remove? or make a logVf maybe but print something not just the text?

fgrpc/grpcrunner.go
```diff
@@ -51,6 +58,14 @@ func Dial(o *GRPCRunnerOptions) (conn *grpc.ClientConn, err error) {
 	case strings.HasPrefix(o.Destination, fnet.PrefixHTTPS):
 		creds := credentials.NewTLS(&tls.Config{InsecureSkipVerify: o.Insecure}) // nolint: gosec // explicit flag
 		opts = append(opts, grpc.WithTransportCredentials(creds))
+	case strings.HasPrefix(o.Destination, fnet.PrefixXDS) && strings.EqualFold(os.Getenv(xdsSecuritySupportEnv), "true"):
```
ldemailly: given previous version of fortio didn't know about xds:// anyway, why add the env check in addition to the prefix check?

fgrpc/grpcrunner.go
```diff
@@ -279,6 +295,11 @@ func grpcDestination(dest string) (parsedDest string) {
 		port = fnet.StandardHTTPSPort
 		log.Infof("stripping https scheme. grpc destination: %v. grpc port: %s",
 			parsedDest, port)
+	case strings.HasPrefix(dest, fnet.PrefixXDS):
+		if _, _, err := net.SplitHostPort(dest[len(fnet.PrefixXDS):]); err == nil {
+			return dest
+		}
+		return dest + fnet.NormalizePort(fnet.DefaultGRPCPort)
```
ldemailly, line 298: that's not quite right for the case there would be a port in the xds://url ?
why not letting it fall through like the rest of the code after removing `xds://` like `http://` is done?
is there the notion of host:port in xds?

fgrpc/grpcrunner.go
```diff
@@ -30,9 +30,16 @@ import (
 	"fortio.org/fortio/periodic"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/credentials/xds"
 	"google.golang.org/grpc/health/grpc_health_v1"
+	_ "google.golang.org/grpc/xds"
```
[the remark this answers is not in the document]
ldemailly: it could I suppose use a comment indeed

fgrpc/grpcrunner.go
stevenctl: I was under the impression gRPC would reject this option if the env wasn't set but it actually seems to be fine.

fgrpc/grpcrunner.go
stevenctl: This does: If there is a port, just keep it; but for no port use the default port. I have a test to check that it works.

I think the reason I didn't fallthrough is it would strip the `xds:///` which indicates to gRPC to use the resolver.

ldemailly: can you give a full example of a couple of url; or point me to some; I'm very unfamiliar with xds

fgrpc/grpcrunner_test.go
```diff
@@ -385,6 +385,16 @@ func TestGRPCDestination(t *testing.T) {
 			"https://2001:dba::1/",
 			"[2001:dba::1]:443",
 		},
+		{
+			"XDS url with port",
+			"xds:///fortioserver.twopods-istio.svc.cluster.local:8079",
```
ldemailly: try changing the port to show it sticks

fgrpc/grpcrunner.go
ldemailly: reason I ask is
```golang
	if _, _, err := net.SplitHostPort(parsedDest); err == nil {
```
below is supposed to handle that

fnet/network.go
```diff
@@ -80,6 +82,8 @@ func ChangeMaxPayloadSize(newMaxPayloadSize int) {
 	if err != nil {
 		log.Errf("Error changing payload size, read for %d random payload failed: %v", newMaxPayloadSize, err)
 	}
+	// slower but we seem to only do this on init anyway, and this guarantees UTF-8
+	Payload = []byte(base64.StdEncoding.EncodeToString(Payload)[:len(Payload)])
```
ldemailly, line 86: it's not quite same pseudo random, it's now pseudo random utf8, except it can still end up truncated and invalid at the end; what breaks when left as bytes?

stevenctl: gRPC will refuse to send it if it's not valid utf-8

ldemailly: There must be a type that is bytes, not string?

stevenctl: Do you mean here in the generation part, or sending a byte-array directly in the gRPC payload rather than string (changing the protos)? 

For the generation part I didn't see anything buillt-in that could do this. If we generate `< len(payload)` bytes we can manually zero-pad the start of the byte array to be utf-8 compliant.

fgrpc/grpcrunner.go
stevenctl: Updated the test. AFAICT this is just a bit of code-duplication here to avoid adding something like `if xds { preserve scheme = true } ... if preserve scheme { ... }` in the outer block.

fnet/network.go
ldemailly: 3 different issues

- I don't quite believe there isn't a way to send bytes; maybe if this is for the PingServer proto; we can just change the proto if there is a type mismatch

- Can you show me in the unchanged code how to reproduce the failure? (or give details; refuse to send how exactly?)

- If we really wanted to send utf string then we should indeed generate that rather than truncate (but that's last resort and shouldn't be necessary)

fgrpc/grpcrunner.go
```diff
@@ -51,6 +55,14 @@ func Dial(o *GRPCRunnerOptions) (conn *grpc.ClientConn, err error) {
 	case strings.HasPrefix(o.Destination, fnet.PrefixHTTPS):
 		creds := credentials.NewTLS(&tls.Config{InsecureSkipVerify: o.Insecure}) // nolint: gosec // explicit flag
 		opts = append(opts, grpc.WithTransportCredentials(creds))
+	case strings.HasPrefix(o.Destination, fnet.PrefixXDS):
+		// when using the xds:/// scheme, use
+		creds, err := xds.NewClientCredentials(xds.ClientOptions{FallbackCreds: insecure.NewCredentials()})
```
ldemailly, line 60: why is insecure hardcoded vs using o.Insecure?

ldemailly: I guess maybe because of "Fallback" what is fallback for, is it required?

fgrpc/ping.proto
```diff
@@ -2,6 +2,7 @@
 // protoc ping.proto --go_out=plugins=grpc:.
 syntax = "proto3";
 package fgrpc;
+option go_package="../fgrpc";
```
ldemailly: that looks very odd; we're already in fgrpc?

fgrpc/ping_grpc.pb.go
```diff
@@ -0,0 +1,101 @@
+// Code generated by protoc-gen-go-grpc. DO NOT EDIT.
```
ldemailly, line 1: shouldn't be a new file... there is already  ping.pb.go

fgrpc/ping.proto
stevenctl: This is a hack for now –  I don't know why my protoc makes me specify this, makes me use a `/` character. 

An alternative would be `go_package ./fgrpc` and then running the protoc from `..`

I haven't had a chance to clean this up or figure out the right way.

fgrpc/ping_grpc.pb.go
stevenctl: The protoc commands as originally suggested in the proto file don't work with the latest protoc (3.17.3)

And newer versions split the gRPC parts into their own file.

fgrpc/ping.proto
stevenctl: Looks like specifying the fully qualified package + using source_relative is better.

fgrpc/grpcrunner.go
stevenctl: FallbackCreds says: "if the XDS server doesn't send me security configuration, fallback to this mode". This basically gives full control of security options to the control plane. I could require the `o.Insecure` to allow the fallback?

fnet/network.go
stevenctl: > I don't quite believe there isn't a way to send bytes; maybe if this is for the PingServer proto; we can just change the proto if there is a type mismatch

We could totally change the proto to send bytes. I was just trying to avoid changing the protos originally. 

> Can you show me in the unchanged code how to reproduce the failure? (or give details; refuse to send how exactly?) 

I'll get repro steps later today. 

> If we really wanted to send utf string then we should indeed generate that rather than truncate (but that's last resort and shouldn't be necessary)

Agreed – but I think the goal is "create a payload of size n" which `byte[]` is better suited for.

BrendanNRyan:
Could you please advise the status of this Pull request. Mainly, are there any plans to merge this functionality to Fortio master branch ?

ldemailly:
> Could you please advise the status of this Pull request. Mainly, are there any plans to merge this functionality to Fortio master branch ?


Well not as it is but if we can resolve the issues, sure

This being said it's a big change bringing in a bunch of dependencies; so I am conflicted.

ldemailly:
closing this for now but do reopen if you want

Issue #484, opened 2021-07-30
dfags: support for slices (arrays) of struct in DynJSON


ldemailly:
Right now the type for `dflag.DynJSON` has to be `&someType` and can't be directly `[]someType`

forcing you to create an extra
```golang
type someTypeList struct {
	List []someType
}
// to pass to
myList := dflag.DynJSON(flag.CommandLine, "mylist", &someTypeList{}, "...t")
```
instead of directly

ldemailly:
fixed by #491 (Thanks Jerry!) need to be released in 1.18.1 (or 1.19.0)

PR #491, opened 2021-10-20
Direct support of Json Array in DynJson

wuhaoyujerry:
Address the issue #484

Provide direct support of json array in DynJson. Added related test cases.

### DIFF
--- a/dflag/dynjson.go
+++ b/dflag/dynjson.go
@@ -16,10 +16,14 @@ import (
 // New values based on the default constructor of `value` type will be created on each update.
 func DynJSON(flagSet *flag.FlagSet, name string, value interface{}, usage string) *DynJSONValue {
 	reflectVal := reflect.ValueOf(value)
-	if (reflectVal.Kind() != reflect.Ptr || reflectVal.Elem().Kind() != reflect.Struct) &&
-		(reflectVal.Kind() != reflect.Ptr || reflectVal.Elem().Kind() != reflect.Slice) {
+
+	notPtr := reflectVal.Kind() != reflect.Ptr
+	notStruct := reflectVal.Elem().Kind() != reflect.Struct
+	notSlice := reflectVal.Elem().Kind() != reflect.Slice
+	if notPtr || (notStruct && notSlice) {
 		panic("DynJSON value must be a pointer to a struct or to a slice")
 	}
+
 	dynValue := &DynJSONValue{
 		ptr:        unsafe.Pointer(reflectVal.Pointer()),
 		structType: reflectVal.Type().Elem(),
--- a/dflag/dynjson.go
+++ b/dflag/dynjson.go
@@ -17,10 +17,8 @@ import (
 func DynJSON(flagSet *flag.FlagSet, name string, value interface{}, usage string) *DynJSONValue {
 	reflectVal := reflect.ValueOf(value)
 
-	notPtr := reflectVal.Kind() != reflect.Ptr
-	notStruct := reflectVal.Elem().Kind() != reflect.Struct
-	notSlice := reflectVal.Elem().Kind() != reflect.Slice
-	if notPtr || (notStruct && notSlice) {
+	if reflectVal.Kind() != reflect.Ptr ||
+		(reflectVal.Elem().Kind() != reflect.Struct && reflectVal.Elem().Kind() != reflect.Slice) {
 		panic("DynJSON value must be a pointer to a struct or to a slice")
 	}
 
--- a/dflag/dynjson.go
+++ b/dflag/dynjson.go
@@ -16,8 +16,9 @@ import (
 // New values based on the default constructor of `value` type will be created on each update.
 func DynJSON(flagSet *flag.FlagSet, name string, value interface{}, usage string) *DynJSONValue {
 	reflectVal := reflect.ValueOf(value)
-	if reflectVal.Kind() != reflect.Ptr || reflectVal.Elem().Kind() != reflect.Struct {
-		panic("DynJSON value must be a pointer to a struct")
+	if (reflectVal.Kind() != reflect.Ptr || reflectVal.Elem().Kind() != reflect.Struct) &&
+		(reflectVal.Kind() != reflect.Ptr || reflectVal.Elem().Kind() != reflect.Slice) {
+		panic("DynJSON value must be a pointer to a struct or to a slice")
 	}
 	dynValue := &DynJSONValue{
 		ptr:        unsafe.Pointer(reflectVal.Pointer()),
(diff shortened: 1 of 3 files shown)


dflag/dynjson.go
```diff
@@ -16,8 +16,9 @@ import (
 // New values based on the default constructor of `value` type will be created on each update.
 func DynJSON(flagSet *flag.FlagSet, name string, value interface{}, usage string) *DynJSONValue {
 	reflectVal := reflect.ValueOf(value)
-	if reflectVal.Kind() != reflect.Ptr || reflectVal.Elem().Kind() != reflect.Struct {
-		panic("DynJSON value must be a pointer to a struct")
+	if (reflectVal.Kind() != reflect.Ptr || reflectVal.Elem().Kind() != reflect.Struct) &&
+		(reflectVal.Kind() != reflect.Ptr || reflectVal.Elem().Kind() != reflect.Slice) {
```
ldemailly: seems that the first part can be factored in?

`notPtr || (notStruct && notSlice)` ?

dflag/dynjson_test.go
```diff
@@ -12,19 +12,29 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-var defaultJSON = &outerJSON{
+var defaultJSONOne = &outerJSON{
```
ldemailly: maybe leave that one alone for most tests - and just add 1 new deserialization test of a slice/list/array?

dflag/dynjson_test.go
```diff
@@ -12,19 +12,29 @@ import (
...
 
+var defaultJSONTwo = &outerJSON{
+	FieldInts:   []int{1, 3, 3, 7},
+	FieldString: "non-empty",
+	FieldInner: &innerJSON{
+		FieldBool: true,
+	},
+}
+
+var defaultJSONArray = &[]outerJSON{*defaultJSONOne, *defaultJSONTwo}
```
ldemailly: given you use the same values; how is that better than `outerJSON{*defaultJSON, *defaultJSON)`?
(what does Two adds?)

dflag/dynjson_test.go
ldemailly: (minimizing the diff helps having less things to review; unless it's a much cleaner/necessary refactor; which then would be a shallow change in its own diff)

ldemailly:
Thanks for working on this! most appreciated!

dflag/dynjson.go
wuhaoyujerry: fixed

dflag/dynjson_test.go
wuhaoyujerry: reverted

dflag/dynjson_test.go
wuhaoyujerry: Changed the input data `defaultJSONTwo` to have different value

dflag/dynjson.go
ldemailly: Sorry, when I used "notPtr" I didn't mean to create a variable for that, it was just so on my phone I didn't have to type the whole `reflectVal.Kind() != reflect.Ptr`

Isn't it likely that we need the short circuiting? ie one shouldn't try to get reflectVal.Elem()[.Kind()] if reflectVal.Kind() != reflect.Ptr  ?

dflag/dynjson.go
```diff
@@ -16,9 +16,14 @@ import (
 // New values based on the default constructor of `value` type will be created on each update.
 func DynJSON(flagSet *flag.FlagSet, name string, value interface{}, usage string) *DynJSONValue {
 	reflectVal := reflect.ValueOf(value)
-	if reflectVal.Kind() != reflect.Ptr || reflectVal.Elem().Kind() != reflect.Struct {
-		panic("DynJSON value must be a pointer to a struct")
+
+	notPtr := reflectVal.Kind() != reflect.Ptr
+	notStruct := reflectVal.Elem().Kind() != reflect.Struct
```
ldemailly: (see other comment but also:) check if this line crashes/panics when passed an interface that isn't a pointer?
(maybe on https://play.golang.org/ )

dflag/dynjson_test.go
```diff
@@ -20,6 +20,16 @@ var defaultJSON = &outerJSON{
 	},
 }
 
+var defaultJSONTwo = &outerJSON{
+	FieldInts:   []int{2, 3, 4, 5},
+	FieldString: "non-empty",
```
ldemailly, line 25: "value2" maybe ?

dflag/dynjson_test.go
```diff
@@ -56,7 +66,7 @@ func TestDynJSON_FiresValidators(t *testing.T) {
 
 	DynJSON(set, "some_json_1", defaultJSON, "Use it or lose it").WithValidator(validator)
 
-	assert.NoError(t, set.Set("some_json_1", `{"ints": [42], "string":"bar"}`), "no error from validator when inputo k")
+	assert.NoError(t, set.Set("some_json_1", `{"ints": [42], "string":"bar"}`), "no error from validator")
```
ldemailly, line 69: thx, that's a weird typo!

dflag/dynjson.go
ldemailly: if it doesn't crash when passed a non ptr then you can leave it as you did, otherwise re-inline just with the () and sharing the first not a pointer case as first short circuit

dflag/dynjson.go
ldemailly: panic: reflect: call of reflect.Value.Elem on struct Value
instead of DynJSON value must be a pointer to a struct

dflag/dynjson.go
wuhaoyujerry: thx for pointing out the issue. I changed it back to inline to use the short circuit.

dflag/dynjson.go
wuhaoyujerry: fixed, lemme know if you still have question

dflag/dynjson_test.go
wuhaoyujerry: I changed it to `JSONValue2` because I guess the second value shouldn't be default. Lemme know if I understand this correctly. Also, what's the naming convention for this kind of variables? Thx.

ldemailly: I meant you did change the ints to be different from the other entry, bit the FieldString had same value; to check the serialization/deserialization isn't accidentally returning same/mixed up values; having unique values for all test values is useful

ldemailly: i guess it's not tested per se if I read the test correctly the values are changed in the test anyway so, sgtm

ldemailly:
Going to merge this! thx a lot for the contribution!

ldemailly (approved):

PR #493, opened 2021-10-23
rest/json run trigger api + more fixes/improvments

### LINKED ISSUE #492: Async server run trigger mode
Right now running a test on fortio server url is synchronous
Let's add async mode that just returns the runid 
(and/or cancel url; like the interrupt button when using web ui)

Consider making it JSON too

Potentially making it compatible with flagger webhooks (ie read from "metadata" if it is in the payload)

ldemailly:
new endpoints:
- `fortio/rest/run` similar to fetching fortio/ without needing `load=Start` to start a run
  - plus `async` query arg or json value "`on`" will make the run asynchronous (returns just the runid of the run instead of waiting for the result)
  - plus read all the run configuration from either query args or jsonPath posted info (so for a large payload; must be put in json vs POST query args in `/fortio/` endpoint)
  - compatible with [flagger](https://github.com/fluxcd/flagger) and other webhooks
e.g
```
curl -v -d '{"metadata": {"url":"localhost:8080", "c":"1", "n":"1", "async":"on", "save":"on"}}' \
     "localhost:8080/fortio/rest/run?jsonPath=.metadata"
```
makes a 1 connection 1 query run for localhost:8080 url asynchronously and saves results

or minimally:
```
curl -s -d '{"url":"localhost:8080"}' "localhost:8080/fortio/rest/run" | jq
```
- `fortio/rest/stop` stops all runs or by id

Also:
- added RunID in options and result, use it in filename generation when non 0 (server run ids auto increment)
- fixed bug where destination/url in ui wouldn't show unless labels were present
- show the runid there as (runid) prefix in UI

Fixes #492

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ You can set a default value for all these by passing `-echo-server-default-param
   * Download/sync from an Amazon S3 or Google Cloud compatible bucket listings [XML URLs](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html)
 
 * API to trigger and cancel runs from the running server (like the form ui but more directly and with `async=on` option)
-  * `/fortio/rest/run` starts a run; the arguments are either from the command line or from POSTed JSON; `jsonPath` can be provided to look for in a subset of the json object, for instance `jsonPath=metadata` allows to use the flagger webhook meta data for fortio run parameters (see #493).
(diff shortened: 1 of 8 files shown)


ldemailly:
manually tested (for now) see description. ideally would refactor uihandler some more and share + have automated tests.

ldemailly:
pre released as `fortio/fortio:1.18.0-pre1` tag/docker image

PR #494, opened 2021-11-05
create releases from github actions

ldemailly:
create release artifact, push to docker and github release page


ldemailly:
took 12 tags but finally fully working

PR #495, opened 2021-11-07
Add offset to the histogram of the response time

debbyku:
It is to add offsset to the histogram of the response time.

Currently we run the fortio on a service with predefined latency, say, 200ms, 500ms, 1s to simulate production behavior.  Fortio does not give accurate response time as the histogram of the response time is always starting from 0.   

In this PR, a new flag is added.  it is
-o :  offset to histogram of the response time
e.g.  if the latency of the service is 200ms,  i.e. sleep 200ms in the service.  the command for the fortio is : '
  
fortio load -o 0.2s    ......

### DIFF
--- a/fortio_main.go
+++ b/fortio_main.go
@@ -104,7 +104,7 @@ var (
 	durationFlag    = flag.Duration("t", defaults.Duration, "How long to run the test or 0 to run until ^C")
 	percentilesFlag = flag.String("p", "50,75,90,99,99.9", "List of pXX to calculate")
 	resolutionFlag  = flag.Float64("r", defaults.Resolution, "Resolution of the histogram lowest buckets in seconds")
-	offsetFlag      = flag.Float64("o", defaults.Offset, "Offset of the histogram data in seconds")
+	offsetFlag      = flag.Duration("o", defaults.Offset, "Offset duration of the histogram data")
 	goMaxProcsFlag  = flag.Int("gomaxprocs", 0, "Setting for runtime.GOMAXPROCS, <1 doesn't change the default")
 	profileFlag     = flag.String("profile", "", "write .cpu and .mem profiles to `file`")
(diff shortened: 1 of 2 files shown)


fortio_main.go
```diff
@@ -104,6 +104,7 @@ var (
 	durationFlag    = flag.Duration("t", defaults.Duration, "How long to run the test or 0 to run until ^C")
 	percentilesFlag = flag.String("p", "50,75,90,99,99.9", "List of pXX to calculate")
 	resolutionFlag  = flag.Float64("r", defaults.Resolution, "Resolution of the histogram lowest buckets in seconds")
+	offsetFlag      = flag.Float64("o", defaults.Offset, "Offset of the histogram data in seconds")
```
ldemailly: should probably be a duration (like -t)

debbyku: Code updated and pushed again

periodic/periodic.go
```diff
@@ -406,7 +408,8 @@ func (r *periodicRunner) Run() RunnerResults {
 	}
 	start := time.Now()
 	// Histogram  and stats for Function duration - millisecond precision
-	functionDuration := stats.NewHistogram(0, r.Resolution)
+	offsetSec := float64(r.Offset) / float64(time.Second)
```
ldemailly: https://pkg.go.dev/time#Duration.Seconds

ldemailly (changes requested):
see above

periodic/periodic.go
debbyku: thanks. It is much easier to get the seconds.  Code updated.

ldemailly (approved):
thanks!

ldemailly:
I should have thought about this sooner, but I am going to change the flag to keep the short `-o` for future

ldemailly:
#496 - so please use 1.18.2 and `-offset` - thanks!

debbyku:
thanks

在 2021年11月24日週三 02:21，Laurent Demailly ***@***.***> 寫道：

PR #499, opened 2021-12-06
Add --uniform flag

### LINKED ISSUE #498: Add --uniform mode
Currently, requests are sent in batches. This is not ideal for testing latency, IMO - neither fortio nor the backend can handle large bursts of requests as well as they could a steady stream. It also is less likely to be indicative of real world scenarios

`-jitter` is a step in the right direction, but still not ideal. It would be ideal if requests were as uniform as possible - for example, if we had 10qps we would send a request a 0, 0.1, 0.2, ... etc, regardless of number of connections.

The graph above plots every request made by fortio with `-c 128 -qps 128`. There are three groupings: a fork with a `--uniform` implemented and used, defaults, and then jitter.

You can see `-jitter` on its own reduces latency substantially - but the first request is not jittered so we get a spike which skews metrics. Additionally, its still pretty grouped.

Uniform mode shows much more reliable approach.

I have two implementation ideas:
1. For each thread, delay its start based on thread id. This is pretty simple:
```
delaySeconds := float64(perThreadQPS)-(float64(perThreadQPS)/float64(r.NumThreads)*float64(r.NumThreads-id))
start = start.Add(time.Duration(delaySeconds*float64(time.Second)))
```
2. Have a single ticker that triggers the appropriate goroutine to send a request

I have implemented 1 above and it seems to work decently. I am happy to send a PR but wanted to open issue for any discussion

howardjohn:
Fixes https://github.com/fortio/fortio/issues/498. See discussion in issue

### DIFF
--- a/periodic/periodic.go
+++ b/periodic/periodic.go
@@ -507,13 +507,19 @@ func runOne(id int, runnerChan chan struct{},
 	useExactly := (r.Exactly > 0)
 	f := r.Runners[id]
 	if useQPS && r.Uniform {
-		delayBetweenRequest := 1. / float64(perThreadQPS)
+		delayBetweenRequest := 1. / perThreadQPS
 		// When using uniform mode, we should wait a bit relative to our QPS and thread ID.
 		// For example, with 10 threads and 1 QPS, thread 8 should delay 0.7s.
 		delaySeconds := delayBetweenRequest - (delayBetweenRequest / float64(r.NumThreads) * float64(r.NumThreads-id))
 		delayDuration := time.Duration(delaySeconds * float64(time.Second))
 		start = start.Add(delayDuration)
-		time.Sleep(delayDuration)
+		log.Debugf("%s sleep %v for uniform distribution", tIDStr, delayDuration)
+		select {
+		case <-runnerChan:
+			return
+		case <-time.After(delayDuration):
+			// continue normal execution
+		}
 	}
 
 MainLoop:
--- a/periodic/periodic.go
+++ b/periodic/periodic.go
@@ -132,6 +132,11 @@ type RunnerOptions struct {
 	// Enabling jitter (+/-10%) allows these requests to be de-synchronized
 	// When enabled, it is only effective in the '-qps' mode.
 	Jitter bool
+	// When multiple clients are used to generate requests, they tend to send
+	// requests very close to one another, causing a thundering herd problem
+	// Enabling uniform causes the requests between connections to be uniformly staggered.
+	// When enabled, it is only effective in the '-qps' mode.
+	Uniform bool
 	// Optional run id; used by the server to identify runs.
 	RunID int64
 	// Optional Offect Duration; to offset the histogram function duration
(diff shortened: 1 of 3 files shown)


periodic/periodic.go
```diff
@@ -497,9 +502,19 @@ func runOne(id int, runnerChan chan struct{},
 	tIDStr := fmt.Sprintf("T%03d", id)
 	perThreadQPS := r.QPS / float64(r.NumThreads)
 	useQPS := (perThreadQPS > 0)
+
 	hasDuration := (r.Duration > 0)
 	useExactly := (r.Exactly > 0)
 	f := r.Runners[id]
+	if useQPS && r.Uniform {
+		delayBetweenRequest := 1. / float64(perThreadQPS)
```
ldemailly: perThreadQPS is already a float.

periodic/periodic.go
```diff
@@ -497,9 +502,19 @@ func runOne(id int, runnerChan chan struct{},
...
+		// When using uniform mode, we should wait a bit relative to our QPS and thread ID.
+		// For example, with 10 threads and 1 QPS, thread 8 should delay 0.7s.
+		delaySeconds := delayBetweenRequest - (delayBetweenRequest / float64(r.NumThreads) * float64(r.NumThreads-id))
+		delayDuration := time.Duration(delaySeconds * float64(time.Second))
```
ldemailly, line 514: Do you think this still works:
https://github.com/fortio/fortio/blob/v1.19.0/periodic/periodic.go#L529-L532
can you try/does it invalidate it (and maybe it's fine if it does for -uniform and low duration corner cases but I am curious)

periodic/periodic.go
```diff
@@ -497,9 +502,19 @@ func runOne(id int, runnerChan chan struct{},
...
+		start = start.Add(delayDuration)
+		time.Sleep(delayDuration)
```
ldemailly: in theory all sleep times are recorded through
sleepTimes.Record
but maybe this initial offsetting isn't - it will change somewhat the total duration of the run and the end case for the last threads I think (might miss one request at the end?)

what happens if you do
-qps 100 -c 10 -t 10s

in theory each connection would do exactly 100/10*10 = 100 calls and the whole run would be 10s - what does it do with -uniform?

ldemailly:
(and might want to have a test with the flag to get some coverage too)

periodic/periodic.go
howardjohn: I think with uniform we need to compromise on QPS or duration - we cannot respect both. Say we have 2 threads, 1 qps, 2s duration

Today we do:
```
         0s   01   2s
thread0: .         .
thread1: .         .
```

This PR does this - favoring exact QPS over exact duration
```
         0s   01   2s    3s
thread0: .         .
thread1:       .         .
```
So it goes past the duration by `delayDuration`.

The alternative would be to slightly increase the QPS to hit the target duration:
```
         0s   01   2s 
thread0: .     .
thread1:    .       .
```

But that seems less desired

periodic/periodic.go
howardjohn: Discussed a bit in the above case, it will extend the total duration a bit (100ms here).

I can add `sleepTimes.Record` if you prefer, but it seems like it shouldn't be included

Without it: `count 40 avg 1.2484387 +/- 0.0005214 min 1.2468261840000001 max 1.249850499 sum 49.937547`
With it: `count 50 avg 1.0888235 +/- 0.3443 min 0 max 1.24983677 sum 54.4411737`

We can see it really skews the results

ldemailly: thanks for the answers!

periodic/periodic.go
ldemailly: agreed, thx!

ldemailly:
Note: added support for uniform in the result json in #501 and in the rest api and UI #512 

all that ships in fortio 1.21.0

PR #500, opened 2021-12-06
Warmup requests in parallel

howardjohn:
Currently warmup takes `connections * latency` to complete. For high
connections and/or latency, this number is quite large. For example, if
I have 1000 connections and 250ms latency, this takes 4 minutes to warm
up.

With this patch, warmup is now `max(latency)`.

### DIFF
--- a/fhttp/httprunner.go
+++ b/fhttp/httprunner.go
@@ -24,6 +24,7 @@ import (
 	"fortio.org/fortio/log"
 	"fortio.org/fortio/periodic"
 	"fortio.org/fortio/stats"
+	"golang.org/x/sync/errgroup"
 )
 
 // Most of the code in this file is the library-fication of code originally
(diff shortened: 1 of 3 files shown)


fhttp/httprunner.go
```diff
@@ -24,6 +24,7 @@ import (
 	"fortio.org/fortio/log"
 	"fortio.org/fortio/periodic"
 	"fortio.org/fortio/stats"
+	"golang.org/x/sync/errgroup"
```
ldemailly: I prefer not adding more dependencies unless terribly valuable;
what does errgroup does that can't be done with sync.WaitGroup ?

ldemailly:
To do the initial connection in parallel I would think we'd (optionally) move it inside the existing goroutines (that are setup just after) ?

Btw this will have the opposite effect of increasing "burst" mentioned in #498 :)

howardjohn:
> Btw this will have the opposite effect of increasing "burst" mentioned in https://github.com/fortio/fortio/issues/498 :)

Only on the warmup though, right?

fhttp/httprunner.go
howardjohn: Not that much  - it is ultimately about 20 lines of code wrapping it to give error handling. I just inlined it instead now

ldemailly:
Thanks for this (and sorry for the delay)!

ldemailly:
> Only on the warmup though, right?

yes but I think it's quite a burst of connections/syn flood - further thinking about it I think we probably need this behind a flag or some option to spread to some maximum of parallelization or some time interval

PR #501, opened 2021-12-06
Add option for request access logs

howardjohn:
This allows fine grained traffic analysis. For example, this helped analyze https://github.com/fortio/fortio/issues/498

### DIFF
--- a/fortio_main.go
+++ b/fortio_main.go
@@ -397,7 +397,7 @@ func fortioLoad(justCurl bool, percList []float64) {
 		Offset:      *offsetFlag,
 	}
 	if *accessLogFileFlag != "" {
-		al, err := periodic.NewAccessLogger(*accessLogFileFlag, *accessLogFileFormat)
+		al, err := periodic.NewFileAccessLogger(*accessLogFileFlag, *accessLogFileFormat)
 		if err != nil {
 			_, _ = fmt.Fprintf(out, "Aborting because of %v\n", err)
 			os.Exit(1)
--- a/fortio_main.go
+++ b/fortio_main.go
@@ -174,6 +174,11 @@ var (
 	mirrorOriginFlag = flag.Bool("multi-mirror-origin", true, "Mirror the request url to the target for multi proxies (-M)")
 	multiSerialFlag  = flag.Bool("multi-serial-mode", false, "Multi server (-M) requests one at a time instead of parallel mode")
 	udpTimeoutFlag   = flag.Duration("udp-timeout", udprunner.UDPTimeOutDefaultValue, "Udp timeout")
+
+	accessLogFileFlag = flag.String("access-log-file", "",
+		"file to log all requests to. Maybe have performance impacts")
+	accessLogFileFormat = flag.String("access-log-format", "",
+		"format for access log. Support values: [json, influx]")
 )
 
 // nolint: funlen // well yes it's fairly big and lotsa ifs.
@@ -390,6 +395,14 @@ func fortioLoad(justCurl bool, percList []float64) {
 		Jitter:      *jitterFlag,
 		RunID:       *bincommon.RunIDFlag,
 		Offset:      *offsetFlag,
+	}
+	if *accessLogFileFlag != "" {
+		al, err := periodic.NewAccessLogger(*accessLogFileFlag, *accessLogFileFormat)
+		if err != nil {
+			_, _ = fmt.Fprintf(out, "Aborting because of %v\n", err)
+			os.Exit(1)
+		}
+		ro.AccessLogger = al
 	}
 	var res periodic.HasRunnerResult
 	var err error
--- a/fortio_main.go
+++ b/fortio_main.go
@@ -177,9 +177,9 @@ var (
 	udpTimeoutFlag   = flag.Duration("udp-timeout", udprunner.UDPTimeOutDefaultValue, "Udp timeout")
 
 	accessLogFileFlag = flag.String("access-log-file", "",
-		"file to log all requests to. Maybe have performance impacts")
-	accessLogFileFormat = flag.String("access-log-format", "",
-		"format for access log. Support values: [json, influx]")
+		"file `path` to log all requests to. Maybe have performance impacts")
+	accessLogFileFormat = flag.String("access-log-format", "json",
+		"`format` for access log. Supported values: [json, influx]")
 )
 
 // nolint: funlen // well yes it's fairly big and lotsa ifs.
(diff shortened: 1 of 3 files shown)


periodic/periodic.go
```diff
@@ -488,6 +489,32 @@ func (r *periodicRunner) Run() RunnerResults {
 	return result
 }
 
+type AccessLogger struct {
```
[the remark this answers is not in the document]
ldemailly: what it says ^

periodic/periodic.go
```diff
@@ -488,6 +489,32 @@ func (r *periodicRunner) Run() RunnerResults {
...
+	}
+	return &AccessLogger{file: f, format: format}, nil
+}
+
+func (a *AccessLogger) Report(thread int, time int64, latency float64) {
+	a.mu.Lock()
+	switch a.format {
+	case "influx":
+		// https://docs.influxdata.com/influxdb/v2.1/reference/syntax/line-protocol/
+		fmt.Fprintf(a.file, `latency,thread=%d value=%f %d`+"\n", thread, latency, time)
```
ldemailly: is the compiler smart enough to not keep redoing the \n concat? if not make that constant once outside or for this case just one string as there is no `"` to `\"`

ldemailly:
This is nice, thanks (see codeclimate and my minor comments) 

did you check for instance a self max qps with access log drops the max from say 50k qps to... ? (just to get an idea)

ldemailly:
gentle ping :) these are cool; it'd be nice to get it in

howardjohn:
I haven't forgotten this but have been busy - hoping to get some time soon.

howardjohn:
Did some (quick) test:

```
max qps file: 87k
max qps file (ramfs): 104k
max qps dev/null: 118k
max qps no file: 121k
```

So there is of course some impact, but not too bad. Updated the PR with the comments as well

fortio_main.go
```diff
@@ -391,6 +396,14 @@ func fortioLoad(justCurl bool, percList []float64) {
 		RunID:       *bincommon.RunIDFlag,
 		Offset:      *offsetFlag,
 	}
+	if *accessLogFileFlag != "" {
+		al, err := periodic.NewFileAccessLogger(*accessLogFileFlag, *accessLogFileFormat)
+		if err != nil {
+			_, _ = fmt.Fprintf(out, "Aborting because of %v\n", err)
```
ldemailly: this would typically be
```
			log.Fatalf("Unable to create access logger: %v", err)
```
(even if the stack trace can be confusing to the users, if we expect this to happen a lot then
```
log.Errf("Aborting because of access logger error: %v", err)
os.Exit(1)
```

fortio_main.go
```diff
@@ -174,6 +174,11 @@ var (
 	mirrorOriginFlag = flag.Bool("multi-mirror-origin", true, "Mirror the request url to the target for multi proxies (-M)")
 	multiSerialFlag  = flag.Bool("multi-serial-mode", false, "Multi server (-M) requests one at a time instead of parallel mode")
 	udpTimeoutFlag   = flag.Duration("udp-timeout", udprunner.UDPTimeOutDefaultValue, "Udp timeout")
+
+	accessLogFileFlag = flag.String("access-log-file", "",
+		"file to log all requests to. Maybe have performance impacts")
+	accessLogFileFormat = flag.String("access-log-format", "",
+		"format for access log. Support values: [json, influx]")
```
ldemailly: `Supported` or `Supports` no?

fortio_main.go
```diff
@@ -174,6 +174,11 @@ var (
 	mirrorOriginFlag = flag.Bool("multi-mirror-origin", true, "Mirror the request url to the target for multi proxies (-M)")
 	multiSerialFlag  = flag.Bool("multi-serial-mode", false, "Multi server (-M) requests one at a time instead of parallel mode")
 	udpTimeoutFlag   = flag.Duration("udp-timeout", udprunner.UDPTimeOutDefaultValue, "Udp timeout")
+
+	accessLogFileFlag = flag.String("access-log-file", "",
+		"file to log all requests to. Maybe have performance impacts")
+	accessLogFileFormat = flag.String("access-log-format", "",
```
ldemailly: shouldn't there be a default value, otherwise there is a dead end (someone sets only `-access-log-file` and it fails [silently even right now])

periodic/periodic.go
```diff
@@ -488,6 +489,41 @@ func (r *periodicRunner) Run() RunnerResults {
...
+	Report(thread int, time int64, latency float64)
+}
+
+// NewFileAccessLogger creates an AccessLogger that writes to the provided file in the provided format.
+func NewFileAccessLogger(file, format string) (AccessLogger, error) {
+	f, err := os.OpenFile(file, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
+	if err != nil {
+		return nil, err
+	}
+	return &fileAccessLogger{file: f, format: format}, nil
```
ldemailly: format isn't validated... so it'll fail silently/weirdly later?

I would maybe log both errors here and then in caller (main) you can just
```
  // error already logged
  os.Exit(1)
```
instead of the log.Fatalf... I mentioned there

ldemailly (changes requested):
Thanks for this, a few small things (I can do them if you're busy)

periodic/periodic.go
ldemailly: done

fortio_main.go
ldemailly: done

fortio_main.go
ldemailly: done

fortio_main.go
ldemailly: done, also added a more direct method on the runner options

periodic/periodic.go
```diff
@@ -493,6 +500,93 @@ func (r *periodicRunner) Run() RunnerResults {
 	return result
 }
 
+// AccessLoggerType is the possible formats of the access logger (ACCESS_JSON or ACCESS_INFLUX).
+type AccessLoggerType int
+
+const (
+	ACCESS_JSON AccessLoggerType = iota
```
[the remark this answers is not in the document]
ldemailly: k boss

periodic/periodic.go
```diff
@@ -493,6 +500,93 @@ func (r *periodicRunner) Run() RunnerResults {
...
+}
+
+// AccessLogger defines an interface to report a single request.
+type AccessLogger interface {
+	// Report logs a single request to a file.
+	Report(thread int, time int64, latency float64)
+	Info() string
+}
+
+func (r *RunnerOptions) AddAccessLogger(filePath, format string) error {
```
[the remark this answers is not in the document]
ldemailly: done

PR #513, opened 2022-03-02
Add message in result to indicate if uniform pattern is being used

wuhaoyujerry:
Show if the uniform traffic pattern is used in result. Example:
```
Jitter: false
Uniform: false
```


ldemailly:
thx! I did add it in json but forgot the stdout!

ldemailly:
come to think of it maybe we should have put this on a single/same line (as jitter)

maybe in follow up

PR #519, opened 2022-03-17
Fix the empty body content error

### LINKED ISSUE #520: Standard client, POST, after server close a connection: Unable to send POST request for ... http: ContentLength= with Body length 0 errors
Fixed in #519 (see repro there)

```
Unable to send POST request for http://localhost:8080/echo?close=true : Post "http://localhost:8080/echo?close=true": http: ContentLength=3 with Body length 0
```

wuhaoyujerry:
**Problem**
When we send multiple POST requests, we notice the error 
```
11:19:01 E http_client.go:379> [0] Unable to send POST request for `link` : Post `link`: http: ContentLength=259 with Body length 0
```
**Root Cause**
When the first request is sent, the request body is read to the end of file. So the following request body's starting index is already at end of file, which cause the body length to be 0.

**Solution**
Reset the request body's starting index before sending the request. No need to recreate the request everytime. Inspired by this [post](https://stackoverflow.com/a/47348886)

Fixes #520

### DIFF
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -371,7 +371,10 @@ func (c *Client) Fetch() (int, []byte, int) {
 		bodyBytes := []byte(body)
 		c.req.ContentLength = int64(len(bodyBytes))
 		c.req.Body = ioutil.NopCloser(bytes.NewReader(bodyBytes))
+	} else if c.body != "" {
+		c.req.Body = ioutil.NopCloser(bytes.NewReader([]byte(c.body)))
 	}
+
 	resp, err := c.client.Do(c.req)
 	if err != nil {
 		log.Errf("[%d] Unable to send %s request for %s : %v", c.id, c.req.Method, c.url, err)
--- a/fhttp/http_test.go
+++ b/fhttp/http_test.go
@@ -528,7 +528,7 @@ func TestPayloadWithEchoBack(t *testing.T) {
 	}
 }
 
-// Test Post request with std client and the socket close after answering
+// Test Post request with std client and the socket close after answering.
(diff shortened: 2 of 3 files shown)


ldemailly:
very nice find! to repro the issue:
```
go run . load -payload foo -stdclient http://localhost:8080/echo?close=true
# without the fix
$ go run -race . load -payload foo -stdclient http://localhost:8080/echo?close=true
Fortio dev running at 8 queries per second, 16->16 procs, for 5s: http://localhost:8080/echo?close=true
16:41:21 I httprunner.go:87> Starting http test for http://localhost:8080/echo?close=true with 4 threads at 8.0 qps and parallel warmup
Starting at 8 qps with 4 thread(s) [gomax 16] for 5s : 10 calls each (total 40)
16:41:21 E http_client.go:377> [3] Unable to send POST request for http://localhost:8080/echo?close=true : Post "http://localhost:8080/echo?close=true": http: ContentLength=3 with Body length 0
16:41:21 E http_client.go:377> [0] Unable to send POST request for http://localhost:8080/echo?close=true : Post "http://localhost:8080/echo?close=true": http: ContentLength=3 with Body length 0
16:41:21 E http_client.go:377> [2] Unable to send POST request for http://localhost:8080/echo?close=true : Post "http://localhost:8080/echo?close=true": http: ContentLength=3 with Body length 0
16:41:21 E http_client.go:377> [1] Unable to send POST request for http://localhost:8080/echo?close=true : Post "http://localhost:8080/echo?close=true": http: ContentLength=3 with Body length 0
...
```

ie do POST with std client and have the server side close the connection after the reply

ldemailly:
fantastic! will release asap

PR #523, opened 2022-03-19
allow a headers json array in json rest api payload ...

### LINKED ISSUE #518: Allow headers to be passed in jsonPath to resthandler
We need ability to pass headers intended for the target in the jsonPath. Fortio targets may need to be called with headers like `-H Content-Type:application/json`. The restHandler today does not accept headers in the jsonPath https://github.com/fortio/fortio/blob/master/ui/restHandler.go#L123-L144. 
However it does read the header on its incoming request and adds those to the target call. https://github.com/fortio/fortio/blob/master/ui/restHandler.go#L196

Add ability to pass headers intended for target in the jsonpath.

ldemailly:
- Fix #518: allow a `headers` json array in json rest api payload
- Expanded the README.md to cover the rest handler
- Also separate in fortio curl fast client the headers (to stderr) from the body (to stdout) previously everything was to stdout, unless `-curl-stdout-headers` is passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,7 +336,7 @@ You can set a default value for all these by passing `-echo-server-default-param
   * Download/sync from an Amazon S3 or Google Cloud compatible bucket listings [XML URLs](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html)
 
 * API to trigger and cancel runs from the running server (like the form ui but more directly and with `async=on` option)
-  * `/fortio/rest/run` starts a run; the arguments are either from the command line or from POSTed JSON; `jsonPath` can be provided to look for in a subset of the json object, for instance `jsonPath=metadata` allows to use the flagger webhook meta data for fortio run parameters (see [#493](https://github.com/fortio/fortio/pull/493)).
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -602,6 +605,186 @@ Response Body/Total Sizes : count 40 avg 12565.2 +/- 301.9 min 12319 max 13665 s
...
 ```
 
+
+### Remote triggered load test (server mode rest API)
+
+New since 1.18 the server has a `fortio/rest/run` endpoint similar to what the form UI submit in `fortio/` to start a run.
+  - plus `async` query arg or json value `"on"` will make the run asynchronous (returns just the runid of the run instead of waiting for the result)
+  - plus read all the run configuration from either query args or jsonPath POSTed info
+  - compatible with [flagger](https://github.com/fluxcd/flagger) and other webhooks
+  - New in 1.22: use `headers` json array to send headers (or multiple `&H=` query args)
```
tejaspbajaj, line 615: maybe add a comment, that if the same header is passed in both places, `headers` will overwrite &H

tejaspbajaj:
LGTM, thanks!

README.md
ldemailly: good point, for everything else the command line takes precedence so maybe I should flip the order for consistency

PR #529, opened 2022-03-25
Fast client support for https/TLS

### LINKED ISSUE #528: TLS/https support for fast client too (so we get socket count, higher performance etc)

ldemailly:
Support TLS for fast client too; that way we get socket count and higher performance (http 1.x only though, for h2(c) you'd want to specify to use the `-stdclient`)

fixes #528

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ set, restores pre 1.21 behavior
   -static-dir path
         Deprecated/unused path.
   -stdclient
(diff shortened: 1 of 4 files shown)


ldemailly:
Re: codecov/patch 'failure': 

CA certs etc are tested with release-tests, coverage hasn't actually changed for the diff, just moved the code to a shared function between standard and fast client

tejaspbajaj:
lgtm 👍

PR #544, opened 2022-04-14
no catchup mode (fixed/set maximum qps; skip requests when falling behind)

### LINKED ISSUE #543: max catch-up target qps / no catchup mode
When doing long running load at say 1000 qps and the system tested drops to only be able to do say 500 qps for 1 minute; fortio tries to catch up at max qps which can make the target system further fall over

let's add a "max-catch-up" rate maybe default to 150% and could be set to 100% if we never want to exceed max

ldemailly:
- [x] minor clarifications and improvements in periodic. document access logger for max qps too
- [x] actual no catchup/max catchup mode (Fix #543)
- [x] minor improvement on listen and http echo server log output

manually tested using various echo with % of delay above qps; yields stuff like
```
16:11:28 V periodic.go:675> T000 request took too long 0.2125 s, would sleep -13.968ms, skipping iter 171
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,16 +360,17 @@ There is also the GRPC health and ping servers, as well as the http->https redir
 
 ```Shell
 $ fortio server &
-14:11:05 I fortio_main.go:171> Not using dynamic flag watching (use -config to set watch directory)
-Fortio X.Y.Z tcp-echo server listening on [::]:8078
-Fortio X.Y.Z grpc 'ping' server listening on [::]:8079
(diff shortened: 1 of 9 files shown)


ldemailly:
To review: https://github.com/fortio/fortio/pull/544/files?diff=unified&w=1

That extra while loop is a bit odd as we could directly calculate how many iter we may need to skip but it helps for the boundary case (exit when reached num calls etc)

wuhaoyujerry (approved):
LGTM, in the future we should add more unit test cases

PR #556, opened 2022-04-28
Add more visibility to IP address resolution when using DNS

wuhaoyujerry:
Add logs at the end of the test to show the IP address each thread resolve to when using DNS and the usage count of each IP address.

Sample run result:
```
❯ go run . load https://fortio.org/
Fortio dev running at 8 queries per second, 10->10 procs, for 5s: https://fortio.org/
17:52:06 I httprunner.go:93> Starting http test for https://fortio.org/ with 4 threads at 8.0 qps and parallel warmup
Starting at 8 qps with 4 thread(s) [gomax 10] for 5s : 10 calls each (total 40)
17:52:11 I periodic.go:723> T001 ended after 5.041781917s : 10 calls. qps=1.9834257341202648
17:52:11 I periodic.go:723> T002 ended after 5.042107167s : 10 calls. qps=1.9832977897512427
17:52:11 I periodic.go:723> T003 ended after 5.043662s : 10 calls. qps=1.9826863893734352
17:52:11 I periodic.go:723> T000 ended after 5.044754833s : 10 calls. qps=1.9822568848312554
Ended after 5.04481125s : 40 calls. qps=7.9289
Sleep times : count 36 avg 0.50980665 +/- 0.008107 min 0.478828666 max 0.522668222 sum 18.3530394
Aggregated Function Time : count 40 avg 0.04440421 +/- 0.007715 min 0.032523166 max 0.075609875 sum 1.77616841
>= 0.0325232 <= 0.035 , 0.0337616 , 2.50, 1
> 0.035 <= 0.04 , 0.0375 , 27.50, 10
> 0.04 <= 0.045 , 0.0425 , 67.50, 16
> 0.045 <= 0.05 , 0.0475 , 80.00, 5
> 0.05 <= 0.06 , 0.055 , 97.50, 7
> 0.07 <= 0.0756099 , 0.0728049 , 100.00, 1
# target 50% 0.0428125
# target 75% 0.048
# target 90% 0.0557143
# target 99% 0.0733659
# target 99.9% 0.0753855
Error cases : no data
[0] Host https://fortio.org/ resolve to IP address: 104.21.74.52:443 
[1] Host https://fortio.org/ resolve to IP address: 104.21.74.52:443 
[2] Host https://fortio.org/ resolve to IP address: 104.21.74.52:443 
[3] Host https://fortio.org/ resolve to IP address: 104.21.74.52:443 
Uniform: false, Jitter: false
IP address 104.21.74.52:443 usage count: 4
Code 200 : 40 (100.0 %)
Response Header Sizes : count 40 avg 1130.65 +/- 5.764 min 1120 max 1145 sum 45226
Response Body/Total Sizes : count 40 avg 8536.525 +/- 5.886 min 8525 max 8550 sum 341461
All done 40 calls (plus 4 warmup) 44.404 ms avg, 7.9 qps
```

### DIFF
--- a/dflag/configmap/updater_test.go
+++ b/dflag/configmap/updater_test.go
@@ -124,8 +124,7 @@ type (
 
 // eventually tries a given Assert function 5 times over the period of time.
 func eventually(t *testing.T, duration time.Duration,
-	af assertFunc, expected interface{}, actual getter, msgFmt string, msgArgs ...interface{},
-) {
+	af assertFunc, expected interface{}, actual getter, msgFmt string, msgArgs ...interface{}) {
 	increment := duration / 5
 	for i := 0; i < 5; i++ {
 		time.Sleep(increment)
--- a/dflag/configmap/updater_test.go
+++ b/dflag/configmap/updater_test.go
@@ -124,7 +124,8 @@ type (
 
 // eventually tries a given Assert function 5 times over the period of time.
 func eventually(t *testing.T, duration time.Duration,
-	af assertFunc, expected interface{}, actual getter, msgFmt string, msgArgs ...interface{}) {
+	af assertFunc, expected interface{}, actual getter, msgFmt string, msgArgs ...interface{},
+) {
 	increment := duration / 5
 	for i := 0; i < 5; i++ {
 		time.Sleep(increment)
--- a/dflag/dynstringset.go
+++ b/dflag/dynstringset.go
@@ -78,7 +78,8 @@ func (d *DynStringSetValue) WithValidator(validator func(map[string]struct{}) er
 // WithNotifier adds a function that is called every time a new value is successfully set.
 // Each notifier is executed asynchronously in a new go-routine.
 func (d *DynStringSetValue) WithNotifier(notifier func(oldValue map[string]struct{},
-	newValue map[string]struct{})) *DynStringSetValue {
+	newValue map[string]struct{}),
+) *DynStringSetValue {
 	d.notifier = notifier
 	return d
 }
--- a/dflag/dynstringset.go
+++ b/dflag/dynstringset.go
@@ -78,8 +78,7 @@ func (d *DynStringSetValue) WithValidator(validator func(map[string]struct{}) er
 // WithNotifier adds a function that is called every time a new value is successfully set.
 // Each notifier is executed asynchronously in a new go-routine.
 func (d *DynStringSetValue) WithNotifier(notifier func(oldValue map[string]struct{},
-	newValue map[string]struct{}),
-) *DynStringSetValue {
+	newValue map[string]struct{})) *DynStringSetValue {
 	d.notifier = notifier
 	return d
 }
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -399,7 +399,6 @@ func (c *Client) Fetch() (int, []byte, int) {
 }
 
 func (c *Client) GetIPAddress() string {
-	//TODO implement me
 	return stdClientIP
 }
 
@@ -434,7 +433,10 @@ func NewStdClient(o *HTTPOptions) (*Client, error) {
 				addr = o.Resolve + addr[strings.LastIndex(addr, ":"):]
 			}
 			// TODO: Find out how many time this get called. Should be num of conn + error
-			conn, _ := net.Dial(network, addr)
+			conn, err := net.Dial(network, addr)
+			if err != nil {
+				log.Errf("Fail to dial addr %s, err msg: %s\n", addr, err)
+			}
 			stdClientIP = conn.RemoteAddr().String()
 
 			return (&net.Dialer{
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -310,7 +310,7 @@ type Client struct {
 	bodyContainsUUID     bool // if body contains the "{uuid}" pattern (lowercase)
 	logErrors            bool
 	id                   int
-	socketCount          *int
+	socketCount          int
 }
 
 // Close cleans up any resources used by NewStdClient.
@@ -327,7 +327,7 @@ func (c *Client) Close() int {
 	if c.transport != nil {
 		c.transport.CloseIdleConnections()
 	}
-	return *c.socketCount
+	return c.socketCount
 }
 
 // ChangeURL only for standard client, allows fetching a different URL.
@@ -424,7 +424,22 @@ func NewStdClient(o *HTTPOptions) (*Client, error) {
 		return nil, err
 	}
 
-	var socketCount int
+	client := Client{
+		url:                  o.URL,
+		path:                 req.URL.Path,
+		pathContainsUUID:     strings.Contains(req.URL.Path, uuidToken),
+		rawQuery:             req.URL.RawQuery,
+		rawQueryContainsUUID: strings.Contains(req.URL.RawQuery, uuidToken),
+		body:                 o.PayloadString(),
+		bodyContainsUUID:     strings.Contains(o.PayloadString(), uuidToken),
+		req:                  req,
+		client: &http.Client{
+			Timeout: o.HTTPReqTimeOut,
+		},
+		id:        o.ID,
+		logErrors: o.LogErrors,
+	}
+
 	tr := http.Transport{
 		MaxIdleConns:        o.NumConnections,
 		MaxIdleConnsPerHost: o.NumConnections,
@@ -441,35 +456,21 @@ func NewStdClient(o *HTTPOptions) (*Client, error) {
 			}).DialContext(ctx, network, addr)
 
 			req.RemoteAddr = conn.RemoteAddr().String()
-			socketCount++
+			client.socketCount++
 
 			return conn, err
 		},
 		TLSHandshakeTimeout: o.HTTPReqTimeOut,
 	}
+
+	client.client.Transport = &tr
+	client.transport = &tr
+
 	if o.https {
 		tr.TLSClientConfig, err = o.TLSOptions.TLSClientConfig()
 		if err != nil {
 			return nil, err
 		}
-	}
-	client := Client{
-		url:                  o.URL,
-		path:                 req.URL.Path,
-		pathContainsUUID:     strings.Contains(req.URL.Path, uuidToken),
-		rawQuery:             req.URL.RawQuery,
-		rawQueryContainsUUID: strings.Contains(req.URL.RawQuery, uuidToken),
-		body:                 o.PayloadString(),
-		bodyContainsUUID:     strings.Contains(o.PayloadString(), uuidToken),
-		req:                  req,
-		client: &http.Client{
-			Timeout:   o.HTTPReqTimeOut,
-			Transport: &tr,
-		},
-		transport:   &tr,
-		id:          o.ID,
-		logErrors:   o.LogErrors,
-		socketCount: &socketCount,
 	}
 	if !o.FollowRedirects {
 		// Lets us see the raw response instead of auto following redirects.
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -46,6 +46,7 @@ type Fetcher interface {
 	// Close() cleans up connections and state - must be paired with NewClient calls.
 	// returns how many sockets have been used (Fastclient only)
 	Close() int
+	GetIPAddress() string
 }
 
 const (
@@ -396,6 +397,11 @@ func (c *Client) Fetch() (int, []byte, int) {
 	return code, data, 0
 }
 
+func (c *Client) GetIPAddress() string {
+	//TODO implement me
+	panic("implement me")
+}
+
 // NewClient creates either a standard or fast client (depending on
 // the DisableFastClient flag).
 func NewClient(o *HTTPOptions) (Fetcher, error) {
@@ -426,6 +432,7 @@ func NewStdClient(o *HTTPOptions) (*Client, error) {
 			if o.Resolve != "" {
 				addr = o.Resolve + addr[strings.LastIndex(addr, ":"):]
 			}
+			// TODO: Find out how many time this get called. Should be num of conn + error
 			return (&net.Dialer{
 				Timeout: o.HTTPReqTimeOut,
 			}).DialContext(ctx, network, addr)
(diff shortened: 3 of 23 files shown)


fhttp/http_client.go
```diff
@@ -396,6 +398,10 @@ func (c *Client) Fetch() (int, []byte, int) {
 	return code, data, 0
 }
 
+func (c *Client) GetIPAddress() string {
```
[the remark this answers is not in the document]
wuhaoyujerry: added

fhttp/http_client.go
```diff
@@ -511,6 +524,10 @@ type FastClient struct {
 	tlsConfig    *tls.Config
 }
 
+func (c *FastClient) GetIPAddress() string {
```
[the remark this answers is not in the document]
wuhaoyujerry: added

dflag/configmap/updater_test.go
```diff
@@ -124,7 +124,8 @@ type (
 
 // eventually tries a given Assert function 5 times over the period of time.
 func eventually(t *testing.T, duration time.Duration,
-	af assertFunc, expected interface{}, actual getter, msgFmt string, msgArgs ...interface{}) {
+	af assertFunc, expected interface{}, actual getter, msgFmt string, msgArgs ...interface{},
+) {
```
ldemailly: how come you got formatting changes here? undo? (I don't think gofumpt complains about this file?)

dflag/dynstringset.go
```diff
@@ -78,7 +78,8 @@ func (d *DynStringSetValue) WithValidator(validator func(map[string]struct{}) er
 // WithNotifier adds a function that is called every time a new value is successfully set.
 // Each notifier is executed asynchronously in a new go-routine.
 func (d *DynStringSetValue) WithNotifier(notifier func(oldValue map[string]struct{},
-	newValue map[string]struct{})) *DynStringSetValue {
+	newValue map[string]struct{}),
+) *DynStringSetValue {
```
ldemailly: same

fhttp/http_client.go
```diff
@@ -396,6 +399,11 @@ func (c *Client) Fetch() (int, []byte, int) {
 	return code, data, 0
 }
 
+// GetIPAddress get the ip address that DNS resolves to when using stdClient.
+func (c *Client) GetIPAddress() string {
+	return <-stdClientIP
```
ldemailly: any chance this isn't yet there when this is called? would block?

fhttp/http_client.go
```diff
@@ -426,6 +434,16 @@ func NewStdClient(o *HTTPOptions) (*Client, error) {
 			if o.Resolve != "" {
 				addr = o.Resolve + addr[strings.LastIndex(addr, ":"):]
 			}
+			// TODO: Find out how many time this get called. Should be num of conn + error
```
ldemailly: do increment numsockets while at it?

fhttp/http_client.go
```diff
@@ -426,6 +434,16 @@ func NewStdClient(o *HTTPOptions) (*Client, error) {
...
+			conn, err := net.Dial(network, addr)
```
ldemailly: ? we can't connect 2x and throw away one the socket - that's not right at all

fhttp/http_client.go
```diff
@@ -426,6 +434,16 @@ func NewStdClient(o *HTTPOptions) (*Client, error) {
...
+			if err != nil {
+				log.Errf("Fail to dial addr %s, err msg: %s\n", addr, err)
+			}
+
+			go func() {
```
ldemailly: why?

fhttp/httprunner.go
```diff
@@ -37,6 +37,7 @@ type HTTPRunnerResults struct {
 	periodic.RunnerResults
 	client   Fetcher
 	RetCodes map[int]int64
+	IPCount  map[string]int64
```
ldemailly: 👍 but should be in shared results ideally? (even if we start with just http runner all the runners involving network should do that - but we can postpone that and move it later (there isn't a NetworkRunnerResults that would extend periodic' or we can just assume most of fortio is about network and put it there)

fhttp/httprunner.go
```diff
@@ -182,6 +184,11 @@ func RunHTTPTest(o *HTTPRunnerOptions) (*HTTPRunnerResults, error) {
 	// unused ones. We also must cleanup all the created clients.
 	keys := []int{}
 	for i := 0; i < numThreads; i++ {
+		// Get the report on the IP address each thread use to send traffic
+		ip := httpstate[i].client.GetIPAddress()
+		log.Infof("[%d] Host %s resolve to IP address: %s \n", i, o.URL, ip)
```
ldemailly: - no trailing space before \n and no \n anyway in log.*

- o.URL isn't the host so either put just
`"[%d] %s resolved to: %s"`

or get the host part

fhttp/httprunner.go
```diff
@@ -201,6 +208,9 @@ func RunHTTPTest(o *HTTPRunnerOptions) (*HTTPRunnerResults, error) {
 		_, _ = fmt.Fprintf(out, "Sockets used: %d (for perfect keepalive, would be %d)\n", total.SocketCount, r.Options().NumThreads)
 	}
 	_, _ = fmt.Fprintf(out, "Uniform: %t, Jitter: %t\n", total.Uniform, total.Jitter)
+	for k, v := range total.IPCount {
+		_, _ = fmt.Fprintf(out, "IP address %s usage count: %d\n", k, v)
```
ldemailly: maybe a header line, something like
```
IP addresses distribution:
1.1.1.1: 23
1.1.1.2: 7
...
```
and sort either by count or ip

ldemailly (changes requested):

fhttp/httprunner.go
wuhaoyujerry: updated

fhttp/httprunner.go
wuhaoyujerry: updated

dflag/configmap/updater_test.go
wuhaoyujerry: reverted. I run the `golangci-lint run --fix ` locally and this file got changed.

dflag/dynstringset.go
wuhaoyujerry: reverted

fhttp/http_client.go
wuhaoyujerry: removed

fhttp/http_client.go

fhttp/http_client.go
wuhaoyujerry: removed

fhttp/http_client.go
wuhaoyujerry: added implementation and a unit test. Now the stdclient will also show the socket count.

fhttp/httprunner.go
wuhaoyujerry: Added a todo for this

dflag/configmap/updater_test.go
ldemailly: ic, you have to make sure to use the version in the build image, which may not be same as on your mac

fhttp/http_client.go
ldemailly: awesome!

wuhaoyujerry:
@ldemailly Do you have any other questions about this PR? If not can we merge this? Thanks.

fhttp/http_client.go
```diff
@@ -308,6 +310,7 @@ type Client struct {
 	bodyContainsUUID     bool // if body contains the "{uuid}" pattern (lowercase)
 	logErrors            bool
 	id                   int
+	socketCount          *int
```
ldemailly: why is it a pointer?

ldemailly: I guess because the client is created after the transport but could either swap the order / have the initial client sooner or possibly have the pointer be only inside the transport

fhttp/http_utils.go
```diff
@@ -627,3 +627,11 @@ func (r *SyncReader) Read(p []byte) (n int, err error) {
 
 	return r.reader.Read(p)
 }
+
+func getIPUsageCount(ipCountMap map[string]int) (count int) {
```
ldemailly: this should only be needed in the test so should be in the test file?

fhttp/httprunner.go
```diff
@@ -193,14 +206,27 @@ func RunHTTPTest(o *HTTPRunnerOptions) (*HTTPRunnerResults, error) {
 		total.sizes.Transfer(httpstate[i].sizes)
 		total.headerSizes.Transfer(httpstate[i].headerSizes)
 	}
+
+	// Sort the ip address form largest to smallest based on its usage count
```
ldemailly, line 204: I think this is more complicated that it needs to be, see how `keys` and `total.RetCodes[k]` is handled (not the best name `keys` but still same principle)

fhttp/httprunner_test.go
```diff
@@ -121,11 +125,8 @@ func testHTTPNotLeaking(t *testing.T, opts *HTTPRunnerOptions) {
 	if ngAfter > ngBefore2+8 {
 		t.Errorf("Goroutines after test %d, expected it to stay near %d", ngAfter, ngBefore2)
 	}
-	if !opts.DisableFastClient {
```
ldemailly, line 124: 👍 nice that this passes for both now!

fhttp/httprunner_test.go
```diff
@@ -217,6 +217,14 @@ func TestClosingAndSocketCount(t *testing.T) {
 	}
 }
 
+func TestClosingAndSocketCountFastClient(t *testing.T) {
+	testClosingAndSocketCount(t, &HTTPRunnerOptions{})
+}
+
+func TestClosingAndSocketCountStdClient(t *testing.T) {
+	testClosingAndSocketCount(t, &HTTPRunnerOptions{HTTPOptions: HTTPOptions{DisableFastClient: true}})
```
ldemailly, line 225: nice thanks for the tests!

ui/restHandler.go
```diff
@@ -236,7 +236,8 @@ func RESTRunHandler(w http.ResponseWriter, r *http.Request) { // nolint: funlen
 
 // Run executes the run (can be called async or not, writer is nil for async mode).
 func Run(w http.ResponseWriter, r *http.Request, jd map[string]interface{},
-	runner, url string, ro periodic.RunnerOptions, httpopts *fhttp.HTTPOptions) {
+	runner, url string, ro periodic.RunnerOptions, httpopts *fhttp.HTTPOptions,
+) {
```
ldemailly: undo (and self review diffs)

fhttp/http_client.go
ldemailly: so yes let's use an int and just swap most of the init order, making the client first and then assinging the .client and .transport components

fhttp/httprunner.go
ldemailly: though to sort by value vs by key is indeed a bit more involved but not much:
```golang
    sort.Slice(keys, func(i, j int) bool { return counts[keys[i]] > counts[keys[j]] })
```

ui/restHandler.go
wuhaoyujerry: fixed

fhttp/httprunner.go
```diff
@@ -193,14 +200,27 @@ func RunHTTPTest(o *HTTPRunnerOptions) (*HTTPRunnerResults, error) {
...
+	ipList := make([]string, 0, len(total.IPCountMap))
+	for k, _ := range total.IPCountMap {
```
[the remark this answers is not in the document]
wuhaoyujerry: fixed

fhttp/httprunner.go
wuhaoyujerry: updated

fhttp/http_utils.go
wuhaoyujerry: moved to httprunner_test.go

fhttp/http_client.go

fhttp/http_client.go
```diff
@@ -426,35 +451,27 @@ func NewStdClient(o *HTTPOptions) (*Client, error) {
 			if o.Resolve != "" {
 				addr = o.Resolve + addr[strings.LastIndex(addr, ":"):]
 			}
-			return (&net.Dialer{
+			conn, err := (&net.Dialer{
 				Timeout: o.HTTPReqTimeOut,
 			}).DialContext(ctx, network, addr)
+
+			req.RemoteAddr = conn.RemoteAddr().String()
```
ldemailly: probably should only if err != nil or `conn` may be nil and it would crash?

and also maybe only if not already set to avoid keep calling RemoteAddr().String() {tradeoff being between getting the first address or last address picked} - but it should be rare so that part is fine (maybe a log when it does change from previous non empty value then if we keep reassigning)

fhttp/httprunner.go
```diff
@@ -193,14 +200,27 @@ func RunHTTPTest(o *HTTPRunnerOptions) (*HTTPRunnerResults, error) {
 		total.sizes.Transfer(httpstate[i].sizes)
 		total.headerSizes.Transfer(httpstate[i].headerSizes)
 	}
+
+	// Sort the ip address form largest to smallest based on its usage count
+	ipList := make([]string, 0, len(total.IPCountMap))
+	for k := range total.IPCountMap {
+		ipList = append(ipList, k)
```
ldemailly, line 207: it's somewhat minor but you should be able to get that list in the above loop; like `keys` - though maybe it's easier/as clear as you have it now; given we don't expect hundreds of ips anyway

ldemailly (commented):
nice! just 1 important and 1 optional change

ldemailly (approved):
thx

fhttp/http_test.go
```diff
@@ -691,6 +691,22 @@ func TestBadUrl(t *testing.T) {
...
+func TestBadURLStdClient(t *testing.T) {
+	opts := NewHTTPOptions("not a valid url")
+	cli, err := NewStdClient(opts)
+	if cli != nil || err == nil {
+		t.Errorf("config1: got a client %v despite bogus url %s", cli, opts.URL)
+		cli.Close()
+	}
+
+	opts.URL = "http://doesnotexist.fortio.org"
+	cli, err = NewStdClient(opts)
```
ldemailly: needs to be _ or check err

ldemailly:
```suggestion
	cli, _ = NewStdClient(opts)
```

fhttp/http_client.go
ldemailly: thx

PR #576, opened 2022-05-20
add configurable `-dns-method` to resolve dns

ldemailly:
with the default being cached round robin (`cached-rr`) to ensure as balanced as possible ip (if -c is multiple of the # of ips returned for the target

also usable `rnd` for a random address, `first` for the first one, `rr` for the previously implemented round robin

### DIFF
--- a/fnet/network.go
+++ b/fnet/network.go
@@ -27,7 +27,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"fortio.org/fortio/dflag"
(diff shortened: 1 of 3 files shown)


fnet/network.go
```diff
@@ -296,46 +319,84 @@ func Resolve(host string, port string) (*net.TCPAddr, error) {
 	return &net.TCPAddr{IP: addr.IP, Port: addr.Port}, nil
 }
 
+// Clear DNS cache for cached-rr resolution mode. For instance in case of error, to force re-resolving
```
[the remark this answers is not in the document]
ldemailly: fixed

fnet/network.go
```diff
@@ -63,19 +62,43 @@ var (
...
+	dnsRoundRobin uint32 = 0
 )
 
+func dnsValidator(inp string) error {
+	valid := map[string]bool{
+		"cached-rr": true,
+		"rnd":       true,
+		"rr":        true,
+		"first":     true,
+	}
```
bseenu, line 91: why map instead of slice here ?

bseenu (approved):

fnet/network.go
ldemailly: o(1) / easy to check without searching

PR #592, opened 2022-06-10
Updates /rest/status to provide runner status

wiardvanrij:
Implements information about what is running by re-using PeriodicRunner.RunnerOptions
I did add one basic test, but after fiddling around by mocking the results it became a bit messy. So I kept it simple for now.

Status with runid that does not exist:

```
 curl -d 'runid=200' http://localhost:8080/fortio/rest/status
{"error": runid not found}
```

Status without runid with no runs:

```
curl http://localhost:8080/fortio/rest/status
{"Results":null}
```

Status with multiple runs:

```
curl -k http://localhost:8080/fortio/rest/status
{"results":[{"runid":1,"status":{"qps":10,"numthreads":1,"labels":"Fortio","exactly":0,"jitter":false,"uniform":false,"runid":1,"nocatchup":false}},{"runid":2,"status":{"qps":10,"numthreads":8,"labels":"Fortio","exactly":0,"jitter":false,"uniform":false,"runid":2,"nocatchup":false}}]}
```

Single run:

```
curl -k http://localhost:8080/fortio/rest/status
{"results":[{"runid":1,"status":{"qps":10,"numthreads":1,"labels":"Fortio","exactly":0,"jitter":false,"uniform":false,"runid":1,"nocatchup":false}}]}
```

Single run, explicit defined:

```
curl -d 'runid=1' http://localhost:8080/fortio/rest/status
{"results":[{"runid":1,"status":{"qps":10,"numthreads":1,"labels":"Fortio","exactly":0,"jitter":false,"uniform":false,"runid":1,"nocatchup":false}}]}
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,9 +11,9 @@ require (
 )
 
 require (
+	github.com/google/go-cmp v0.5.7 // indirect
 	golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a // indirect
 	golang.org/x/text v0.3.7 // indirect
 	google.golang.org/genproto v0.0.0-20220525015930-6ca3db687a9d // indirect
 	google.golang.org/protobuf v1.28.0 // indirect
-	gopkg.in/yaml.v3 v3.0.0 // indirect
 )
--- a/periodic/periodic.go
+++ b/periodic/periodic.go
@@ -106,49 +106,49 @@ func NewAborter() *Aborter {
 // RunnerOptions are the parameters to the PeriodicRunner.
 type RunnerOptions struct {
 	// Type of run (to be copied into results)
-	RunType string
+	RunType string `json:"-"`
 	// Array of objects to run in each thread (use MakeRunners() to clone the same one)
-	Runners []Runnable
+	Runners []Runnable `json:"-"`
 	// At which (target) rate to run the Runners across NumThreads.
-	QPS float64
+	QPS float64 `json:"qps"`
 	// How long to run the test for. Unless Exactly is specified.
-	Duration time.Duration
+	Duration time.Duration `json:"-"`
 	// Note that this actually maps to gorountines and not actual threads
 	// but threads seems like a more familiar name to use for non go users
 	// and in a benchmarking context
-	NumThreads  int
-	Percentiles []float64
-	Resolution  float64
+	NumThreads  int       `json:"numthreads"`
+	Percentiles []float64 `json:"-"`
+	Resolution  float64   `json:"-"`
 	// Where to write the textual version of the results, defaults to stdout
-	Out io.Writer
+	Out io.Writer `json:"-"`
 	// Extra data to be copied back to the results (to be saved/JSON serialized)
-	Labels string
+	Labels string `json:"labels"`
 	// Aborter to interrupt a run. Will be created if not set/left nil. Or you
 	// can pass your own. It is very important this is a pointer and not a field
 	// as RunnerOptions themselves get copied while the channel and lock must
 	// stay unique (per run).
-	Stop *Aborter
+	Stop *Aborter `json:"-"`
 	// Mode where an exact number of iterations is requested. Default (0) is
 	// to not use that mode. If specified Duration is not used.
-	Exactly int64
+	Exactly int64 `json:"exactly"`
 	// When multiple clients are used to generate requests, they tend to send
 	// requests very close to one another, causing a thundering herd problem
 	// Enabling jitter (+/-10%) allows these requests to be de-synchronized
 	// When enabled, it is only effective in the '-qps' mode.
-	Jitter bool
+	Jitter bool `json:"jitter"`
 	// When multiple clients are used to generate requests, they tend to send
 	// requests very close to one another, causing a thundering herd problem
 	// Enabling uniform causes the requests between connections to be uniformly staggered.
 	// When enabled, it is only effective in the '-qps' mode.
-	Uniform bool
+	Uniform bool `json:"uniform"`
 	// Optional run id; used by the server to identify runs.
-	RunID int64
+	RunID int64 `json:"runid"`
 	// Optional Offset Duration; to offset the histogram function duration
-	Offset time.Duration
+	Offset time.Duration `json:"-"`
 	// Optional AccessLogger to log every request made. See AddAccessLogger.
-	AccessLogger AccessLogger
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -11,9 +11,9 @@ require (
 )
 
 require (
+	github.com/google/go-cmp v0.5.7 // indirect
 	golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a // indirect
 	golang.org/x/text v0.3.7 // indirect
 	google.golang.org/genproto v0.0.0-20220525015930-6ca3db687a9d // indirect
 	google.golang.org/protobuf v1.28.0 // indirect
-	gopkg.in/yaml.v3 v3.0.0 // indirect
```
ldemailly, line 18: this needs to stay because otherwise we have a security issue (unless other deps got updated that makes it unneeded now)

go.mod
```diff
@@ -11,9 +11,9 @@ require (
 )
 
 require (
+	github.com/google/go-cmp v0.5.7 // indirect
```
ldemailly, line 14: I am typically against new dependencies unless they somehow are super hyper useful and necessary, ie needs a justification (good part though is at least it's not a package from a rando :)  and it's indirect so... )

is this coming from  httptest.NewRecorder() maybe ? (I can do `go mod why` I guess)

periodic/periodic.go
```diff
@@ -106,49 +106,49 @@ func NewAborter() *Aborter {
 // RunnerOptions are the parameters to the PeriodicRunner.
 type RunnerOptions struct {
 	// Type of run (to be copied into results)
-	RunType string
+	RunType string `json:"-"`
```
ldemailly, line 109: that doesn't seem right? 

also why I know it's nice to spell out the json name, the defaults do work and I tend to prefer less code, less typos over more (ie not change this part at all unless it improves something?)

ui/restHandler.go
```diff
@@ -313,9 +325,49 @@ func Run(w http.ResponseWriter, r *http.Request, jd map[string]interface{},
...
+	statusResult, err := StatusByRunID(runid)
+
 	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusServiceUnavailable)
-	w.Write([]byte("{\"error\":\"status not yet implemented\"}"))
+
+	if err != nil {
+		w.WriteHeader(http.StatusNotFound)
+		w.Write([]byte(fmt.Sprintf("{\"error\": %s}", err.Error())))
+	} else {
```
ldemailly, line 337: return in error if, no else (never have an else is a good rule of thumb, unless both if and else are 1 line each or so)

ui/restHandler_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+const emptyResult = `{"results":null}`
+
+func TestEmptyResult(t *testing.T) {
```
ldemailly, line 12: it's going to be hard to do meaningful tests without actually starting some servers (no we won't start having mock objects for everything, let's not go down that path :) )

you can see e2e tests in fhttp/

ui/ has had no go test in part for the bad reason of not killing the coverage % (!) and in part because most of that is excercised somewhat in Webtest.sh

I think we can move the resHandler to fhttp if we're going to properly test it - which we should - eventually

ldemailly (commented):
Biggest issue is that this is not thread safe and worse in a way (not your fault, it's the current design) the results are not meant to be accessed while the run is running (they haven't been aggregated yet)

Here are some minor issues with the code but we need to talk a bit more about design

**_edit_**: looking at the translated output in https://github.com/fortio/fortio/pull/592#issuecomment-1152750942

I realize it's not the status but the input/options; so I guess that can work but maybe not as useful - but can give context/and info so might be good - will look again later

ldemailly:
for the examples if you can do `| jq` and put the output using 3 ` json it'll help

e.g.
```json
{
  "results": [
    {
      "runid": 1,
      "status": {
        "qps": 10,
        "numthreads": 1,
        "labels": "Fortio",
        "exactly": 0,
        "jitter": false,
        "uniform": false,
        "runid": 1,
        "nocatchup": false
      }
    },
    {
      "runid": 2,
      "status": {
        "qps": 10,
        "numthreads": 8,
        "labels": "Fortio",
        "exactly": 0,
        "jitter": false,
        "uniform": false,
        "runid": 2,
        "nocatchup": false
      }
    }
  ]
}
```

go.mod
wiardvanrij: Hmm, I did a `go mod tidy` and it got removed. So I sort of wonder 🤔

periodic/periodic.go
wiardvanrij: From my understanding, it must have declarations if we Marshall it like we do now. Otherwise at least we get slapped by the linter.

The other thing is that I've removed a few from the output. Vars that are not set or chans (which can't be marshalled). I'm fine of leaving the default names, but it should have some Jason declaration.

ldemailly: I have a set of *Result objects too for emitting the json that fortio saves for instance - which don't have stuff that is not needed and can be a base object (but then again this isn't the result :) ) 

Maybe the shared Options are enough

but why removing RunType for instance anyway?

ldemailly:
poke

periodic/periodic.go
wiardvanrij: From my understanding (but I will check again) it is net set. Only after it's done.

wiardvanrij:
Not forgotten, but haven't got the time to go over it again.

ui/restHandler_test.go
ldemailly: I'm doing that disentangling in https://github.com/fortio/fortio/pull/597 as it's really my mess to clean up

wiardvanrij:
Closing this for now, in favour of all the other work you've already done. I'll revisit it when I got more time to see where I can help!

periodic/periodic.go
ldemailly: You're right and it turns into an interesting challenge in #611

ui/restHandler_test.go
ldemailly: #611 has tests ptal

ldemailly:
Thanks, it was useful as prototype and does highlight some issues with the runner options updates

PR #593, opened 2022-06-12
dflags generics - added WithInputMutator and WithValueMutator too

### LINKED ISSUE #475: Add a (simpler) way to modify the dflag values
One can use WithSyncNotifiers() and if careful to unregister/register to avoid a loop; you can process the new value but it would be simpler to have an api that returns the value instead

(use case: trim white spaces for instance on a dynstring or other transformations)

could also simplify how loglevel is done

ldemailly:
Use generics to tremendously shrink/improve the dflag/ code (left the tests unchanged or added to them to test indeed the implementation is compatible; they could be cleaned as well

added WithInputMutator() and WithMutator() too
WithSyncNotifier() now available to all generic types

Fixes #475


wuhaoyujerry (approved):
Wow, the generic really  reduced a lot of duplicate code. Thx!

PR #594, opened 2022-06-16
Max connection reuse

### LINKED ISSUE #533: option to client side force not reusing a connection more than X times
option to client side force not reusing a connection more than X times
default would remain to reuse connection as much as possible (until an error or server close) but we can add a
`-max-connection-reuse n`

One issue is maybe we would want different numbers per thread so they don't do it all at the same time (could be offset by % threadid though that is yet another option (so many/too many)

combined with #532 and #531 we would simulate more easily "new clients"

wuhaoyujerry:
### Feature
Add flag `max-connection-reuse` to force create new connection after a thread reached the max connection reuse. 
### Description
The flag value is a string `min:max` to define the range of max connection reuse thresholds that thread will randomly pick a value within that range. This range will avoid all thread create new connection at the same time.
### Example
```
❯ go run . load -loglevel verbose  -max-connection-reuse 10:20 -c 2 -n 50  https://fortio.org/
...
Starting at 8 qps with 2 thread(s) [gomax 10] : exactly 50, 25 calls each (total 50 + 0)
...
15:53:26 V http_client.go:727> [0] Thread reach the threshold for max connection reuse of 17, force create new connection
15:53:27 V http_client.go:727> [1] Thread reach the threshold for max connection reuse of 12, force create new connection
...
All done 50 calls (plus 0 warmup) 337.438 ms avg, 5.9 qps
```

Close #533


bincommon/commonflags.go
```diff
@@ -106,6 +107,8 @@ var (
 		"http(s) runner warmup done in parallel instead of sequentially. When set, restores pre 1.21 behavior")
 	curlHeadersStdout = flag.Bool("curl-stdout-headers", false,
 		"Restore pre 1.22 behavior where http headers of the fast client are output to stdout in curl mode. now stderr by default.")
+	maxConnectionReuseFlag = flag.String("max-connection-reuse", "",
+		"Range of max number of connections to reuse for each thread, default to unlimited.")
```
ldemailly: maybe an example `(e.g 10:30 means ...)` would help otherwise one is left wondering why it's a string etc

also you can use ` around range or something like
```
"Range `pattern` for the max...
```
that way when you do -h it will show `pattern` instead of `string` for the type

ldemailly: I am wondering how it will be set in the UI too (need to add to both the ui form and the rest handler)

bincommon/commonflags.go
```diff
@@ -170,6 +173,32 @@ func TLSInsecure() bool {
 	return TLSInsecure
 }
 
+func parseMaxConnectionReuse(maxConnectionReuseFlag string) []int {
```
ldemailly: Might be better as a dflag and WithValidator/WithNotifier
(without doing Exit(1) so it can be used in the UI as well)
to populate the pair... not sure, I have to think about it/we can talk

fhttp/http_client.go
```diff
@@ -176,14 +176,15 @@ type HTTPOptions struct {
...
-	SequentialWarmup bool          // whether to do http(s) runs warmup sequentially or in parallel (new default is //)
+	hostOverride       string
+	HTTPReqTimeOut     time.Duration // timeout value for http request
+	UserCredentials    string        // user credentials for authorization
+	ContentType        string        // indicates request body type, implies POST instead of GET
+	Payload            []byte        // body for http request, implies POST if not empty.
+	LogErrors          bool          // whether to log non 2xx code as they occur or not
+	ID                 int           // id to use for logging (thread id when used as a runner)
+	SequentialWarmup   bool          // whether to do http(s) runs warmup sequentially or in parallel (new default is //)
+	MaxConnectionReuse []int         // range of max number of connection to reuse for each thread.
```
ldemailly: [2]int if it's always a range? or an actual pair struct? ([2]int seems simple enough)
though I guess you use empty range for infinite? [comment as much either way even if leaving as a slice:
empty slice = no limit; otherwise must be a slice of 2..?)

fhttp/http_client.go
```diff
@@ -510,29 +511,31 @@ func Fetch(httpOptions *HTTPOptions) (int, []byte) {
...
+	keepAlive          bool
+	parseHeaders       bool // don't bother in http/1.0
+	halfClose          bool // allow/do half close when keepAlive is false
+	reqTimeout         time.Duration
+	uuidMarkers        [][]byte
+	logErrors          bool
+	id                 int
+	https              bool
+	tlsConfig          *tls.Config
+	maxConnectionReuse int
```
ldemailly: sometimes I add a // above to avoid the full whitespace change
or maybe in this case call it  `maxConnReuse` though that's also still longer than `parseHeaders`
maybe `connReuse` as it's not quite the max but range anyway?

fhttp/http_client.go
```diff
@@ -585,11 +588,18 @@ func NewFastClient(o *HTTPOptions) (Fetcher, error) {
 		log.Errf("[%d] Bad url %q : %v", o.ID, urlString, err)
 		return nil, err
 	}
+
+	// Randomly assign a max connection reuse threshold to this thread.
+	var maxConnectionReuse int
+	if len(o.MaxConnectionReuse) == 2 {
+		maxConnectionReuse = o.MaxConnectionReuse[0] + rand.Intn(o.MaxConnectionReuse[1]-o.MaxConnectionReuse[0]+1)
```
ldemailly: if max==min maybe not call rand?

and if they are different, shouldn't we reroll each time we close ? (I wonder how we communicate which thread does what / if that's needed)

fhttp/http_client.go
```diff
@@ -510,29 +511,31 @@ func Fetch(httpOptions *HTTPOptions) (int, []byte) {
...
+	parseHeaders       bool // don't bother in http/1.0
+	halfClose          bool // allow/do half close when keepAlive is false
+	reqTimeout         time.Duration
+	uuidMarkers        [][]byte
+	logErrors          bool
+	id                 int
+	https              bool
+	tlsConfig          *tls.Config
+	maxConnectionReuse int
+	currentConnection  int
```
ldemailly: `reuseCount` ?

fhttp/http_client.go
```diff
@@ -709,13 +719,24 @@ func (c *FastClient) Fetch() (int, []byte, int) {
...
+	reuse := conn != nil
+	if c.maxConnectionReuse != 0 && c.currentConnection >= c.maxConnectionReuse {
+		reuse = false
+		c.currentConnection = 0
+		log.LogVf("[%d] Thread reach the threshold for max connection reuse of %d, force create new connection",
+			c.id, c.maxConnectionReuse)
+	}
+
 	if !reuse {
 		conn = c.connect()
```
ldemailly, line 778: I'd maybe put the  `c.reuseCount=0` here and not worry about only incrementing it if the feature is on; it's a variable on the thread and just an increment and could be useful info to see

fhttp/http_client.go
```diff
@@ -709,13 +719,24 @@ func (c *FastClient) Fetch() (int, []byte, int) {
 	c.headerLen = 0
 	// Connect or reuse existing socket:
 	conn := c.socket
-	reuse := (conn != nil)
+
+	reuse := conn != nil
+	if c.maxConnectionReuse != 0 && c.currentConnection >= c.maxConnectionReuse {
```
ldemailly: I think this may make the minimum reuse to be 2 instead of 1?

ie I should be able to say close each time, I think if I set [1, 1] wouldn't it close/redo every other?

can add a test with the socket count that if I do n: 10 and reuse [1 1] I should get exactly 10 sockets and if I put [2 2] I should get exactly 5 ... [and another boundary check is [9 9] should be... 2 sockets [10 10] should be 1

fhttp/httprunner_test.go
```diff
@@ -471,6 +471,38 @@ func TestAbortOn(t *testing.T) {
 	}
 }
 
+func TestMaxConnectionReuse(t *testing.T) {
```
ldemailly: see the other tests that have an array of input and expected output so you can test as mentioned above something like

n=10 (instead of relying on duration)

and ranges from [1,1] to [10,10]

fhttp/http_client.go

fhttp/http_client.go
wuhaoyujerry: fixed

fhttp/http_client.go
wuhaoyujerry: fixed

fhttp/http_client.go
wuhaoyujerry: fixed

fhttp/http_client.go
wuhaoyujerry: fixed

fhttp/http_client.go
wuhaoyujerry: fixed

fhttp/httprunner_test.go
wuhaoyujerry: updated

bincommon/commonflags.go
wuhaoyujerry: I added this flag to UI and rest handler

bincommon/commonflags.go
```diff
@@ -170,6 +176,18 @@ func TLSInsecure() bool {
 	return TLSInsecure
 }
 
+// ConnectionReuseRangeValidator returns a validator function that checks if the connection reuse range is valid
+// and set in httpOpts.
+func ConnectionReuseRangeValidator(httpOpts *fhttp.HTTPOptions) func(string) error {
+	return func(value string) error {
+		if err := httpOpts.ValidateConnectionReuseRange(value); err != nil {
```
ldemailly: so this probably works given the code in dflag, though I remember some corner case where the validator may be called but we end up setting a different value; but technically this is not a `ValidateX` it's `ValidateAndSetX` (it mutates the httpOpts and it should in theory be split between a validator and a syncnotifier)

All this being said maybe it's fine as is (well except for the name of the function in options) and we should document that it's ok to use a Validator as a Setter to another object (maybe)

bincommon/commonflags.go
```diff
@@ -170,6 +176,18 @@ func TLSInsecure() bool {
...
+			return fmt.Errorf("fail to parse connection reuse range, err: %v", err)
```
ldemailly: %w when wrapping errors - or maybe just return err as is?

bincommon/commonflags_test.go
```diff
@@ -0,0 +1,27 @@
...
+	err := ConnectionReuseRange.Set(value)
+	if err != nil {
+		t.Errorf("Failed to set connection reuse range, err: %v", err)
+	}
+
+	if ConnectionReuseRange.Get() != value {
+		t.Errorf("Connection reuse range doesn't match the set value, expected %v, got %v",
+			ConnectionReuseRange.Get(), value)
+	}
+
```
ldemailly: you're not checking the side effect here (ie that you got 10 and 20 in httpOpts

bincommon/commonflags_test.go
```diff
@@ -0,0 +1,27 @@
+package bincommon
```
ldemailly: fun thing is so far there was no test for that part (no direct test, only through webtest.sh which does test some of the command line)

adding any test here drops the coverage number as a result :)

you could either just move this into the fhttp ones or I guess we can add more tests later and live with the hit in coverage 

this test doesn't really test much outside of the real test about ValidateConnectionReuseRange which you have in TestValidateConnectionReuse so I would maybe drop this file now (not a great thing I realize)

fhttp/http_client.go
```diff
@@ -277,6 +278,39 @@ func (h *HTTPOptions) AddAndValidateExtraHeader(hdr string) error {
...
+	reuseRangeString := strings.Split(inp, ":")
+	var reuseRangeInt []int
+
+	if len(reuseRangeString) > 2 {
+		return fmt.Errorf("more than two integers were provided in the connection reuse range")
+	}
+
+	for _, input := range reuseRangeString {
+		if val, err := strconv.Atoi(input); err != nil {
+			return fmt.Errorf("invalid value for connection reuse range, err: %v", err)
```
ldemailly: same, %w


btw can you see if one of the golang lint ci linter should be able to flag that? pretty sure I saw it mentioned by some linter before (outside of the linter that will say don't use fmt.Errorf at all that is, which I think is't right)

fhttp/http_client.go
```diff
@@ -277,6 +278,39 @@ func (h *HTTPOptions) AddAndValidateExtraHeader(hdr string) error {
...
+		} else {
+			reuseRangeInt = append(reuseRangeInt, val)
+		}
+	}
+
+	if len(reuseRangeInt) == 1 {
+		h.ConnReuseRange = [2]int{reuseRangeInt[0], reuseRangeInt[0]}
+	} else {
+		if reuseRangeInt[0] < reuseRangeInt[1] {
```
ldemailly, line 304: nice to be user friendly

fhttp/httprunner_test.go
```diff
@@ -471,6 +472,98 @@ func TestAbortOn(t *testing.T) {
...
+
+		if res.SocketCount != (int)(expectedSocketReuse) {
+			t.Errorf("Expecting %f socket to be used, got %d", expectedSocketReuse, res.SocketCount)
+		}
+	}
+
+	// Test when connection reuse range min != max
+	for i := 0; i < 5; i++ {
+		opts.ConnReuseRange = [2]int{5, 9}
+		expectedSocketReuse := math.Ceil(float64(opts.Exactly) / float64(opts.ConnReuseRange[0]))
```
ldemailly: a bit confused how come that part doesn't fail... it sees to always use the lower bound (5), and it shouldn't?

or I guess for 10 if you have 5 or more but not 10 it's same as 5... ic you're getting exactly 2?
might be clearer to put `expectedSocketReuse:= 2` with a comment, outside the loop then?

fhttp/httprunner_test.go
```diff
@@ -471,6 +472,98 @@ func TestAbortOn(t *testing.T) {
...
+	err = httpOpts.ValidateConnectionReuseRange("10")
+	if err != nil {
+		t.Errorf("Expect no error when single value is privided, got err: %v.", err)
+	}
+
+	err = httpOpts.ValidateConnectionReuseRange("20:10")
+	if err != nil {
+		t.Errorf("Expect no error when two values are privided, got err: %v.", err)
+	}
+
```
ldemailly, line 562: check the ints we got as a result

go.mod
```diff
@@ -16,5 +16,4 @@ require (
 	golang.org/x/text v0.3.7 // indirect
 	google.golang.org/genproto v0.0.0-20220525015930-6ca3db687a9d // indirect
 	google.golang.org/protobuf v1.28.0 // indirect
-	gopkg.in/yaml.v3 v3.0.1 // indirect
```
ldemailly: I think this is needed but I can check (was a security fix)... maybe just undo go.mod and go.sum

ldemailly (commented):
thx!

bincommon/commonflags.go
wuhaoyujerry: I updated the function name

bincommon/commonflags.go
wuhaoyujerry: Updated

bincommon/commonflags_test.go
wuhaoyujerry: removed the file

bincommon/commonflags_test.go
wuhaoyujerry: removed the file

go.mod
wuhaoyujerry: This was removed when I ran `go mod tidy`, reverted

fhttp/httprunner_test.go
wuhaoyujerry: added

fhttp/httprunner_test.go
wuhaoyujerry: updated

fhttp/http_client.go
wuhaoyujerry: I tried to find a linter that does that but didn't find anything useful. `go vet` can only flag when you don't provide enough argument to a format string but doesn't do type validation. Lemme know if you can find one.

ldemailly: google ftw

https://github.com/polyfloyd/go-errorlint#examples

fhttp/httprunner_test.go
```diff
@@ -471,6 +472,107 @@ func TestAbortOn(t *testing.T) {
...
+		}
+
+		if res.SocketCount != (int)(expectedSocketReuse) {
+			t.Errorf("Expecting %f socket to be used, got %d", expectedSocketReuse, res.SocketCount)
+		}
+	}
+
+	// Test when connection reuse range min != max.
+	// The actual socket count should always be 2 as the connection reuse range varies between 5 and 9.
+	expectedSocketReuse := 2.0
```
ldemailly: should be := 2 without .0 and no cast to int in comparaison

fhttp/httprunner_test.go
```diff
@@ -471,6 +472,107 @@ func TestAbortOn(t *testing.T) {
...
+	for i := 0; i < 5; i++ {
+		opts.ConnReuseRange = [2]int{5, 9}
```
ldemailly: move out of the loop as it's not changing?

the loop is here just because of rand right?

fhttp/httprunner_test.go
```diff
@@ -471,6 +472,107 @@ func TestAbortOn(t *testing.T) {
...
+		res, err := RunHTTPTest(&opts)
+		if err != nil {
+			t.Error(err)
+		}
+
+		if res.SocketCount != (int)(expectedSocketReuse) {
```
ldemailly:
```suggestion
		if res.SocketCount != expectedSocketReuse {
```

fhttp/httprunner_test.go
ldemailly:
```suggestion
	expectedSocketReuse := 2
```

fhttp/httprunner_test.go
ldemailly:
```suggestion
	opts.ConnReuseRange = [2]int{5, 9}
	// Check a few times that despite the range and random 2-9 we still always get 2 connections
	for i := 0; i < 5; i++ {
```

ldemailly: btw if you put 5,10 does it fail? sometimes? always? most times?

ldemailly: see suggestion below

fhttp/httprunner_test.go
wuhaoyujerry: updated

fhttp/http_client.go
wuhaoyujerry: thx! I added the errorlint

fhttp/httprunner_test.go
wuhaoyujerry: Yeah it fails sometime 
``` 
httprunner_test.go:521: Expecting 2 socket to be used, got 1
```

.golangci.yml
```diff
@@ -342,6 +342,7 @@ linters-settings:
 linters:
   enable:
     - unused
+    - errorlint
```
ldemailly, line 345: thanks! and you tried it, if you put back %v it complains?

ldemailly (approved):

.golangci.yml
wuhaoyujerry: yes
```
fhttp/http_client.go:295:75: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)
                        return fmt.Errorf("invalid value for connection reuse range, err: %v", err)
```

PR #596, opened 2022-06-21
cleanup in generics implementation, clarity on DynValueTypes

ldemailly:
Many thanks to Gophers slack #generics for feedback and @rogpeppe for the suggestions

### DIFF
--- a/dflag/dyngeneric.go
+++ b/dflag/dyngeneric.go
@@ -72,11 +72,13 @@ func ValidateDynSliceMinElements[T any](count int) func([]T) error {
 	}
 }
 
+// DynValueTypes are the types currently supported by Parse[T] and thus by Dyn[T].
+// DynJSON is special.
 type DynValueTypes interface {
-	bool | time.Duration | float64 | int64 | string | []string | Set[string] | interface{}
+	bool | time.Duration | float64 | int64 | string | []string | Set[string]
 }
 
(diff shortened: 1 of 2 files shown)


dflag/dyngeneric.go
```diff
@@ -132,46 +134,34 @@ func (d *DynValue[T]) Get() T {
...
-// ComaStringToSlice converts a coma separated string to a slice.
-func ComaStringToSlice(input string) []string {
+// CommaStringToSlice converts a coma separated string to a slice.
+func CommaStringToSlice(input string) []string {
 	// originally the heavy handed csv.NewReader(strings.NewReader(input)).Read()
 	return strings.Split(input, ",")
 }
 
 // Parse converts from string to our supported types (it's the missing generics strconv.Parse[T]).
 func Parse[T any](input string) (val T, err error) {
```
rogpeppe: If you unexport this, you can write an exported wrapper that has the correct type constraint and calls the internal version, so you get the type safety you'd like while still enabling `DynValue` to call it.

```
func Parse[T DynValueTypes](input string) (T, error) {
	return parse[T](input)
}
```

ldemailly: yeah I just was just doing that but had to tweak the tests/add one - done now, thanks!

wuhaoyujerry (approved):

PR #597, opened 2022-06-28
moved rest api and tsv to rapi/ package, restore linters and fix lints

ldemailly:


ldemailly:
hopefully we have good test coverage because that's quite a lot of changes

wuhaoyujerry (approved):
wow this is a lot of work, thx so much

PR #598, opened 2022-06-29
add echo back to the connection reuse flag and revert readme 

wuhaoyujerry:
- Add info about connection reuse flag in the test runner log
- revert the readme to the previous fortio version since the new version is not ready for tag yet
- remove host from GetIPAddress method in http_client

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,9 +160,10 @@ connections, if empty, use https:// prefix for standard internet/system CAs
   -config path
         Config directory path to watch for changes of dynamic flags (empty for
 no watch)
-  -connection-reuse-range min:max
-        Range min:max for the max number of connections to reuse for each thread, 
-default to unlimited. e.g. 10:30 means randomly choose a max connection reuse threshold between 10 and 30 requests.
+  -connection-reuse min:max
+        Range min:max for the max number of connections to reuse for each
+thread, default to unlimited. e.g. 10:30 means randomly choose a max connection
+reuse threshold between 10 and 30 requests.
   -content-type string
         Sets http content type. Setting this value switches the request method
 from GET to POST.
--- a/README.md
+++ b/README.md
@@ -51,13 +51,13 @@ You can install from source:
 The [releases](https://github.com/fortio/fortio/releases) page has binaries for many OS/architecture combinations (see assets).
 
 ```shell
-curl -L https://github.com/fortio/fortio/releases/download/v1.34.0/fortio-linux_amd64-1.34.0.tgz \
+curl -L https://github.com/fortio/fortio/releases/download/v1.33.0/fortio-linux_amd64-1.33.0.tgz \
  | sudo tar -C / -xvzpf -
 # or the debian package
-wget https://github.com/fortio/fortio/releases/download/v1.34.0/fortio_1.34.0_amd64.deb
-dpkg -i fortio_1.34.0_amd64.deb
+wget https://github.com/fortio/fortio/releases/download/v1.33.0/fortio_1.33.0_amd64.deb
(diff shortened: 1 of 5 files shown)


fhttp/httprunner.go
```diff
@@ -187,10 +193,11 @@ func RunHTTPTest(o *HTTPRunnerOptions) (*HTTPRunnerResults, error) {
 	for i := 0; i < numThreads; i++ {
 		// Get the report on the IP address each thread use to send traffic
 		host, ip := httpstate[i].client.GetIPAddress()
-		log.Infof("[%d] %s resolved to %s\n", i, host, ip)
+		currentSocketUsed := httpstate[i].client.Close()
+		log.Infof("[%d] %d socket used, %s resolved to %s\n", i, currentSocketUsed, host, ip)
```
ldemailly: trying to remember why we have the `host` there, can it be somehow different for 1 connection than another? (I don't think so but there must be some reason we put it in the interface for GetIPAddress()?)

fhttp/httprunner.go
```diff
@@ -90,7 +90,13 @@ func RunHTTPTest(o *HTTPRunnerOptions) (*HTTPRunnerResults, error) {
 	if o.SequentialWarmup {
 		warmupMode = "sequential"
 	}
-	log.Infof("Starting http test for %s with %d threads at %.1f qps and %s warmup", o.URL, o.NumThreads, o.QPS, warmupMode)
+
+	connReuseMsg := fmt.Sprintf("with connection reuse range between %d and %d", o.ConnReuseRange[0], o.ConnReuseRange[1])
+	if o.ConnReuseRange == [2]int{0, 0} {
+		connReuseMsg = ""
```
ldemailly: pedantic as this section isn't high performance but in general, avoid doing work to then throw away, so:
```suggestion
	connReuseMsg := ""
	if o.ConnReuseRange != [2]int{0, 0} {
		connReuseMsg = fmt.Sprintf(" with connection reuse [%d, %d]", o.ConnReuseRange[0], o.ConnReuseRange[1])
```
note I also added a space so we can remove the extra one below

fhttp/httprunner.go
```diff
@@ -90,7 +90,13 @@ func RunHTTPTest(o *HTTPRunnerOptions) (*HTTPRunnerResults, error) {
...
+	}
+	log.Infof("Starting http test for %s with %d threads at %.1f qps and %s warmup %s",
```
ldemailly:
```suggestion
	log.Infof("Starting http test for %s with %d threads at %.1f qps and %s warmup%s",
```

fhttp/httprunner.go
wuhaoyujerry: updated

ldemailly (approved):
awesome, thx!

ldemailly:
made minor updates to input (shorter flag) and output (text)

README.md
```diff
@@ -160,9 +160,10 @@ connections, if empty, use https:// prefix for standard internet/system CAs
   -config path
         Config directory path to watch for changes of dynamic flags (empty for
 no watch)
-  -connection-reuse-range min:max
-        Range min:max for the max number of connections to reuse for each thread, 
-default to unlimited. e.g. 10:30 means randomly choose a max connection reuse threshold between 10 and 30 requests.
+  -connection-reuse min:max
+        Range min:max for the max number of connections to reuse for each
+thread, default to unlimited. e.g. 10:30 means randomly choose a max connection
```
ldemailly, line 165: note that the line break comes from running `release/updateFlags.sh`

PR #610, opened 2022-07-21
Add flag to resolve dns name on each new connection

### LINKED ISSUE #532: option to re-resolve name->ip on new connection/connection error (fast client)
Right now we do the name resolution once per thread, if the connection breaks, we use the same IP (which is faster but doesn't cover dns based traffic shifting)

Add `-resolve-each-connect` (name tbd) or such to re resolve on connect()

wuhaoyujerry:
### Feature
Add a flag `no-resolve` to force the client to re-resolve the DNS name for every new connection. 

### Description
The flag is a boolean with default value set to true. For each thread, the number of sockets and the ip address the DNS name resolved to will be recorded. 

### Example
```
❯ go run . load  -connection-reuse 10:20 -no-resolve=false  -c 1 -n 50 https://xxx.net
...
22:51:43 D http_client.go:745> Hostname xxx resolve to ip 10.144.27.56:443
22:51:45 D http_client.go:745> Hostname xxx resolve to ip 10.144.2.238:443
22:51:46 D http_client.go:745> Hostname xxx resolve to ip 10.144.40.90:443
22:51:49 D http_client.go:745> Hostname xxx resolve to ip 10.144.27.56:443
...
# Socket and IP used for each connection:
[0]   5 socket used, resolved to [10.144.40.90:443, 10.144.27.56:443 (2), 10.144.2.238:443 (2)]
...
```

### Progress
- [x] Add the flag and implement the logic
- [x] Add a struct in stat to store the ip address and the number of times it get called
- [x] Add unit test 

Fix #532


bincommon/commonflags.go
```diff
@@ -112,6 +112,8 @@ var (
 		"Range `min:max` for the max number of connections to reuse for each thread, default to unlimited. "+
 			"e.g. 10:30 means randomly choose a max connection reuse threshold between 10 and 30 requests.").
 		WithValidator(ConnectionReuseRangeValidator(&httpOpts))
+	// NoResolveConnFlag is false if we want to resolve the DNS name for each new connection.
+	NoResolveConnFlag = flag.Bool("no-resolve", true, "Re-resolve the DNS name for each connection")
```
ldemailly: you meant false for default right, as we want to turn the new behavior (to re resolve) on

also name would be (despite stutter) `NoReResolveFlag` ? (`-no-reresolve`)

"Keep the initial DNS resolution and don't re-resolve when making new connections (because of error or reuse limit reached)"

fnet/network_test.go
```diff
@@ -537,7 +537,7 @@ func TestDNSMethods(t *testing.T) {
 	if err != nil {
 		t.Errorf("unexpected error in rnd mode for resolve of google: %v", err)
 	}
-	fnet.FlagResolveMethod.Set("rr")
+	err = fnet.FlagResolveMethod.Set("rr")
```
ldemailly, line 540: thx for that fix!

stats/stats.go
```diff
@@ -539,3 +539,31 @@ func RoundToDigits(v float64, digits int) float64 {
 func Round(v float64) float64 {
 	return RoundToDigits(v, 4)
 }
+
+// Map is a type that stores the key value pair which the key is string and value is int.
+type Map struct {
```
ldemailly: Map is quite a vague/generic name here (even in stats.)

maybe it's `Occurrences`  or some such

fhttp/http_client.go
```diff
@@ -1050,3 +1067,11 @@ func generateReuseThreshold(min int, max int) int {
 
 	return min + rand.Intn(max-min+1) // nolint: gosec // we want fast not crypto
 }
+
+func resolve(hostname string, port string, resolve string) (*net.TCPAddr, error) {
```
ldemailly: rename in this function the resolve arg as `override` and/or comment that this will basically resolve either first or 3rd arg if not empty?

fhttp/http_client.go
```diff
@@ -732,6 +738,17 @@ func (c *FastClient) connect() net.Conn {
 	c.socketCount++
 	var socket net.Conn
 	var err error
+
+	// Resolve the DNS name when making new connections.
+	if !c.noResolveEachConn {
+		c.dest, err = resolve(c.hostname, c.port, c.resolve)
```
ldemailly: doesn't that mean it's going to do it one extra time even for 1 connection? (once in client setup and once again at connect)

bincommon/commonflags.go
wuhaoyujerry: updated

stats/stats.go
wuhaoyujerry: updated

fhttp/http_client.go

fhttp/http_client.go
wuhaoyujerry: good point, updated

fhttp/http_client.go
```diff
@@ -635,6 +642,10 @@ func NewFastClient(o *HTTPOptions) (Fetcher, error) {
 		return nil, err
 	}
 
+	if o.NoResolveEachConn && o.Resolve != "" {
+		log.Warnf("Both `-resolve` and `-resolve-on-err` flags are defined, will use same ip address on new conn")
```
ldemailly: flag name update needed

fhttp/http_client.go
```diff
@@ -732,6 +740,17 @@ func (c *FastClient) connect() net.Conn {
 	c.socketCount++
 	var socket net.Conn
 	var err error
+
+	// Resolve the DNS name when making new connections.
+	if c.socketCount > 1 && !c.noResolveEachConn {
+		c.dest, err = resolve(c.hostname, c.port, c.resolve, c.ipAddrUsage)
+		log.Debugf("Hostname %v resolve to ip %v", c.hostname, c.dest)
```
ldemailly: nit: put the [c.id] also in new debug logs

ldemailly (commented):
nice, still need to aggregate the dns occurences right?

fhttp/http_client.go
wuhaoyujerry: Updated, thx for catching this

fhttp/http_client.go
wuhaoyujerry: added

stats/stats_test.go
```diff
@@ -29,6 +29,11 @@ import (
 	"fortio.org/fortio/log"
 )
 
+const (
+	ipOne = "127.0.0.1"
+	ipTwo = "127.0.0.2"
```
ldemailly, line 34: We had talked about but I forgot to set it up until I saw this - I just added:
```
% host ltest.fortio.org
ltest.fortio.org has address 127.0.0.2
ltest.fortio.org has address 127.0.0.3
ltest.fortio.org has address 127.0.0.1
```

fhttp/http_client.go
```diff
@@ -45,9 +46,9 @@ type Fetcher interface {
 	Fetch() (int, []byte, int)
 	// Close() cleans up connections and state - must be paired with NewClient calls.
 	// returns how many sockets have been used (Fastclient only)
```
ldemailly: update the doc comments

fhttp/http_client.go
```diff
@@ -45,9 +46,9 @@ type Fetcher interface {
...
-	Close() int
+	Close()
 	// GetIPAddress() returns the last ip address used by this client connection.
```
ldemailly: here too

fhttp/http_client.go
```diff
@@ -185,6 +186,8 @@ type HTTPOptions struct {
 	ID               int           // id to use for logging (thread id when used as a runner)
 	SequentialWarmup bool          // whether to do http(s) runs warmup sequentially or in parallel (new default is //)
 	ConnReuseRange   [2]int        // range of max number of connection to reuse for each thread.
+	// re-resolve the DNS name when the connection breaks.
```
ldemailly: when false ...

fhttp/http_client.go
```diff
@@ -443,8 +445,10 @@ func (c *Client) Fetch() (int, []byte, int) {
 }
 
 // GetIPAddress get the ip address that DNS resolves to when using stdClient.
-func (c *Client) GetIPAddress() string {
-	return c.req.RemoteAddr
+func (c *Client) GetIPAddress() (*stats.Occurrence, int) {
+	occurrence := stats.NewOccurrence()
+	occurrence.Record(c.req.RemoteAddr)
```
ldemailly: maybe fine for this iteration but couldn't we record in std client as well?

stats/stats.go
```diff
@@ -539,3 +539,46 @@ func RoundToDigits(v float64, digits int) float64 {
...
+
+// Record records a key value pair.
+func (m *Occurrence) Record(key string) {
+	m.ipUsage[key]++
+}
+
+// PrintAndAggregate print the ip usage and aggregate to the total count map.
+func (m *Occurrence) PrintAndAggregate(ipCountMap map[string]int) string {
+	var sb strings.Builder
+	_, _ = fmt.Fprintf(&sb, "[")
```
ldemailly: WriteString("[") 
(the f ending functions are meant for formatting something)

stats/stats.go
```diff
@@ -539,3 +539,46 @@ func RoundToDigits(v float64, digits int) float64 {
...
+	return o
+}
+
+// Record records a key value pair.
+func (m *Occurrence) Record(key string) {
+	m.ipUsage[key]++
+}
+
+// PrintAndAggregate print the ip usage and aggregate to the total count map.
+func (m *Occurrence) PrintAndAggregate(ipCountMap map[string]int) string {
```
ldemailly, line 562: shouldn't it take another Occurence rather than raw map? (not sure, just wondering) - also maybe uint and/or int64 and/or uint64

stats/stats.go
```diff
@@ -539,3 +539,46 @@ func RoundToDigits(v float64, digits int) float64 {
...
+	var sb strings.Builder
+	_, _ = fmt.Fprintf(&sb, "[")
+
+	size := len(m.ipUsage)
+	count := 0
+
+	for k, v := range m.ipUsage {
+		ipCountMap[k] += v
+		_, _ = fmt.Fprintf(&sb, "%s", k)
```
ldemailly: same this is just an append

might be more direct to use a stringbuffer too

ldemailly (commented):
thanks! (quick/partial review from my phone)

stats/stats.go
ldemailly: given we already had IPCountMap I guess it's fine, even though somewhat awkward  api to combine a String() and a Transfer()/aggregate style into one, but it does save a loop

stats/stats.go
```diff
@@ -539,3 +539,46 @@ func RoundToDigits(v float64, digits int) float64 {
...
+			_, _ = fmt.Fprintf(&sb, " (%d)", v)
+		}
+
+		count++
+		if count != size {
+			_, _ = fmt.Fprintf(&sb, ", ")
+		}
+	}
+	_, _ = fmt.Fprintf(&sb, "]")
+
```
ldemailly:
```suggestion
	sb.WriteString("[")

	size := len(m.ipUsage)
	count := 0

	for k, v := range m.ipUsage {
		ipCountMap[k] += v
		sb.WriteString(k)
		if size == 1 {
			break
		}
		sb.WriteString(fmt.Sprintf(" (%d)", v))
		count++
		if count != size {
			sb.WriteString(", ")
		}
	}
	sb.WriteString("]")
```

stats/stats_test.go
ldemailly: note for anyone looking at this later

on linux a fortio server does bind on all the 127.0.0.x addresses but not on macos

fhttp/http_client.go
wuhaoyujerry: Updated

ldemailly (approved):
thanks a lot for this!

PR #611, opened 2022-07-27
new jrpc package, and tests/rework for rest handler

ldemailly:
Adding a new `jrpc` json rpc client and server library

Use it in tests and in rest handler

Also changed the returned async objects so they can be used and deserialized

Plus improvements

- [x] change list api
- [x] save file async ID and url
- [x] golang 1.18.5
- [x] don't run the build/tests as root 

fixes #612 

[linter and build changes makes this harder to review than it should, not counting the new features, tricky synchronization etc -- apologies]

PR #613, opened 2022-07-29
new json rpc package

ldemailly:


ldemailly:
reworked/improved in #611

PR #621, opened 2022-09-06
support passing headers and timeout in jrpc calls

### LINKED ISSUE #619: jgrpc needs to let caller add optional headers
eg to call for instance apis that require a token:
```
req.Header = http.Header{
		"Authorization": {"Token token=" + pagerduty_token},
		"Accept":        {"application/vnd.pagerduty+json;version=2"},
}
```
cc @wiardvanrij

ldemailly:
fixes #619 

sort of breaking api change, added some "jrpc.xxxxURL" backward compatible string url call points

### DIFF
--- a/jrpc/jrpcClient.go
+++ b/jrpc/jrpcClient.go
@@ -62,7 +62,7 @@ type FetchError struct {
 	Bytes []byte
 }
 
-// Destination is the URL and optional additional headers
+// Destination is the URL and optional additional headers.
 type Destination struct {
 	URL     string
 	Headers *http.Header
--- a/jrpc/jrpcClient.go
+++ b/jrpc/jrpcClient.go
@@ -26,13 +26,25 @@ import (
 	"io"
 	"net/http"
 	"time"
+
+	"fortio.org/fortio/version"
+)
+
+const (
+	UserAgentHeader = "User-Agent"
 )
 
 // Default timeout for Call.
 var timeout = 60 * time.Second
 
+// UserAgent is the User-Agent header used by client calls (also used in fhttp/).
+var UserAgent = "fortio.org/fortio-" + version.Short()
+
 // SetCallTimeout changes the timeout for further Call calls, returns
-// the previous value (default in 60s).
+// the previous value (default in 60s). Value is used when a timeout
+// isn't passed in the options. Note this is not thread safe,
+// use Destination.Timeout for changing values outside of main/single
+// thread.
 func SetCallTimeout(t time.Duration) time.Duration {
 	previous := timeout
 	timeout = t
@@ -50,6 +62,13 @@ type FetchError struct {
 	Bytes []byte
 }
 
+// Destination is the URL and optional additional headers
+type Destination struct {
+	URL     string
+	Headers *http.Header
+	Timeout time.Duration
+}
+
 func (fe *FetchError) Error() string {
 	return fmt.Sprintf("%s, code %d: %v (raw reply: %s)", fe.Message, fe.Code, fe.Err, DebugSummary(fe.Bytes, 256))
 }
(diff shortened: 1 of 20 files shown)


jrpc/jrpcClient.go
```diff
@@ -26,13 +26,27 @@ import (
...
 
 // Default timeout for Call.
 var timeout = 60 * time.Second
 
+// UserAgent is the User-Agent header used by client calls (also used in fhttp/).
+var UserAgent = "fortio.org/fortio-" + version.Short()
+
 // SetCallTimeout changes the timeout for further Call calls, returns
-// the previous value (default in 60s).
+// the previous value (default in 60s). Value is used when a timeout
```
wiardvanrij, line 46:
```suggestion
// The previous value (default in 60s). Value is used when a timeout
```

jrpc/jrpcClient.go
```diff
@@ -16,8 +16,8 @@
 // using generics to serialize/deserialize any type.
 package jrpc // import "fortio.org/fortio/jrpc"
 
-// This package is a true self contained library, doesn't rely on our logger nor other packages in fortio/.
-// Client side and common code.
+// This package is a true self contained library, doesn't rely on our logger nor other packages
```
wiardvanrij:
```suggestion
// This package is a true self contained library, that doesn't rely on our logger nor other packages
```

jrpc/jrpcClient.go
```diff
@@ -110,24 +143,43 @@ func CallWithPayload[Q any](url string, bytes []byte) (*Q, error) {
...
 		return -1, res, err
 	}
+	if dest.Headers != nil {
+		req.Header = dest.Headers.Clone()
+	}
+	if len(jsonPayload) > 0 {
+		SetHeaderIfMissing(req.Header, "Content-Type", "application/json; charset=utf-8")
+	}
+	SetHeaderIfMissing(req.Header, "Accept", "application/json")
+	SetHeaderIfMissing(req.Header, UserAgentHeader, UserAgent)
```
wiardvanrij, line 182: I guess it's fine but making `UserAgentHeader` via a constant feels more 'complicated', whats the benefit/use-case?

jrpc/jrpcClient.go
```diff
@@ -110,24 +143,43 @@ func CallWithPayload[Q any](url string, bytes []byte) (*Q, error) {
...
 	if err != nil {
 		return -1, res, err
 	}
+	if dest.Headers != nil {
+		req.Header = dest.Headers.Clone()
+	}
+	if len(jsonPayload) > 0 {
+		SetHeaderIfMissing(req.Header, "Content-Type", "application/json; charset=utf-8")
+	}
+	SetHeaderIfMissing(req.Header, "Accept", "application/json")
```
wiardvanrij, line 181: Maybe set this to something like `text/html,application/xhtml+xml` and only set this if the `jsonPayload` > 0 like on line 178? Not entirely sure what happens if we default to this. I'm absolutely not sure what the 'standard' is on this though. Just thinking out loud.

wiardvanrij (approved):
Basically only a few nits/questions

jrpc/jrpcClient.go
ldemailly: this is continuation of same sentence "returns the previous value" so it would not work with "returns The previous value:

jrpc/jrpcClient.go
ldemailly: it's used 3+ time which is typically my threshold but yeah clarity doesn't improve (a bit like go's http.MethodGet == "GET")

jrpc/jrpcClient.go
ldemailly: why text/html when this expects json? Accept is about the expected reply, the content type for payload > 0 is for what we send

if for some reason the caller needs a different result (not json) then they can set the Accept header in Destination

jrpc/jrpcClient.go
ldemailly: thanks a lot for having looked and the comments though! hopefully it means the API seems ok to you to use

wiardvanrij: I don't see weird/odd things. I have to figure out a use for it (do let me know) to battle test it.

ldemailly: the pager duty call we're making for instance - I'll make a few MRs once this is released and we can iterate if it's not great

PR #627, opened 2022-09-19
remove some nolint for contextcheck

kkHAIKE:
u can try the latest version contextcheck ..
`contextcheck -pkgprefix=fortio ./...`


ldemailly:
thanks a lot! I'm doing that in #630 which gets the new version of your linter!

PR #649, opened 2022-11-06
support User-Agent change and deletion from -H, simplify UI for headers

### LINKED ISSUE #648: can't set user-agent from command line
full headers are reset in the web UI (form and rapi) but can't be changed in command line:

```
fortio curl -loglevel debug -H User-Agent:test123 localhost:8080/debug
[...]
13:57:32 D http_client.go:739> [0] Created client:
127.0.0.1:8080
GET /debug HTTP/1.1
Host: localhost:8080
User-Agent: fortio.org/fortio-1.38.2
User-Agent: test123
[...]
GET /debug HTTP/1.1

headers:

Host: localhost:8080
User-Agent: fortio.org/fortio-1.38.2,test123
```

Let's make user-agent a Set and not Add

ldemailly:
support User-Agent change and deletion from -H, simplify UI for headers to be extra headers only. 

fixes #648

### DIFF
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -279,7 +279,7 @@ func (h *HTTPOptions) AddAndValidateExtraHeader(hdr string) error {
 	switch strings.ToLower(key) {
 	case "host":
 		log.LogVf("Will be setting special Host header to %s", value)
-		h.hostOverride = value
+		h.hostOverride = strings.TrimSpace(value) // This one needs to be trimmed
 	case "user-agent":
 		if value == "" {
 			log.Infof("Deleting default User-Agent: header.")
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -273,17 +273,19 @@ func (h *HTTPOptions) AddAndValidateExtraHeader(hdr string) error {
 		return fmt.Errorf("invalid extra header '%s', expecting Key: Value", hdr)
 	}
 	key := strings.TrimSpace(s[0])
-	value := strings.TrimSpace(s[1])
+	// No TrimSpace for the value, so we can set empty "" vs just whitespace " " which
+	// will get trimmed later but treated differently: not emitted vs emitted empty for User-Agent.
+	value := s[1]
 	switch strings.ToLower(key) {
 	case "host":
 		log.LogVf("Will be setting special Host header to %s", value)
 		h.hostOverride = value
 	case "user-agent":
 		if value == "" {
-			log.Infof("User-Agent: header removed.")
+			log.Infof("Deleting default User-Agent: header.")
 			h.extraHeaders.Del(key)
 		} else {
-			log.LogVf("User-Agent being Set to %s", value)
+			log.Infof("User-Agent being Set to %q", value)
 			h.extraHeaders.Set(key, value)
 		}
 	default:
@@ -341,6 +343,10 @@ func newHTTPRequest(o *HTTPOptions) (*http.Request, error) {
 	req.Header = o.GenerateHeaders()
 	if o.hostOverride != "" {
 		req.Host = o.hostOverride
+	}
+	// Another workaround for std client otherwise trying to set a default User-Agent
+	if _, ok := req.Header["User-Agent"]; !ok {
+		req.Header.Set("User-Agent", "")
 	}
 	if !log.LogDebug() {
 		return req, nil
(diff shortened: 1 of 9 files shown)


fhttp/http_client.go
```diff
@@ -274,10 +274,19 @@ func (h *HTTPOptions) AddAndValidateExtraHeader(hdr string) error {
...
 	key := strings.TrimSpace(s[0])
 	value := strings.TrimSpace(s[1])
-	if strings.EqualFold(key, "host") {
+	switch strings.ToLower(key) {
+	case "host":
 		log.LogVf("Will be setting special Host header to %s", value)
 		h.hostOverride = value
-	} else {
+	case "user-agent":
+		if value == "" {
```
wiardvanrij, line 284: I think we should delete this part and allow setting an empty user-agent. Apparently there is some use-case where people _dont_ want to set a header but requests get proxied. That proxy adds a user agent if it's not set. Therefore people want to set an empty user-agent header to either not make up one nor get a default one later via the tool or the proxy

https://github.com/zalando/skipper/issues/1994

ldemailly: So my goal here is to provide a way for people to use `fortio curl` and delete the (otherwise present by default) `User-Agent:` header (without adding yet another flag if at all possible as we do have too many already)

My first change only had the Set part but then I noticed on the wire sending an empty
```
User-Agent:
```
which waste bytes and looks ugly - also on the server side can't be (in go at least) distinguished from absent

I see that use case to be a lot more practical/useful than if a proxy has some behavior that can be argued as bugged (why add a UA header instead of a proxy-version one?)

ldemailly: the go bug is interesting though (and that's what it is imo, a go bug, to decide to add UA when missing in a reverse proxy) - thanks for pointing that out as it applies to `fortio server -M` as well as https://github.com/fortio/proxy

ldemailly: made some adjustment (yet also didn't yet offer the option to send an empty User-Agent)

ldemailly: Ok with latest update you can use white space to send an empty User-Agent: and empty for delete/no such header. (and avoids getting `User-Agent: Go-http-client/1.1` either)

ldemailly: @wiardvanrij ready for final review now - I think - tia

wiardvanrij (approved):
Implementation looks good, funny how something small ends up into this :) 

The only nit I have, is that technically if you would enforce a linter on it (comments), it's going to say something about the style. It should start with a capital and end with a dot `.`. But I'll leave it up to you if you want to start making that the default/standard to make the comments more uniform.

fhttp/http_test.go
```diff
@@ -1120,6 +1129,11 @@ func TestDebugHandlerSortedHeaders(t *testing.T) {
 	o.AddAndValidateExtraHeader("CCC: ccc")
 	o.AddAndValidateExtraHeader("ZZZ: zzz")
 	o.AddAndValidateExtraHeader("AAA: aaa")
+	// test that headers usually Add (list) but stay in order of being set
```
wiardvanrij, line 1132:
```suggestion
	// Test that headers usually Add (list) but stay in order of being set.
```

wiardvanrij: See my general comment, this is as an example

ldemailly: So the linter (annoyingly tbh) does flag the period missing on comments that are part of the "javadoc" ie in front of functions and types - ie visible comments. Probably not in test files though as that seems overkill. I will check.

ldemailly:
While I totally agree it is useful for documentation to have proper complete sentences, I dont think making complete sentences of local comments in test files or middle of code is useful. even cosmetically I think it detracts from their purpose (and more importantly puts incentive to not comment)

fhttp/http_test.go
ldemailly: confirmed the linter for that still works:

removed period in one of the file:
```
fhttp/http_client.go:41:75: Comment should end in a period (godot)
// Fetcher is the Url content fetcher that the different client implements
                                                                          ^
```

fhttp/http_client.go
ldemailly: > funny how something small ends up into this :)

indeed it was quite a long winded road - thanks for pointing the issues out

PR #654, opened 2022-11-12
added test for #652, stepped on #653 while testing the test, added test for that too

### LINKED ISSUE #652: fortio server should error on extra non flag arguments
If you have a typo:
```
fortio server logelevel verbose
```
ie missing `-` fortio should error out instead of merrily starting

### LINKED ISSUE #653: fortio curl exit status is 0 despite error
```
fortio curl err; echo $?
13:07:50 W http_client.go:152> Assuming http:// on missing scheme for 'err'
13:07:50 E network.go:386> Unable to lookup 'err' : lookup err: no such host
0
```

ldemailly:
- [x] added test for #652, stepped on #653 while testing the test, added test for that too
- [x] fixed #653
- [x] fix #652


fortio_main.go
```diff
@@ -229,34 +240,31 @@ func main() {
 	case "load":
 		fortioLoad(*curlFlag, percList)
 	case "redirect":
-		isServer = true
+		isServer = serverArgCheck()
```
ldemailly, line 243: it's a bit hacky, basically to avoid 2 lines
```golang
isServer = true
serverArgCheck()
```
in 6 places (I suppose an alternative also ugly is make isServer global and set it in the function)

asha-oss (approved):
good job!

PR #671, opened 2022-12-18
add support for setting grpc metadata

### LINKED ISSUE #663: Support setting grpc metadata
Now fortio does not support setting grpc [metadata](https://grpc.io/docs/what-is-grpc/core-concepts/#metadata) like '-H' flag in Http Load Generation and it would be better if it did.
I'd be happy to do the work if needed.

ktalg:
Closes https://github.com/fortio/fortio/issues/663


README.md
```diff
@@ -200,6 +200,8 @@ option unset.
   -grpc-port port
         grpc server port. Can be in the form of host:port, ip:port or port or
 /unix/domain/path or "disabled" to not start the grpc server. (default "8079")
+  -grpc-metadata value
+        Metadata that will be added to the grpc request. e.g. -grpc-metadata "key: value".
```
ldemailly: Can we just use the existing -H ?

fgrpc/pingsrv.go
```diff
@@ -110,16 +111,24 @@ func PingServerTCP(port, cert, key, healthServiceName string, maxConcurrentStrea
...
 		return -1, err // error already logged
 	}
 	msg := &PingMessage{Payload: payload, DelayNanos: delay.Nanoseconds()}
 	cli := NewPingServerClient(conn)
+	mkCtx := func() context.Context {
+		return context.Background()
+	}
+	if md.Len() != 0 {
+		mkCtx = func() context.Context {
+			return metadata.NewOutgoingContext(context.Background(), md)
```
ldemailly: this (as the linters complain - btw do run `make lint` (or `make local-lint` if you have same version of go and `golangci-lint` on your local machine) is a bit odd

given all this is in same context/loop - do like you did above and just have 1 ctx variable?

README.md
ktalg: Sorry for the late reply, I recently got infected with covid-19. grpc metadata is equal to http header most of the time but there are some nuances (like not allowing to set some reserved headers etc.), reusing -H might cause some confusion? I don't know, it's up to you to decide whether to change it or not, I'm ok with both.

fgrpc/pingsrv.go
ktalg: I downloaded `golangci-lint` and did some checking and fixed it.

README.md
ktalg: I think if we intend to reuse `-H`, we can add more clearer flag doc.

ldemailly: sure we can update the doc to say Headers or GRPC meta data but we already have way too many flags so I'm relunctant to add more that are same (assuming, correct me as I don't know, that grpc meta data has the same key:value pair string:string semantic) ?

ktalg: done.

bincommon/commonflags.go
```diff
@@ -119,7 +119,9 @@ var (
 
 // SharedMain is the common part of main from fortio_main and fcurl.
 func SharedMain(usage func(io.Writer, ...interface{})) {
-	flag.Var(&headersFlags, "H", "Additional `header`(s)")
+	flag.Var(&headersFlags, "H", "Additional http headers or grpc custom metadata (some grpc reserved or "+
+		"illegal keys are not supported, see https://grpc.io/docs/what-is-grpc/core-concepts/#metadata). "+
```
ldemailly: What's an example of illegal keys and what do you mean not supported, what happens? 
(in other words; is that warning/info necessary or can we just have the short version, something like

"Additional http header(s) or grpc metadata. Multiple `key:value` pairs can be passed using multiple -H."

run `release/updateFlags.sh` to see the actual help btw, and using \` in the text labels the value

README.md
ldemailly: Btw sorry to hear - hope you're recovering fully/quickly! no rush and thanks for having done all this despite being sick

bincommon/commonflags.go
```diff
@@ -119,7 +119,9 @@ var (
...
+		"e.g. -H \"key:value\".")
```
ldemailly:
```suggestion
	flag.Var(&headersFlags, "H", "Additional http headers or grpc metadata. "+
		"Multiple `key:value` pairs can be passed using multiple -H.")
```

fortio_main.go
```diff
@@ -525,7 +526,7 @@ func grpcClient() {
 	}
 	httpOpts := bincommon.SharedHTTPOptions()
 	if *doHealthFlag {
-		status, err := fgrpc.GrpcHealthCheck(host, *healthSvcFlag, count, &httpOpts.TLSOptions)
+		status, err := fgrpc.GrpcHealthCheck(host, *healthSvcFlag, count, &httpOpts.TLSOptions, grpcMD())
```
ldemailly: pass httpOpts.ExtraHeaders instead? (SharedHTTPOptions() is already called once just above)

fortio_main.go
```diff
@@ -536,9 +537,23 @@ func grpcClient() {
 		}
 		return
 	}
-	_, err := fgrpc.PingClientCall(host, count, httpOpts.PayloadUTF8(), *pingDelayFlag, &httpOpts.TLSOptions)
+	_, err := fgrpc.PingClientCall(host, count, httpOpts.PayloadUTF8(), *pingDelayFlag, &httpOpts.TLSOptions, grpcMD())
```
ldemailly: same

fortio_main.go
```diff
@@ -536,9 +537,23 @@ func grpcClient() {
...
 	if err != nil {
 		// already logged
 		os.Exit(1)
 	}
 }
+
+func grpcMD() map[string][]string {
```
ldemailly: could be `httpHeader2grpcMeta(headers http.Header) {` ? but... what's wrong with just using ExtraHeaders directly ? they are both a `map[string][]string` aren't they?

fortio_main.go
```diff
@@ -536,9 +537,23 @@ func grpcClient() {
...
+	md := make(map[string][]string)
+	headers := bincommon.SharedHTTPOptions().ExtraHeaders
+	for k, v := range headers {
+		k = strings.ToLower(k)
+		if k == "user-agent" {
+			// todo: map standard http headers to grpc md using the dial options
+			continue
+		}
+		md[k] = append(md[k], v...)
```
ldemailly: what differences between that and
```
md[k] = v
```
? (copying instead of referencing maybe but is it necessary?)

fortio_main.go
```diff
@@ -437,6 +437,7 @@ func fortioLoad(justCurl bool, percList []float64) {
 			Payload:            httpOpts.PayloadUTF8(),
 			Delay:              *pingDelayFlag,
 			UsePing:            *doPingLoadFlag,
+			Metadata:           grpcMD(),
```
ldemailly: same

fortio_main.go
```diff
@@ -536,9 +537,23 @@ func grpcClient() {
...
 		os.Exit(1)
 	}
 }
+
+func grpcMD() map[string][]string {
+	md := make(map[string][]string)
+	headers := bincommon.SharedHTTPOptions().ExtraHeaders
+	for k, v := range headers {
+		k = strings.ToLower(k)
+		if k == "user-agent" {
```
ldemailly: what happens if that's set is grpc meta data?

fgrpc/pingsrv.go
```diff
@@ -110,16 +111,21 @@ func PingServerTCP(port, cert, key, healthServiceName string, maxConcurrentStrea
 
 // PingClientCall calls the ping service (presumably running as PingServer on
 // the destination). returns the average round trip in seconds.
-func PingClientCall(serverAddr string, n int, payload string, delay time.Duration, tlsOpts *fhttp.TLSOptions) (float64, error) {
+func PingClientCall(serverAddr string, n int, payload string, delay time.Duration, tlsOpts *fhttp.TLSOptions, md metadata.MD,
```
ldemailly, line 115: not your fault but this argument list is quite long

I already at some point move some common stuff in TLSOptions and wondering if we should have something else common for payload, n (called Exactly in http options), destination/serverAddr (though that last one could be in TLSOptions) - maybe CommonOptions or TLSAndWebOptions or something

(I realize this comment isn't exactly actionable, it's as much for me so I don't forget)

fortio_main.go
ktalg: First, all keys are converted to lowercase here. This is a requirement of grpc-go, otherwise there will be an error.
https://github.com/grpc/grpc-go/blob/f2fbb0e07ebf3dd46aa641ee89c9f17e8083eaf6/stream.go#L170-L175
https://github.com/grpc/grpc-go/blob/f2fbb0e07ebf3dd46aa641ee89c9f17e8083eaf6/internal/metadata/metadata.go#L94-L96

Second, the processing of the `user-agent` is skipped here, just to leave this todo to identify that there is some work to be done here: grpc can only set HTTP standard headers like `user-agent`, `:authority`, etc. by passing `DialOption` to the `Dial()` , and considering the current code structure, the amount of changes to implement the conversion of these standard headers may be a bit large, so I thought it would be better to do it in another MR？
`httpHeader2grpcMeta` is indeed a better name, should use it.

bincommon/commonflags.go
ktalg: got it,I will add some clearer examples and explanation

fortio_main.go
ktalg: 🤔I may have been a little careless, I'll fix it

fortio_main.go
ktalg: It's a bit strange here, the header validation will only work for headers set using the `metadata.NewOutgoingContext` method, which I happened to use (instead of `metadata.AppendToOutgoingContext`)
The second return value is ignored:
```go
If md, _, ok := metadata.FromOutgoingContextRaw(ctx); ok {
```
But I think at least the lowercase conversion here is harmless

fortio_main.go
ktalg: As explained above, and additional work may be required to handle the `host` header separately.

fgrpc/pingsrv.go
ktalg: Yes, I see that, why not just pass `GRPCRunnerOptions`?
```go
	options := fgrpc.GRPCRunnerOptions{
		TLSOptions:  httpOpts.TLSOptions,
		Destination: host,
		Payload:     httpOpts.PayloadUTF8(),
		Delay:       *pingDelayFlag,
		Metadata:    grpcMD(),
	}
	_, err := fgrpc.PingClientCall(count, options)
```

ldemailly: yeah not sure why I didn't do that in the first place - I suppose it might be because GRPCRunnerOptions includes RunnerOptions which is more for a full load test than a single call but it'd still be cleaner to ignore fields than pass so many

fortio_main.go
```diff
@@ -536,9 +538,25 @@ func grpcClient() {
...
 }
+
+// httpHeader2grpcMetadata covert md's key to lowercase and filter invalid key.
+func httpHeader2grpcMetadata(headers map[string][]string) map[string][]string {
+	ret := make(map[string][]string)
+	for k, v := range headers {
+		k = strings.ToLower(k)
+		switch k {
+		case "content-length", "content-type":
+			log.Infof("Skipping setting metadata %s:%v", k, v)
```
ldemailly: does this trigger using -payload for grpc? (it would if the http options are normalized/processed, because that's where it sets these, if you pass a payload, before this is called)

fortio_main.go
```diff
@@ -536,9 +538,25 @@ func grpcClient() {
...
 		// already logged
 		os.Exit(1)
 	}
 }
+
+// httpHeader2grpcMetadata covert md's key to lowercase and filter invalid key.
+func httpHeader2grpcMetadata(headers map[string][]string) map[string][]string {
+	ret := make(map[string][]string)
+	for k, v := range headers {
+		k = strings.ToLower(k)
```
ldemailly, line 552: so http headers are automatically (when using Get for instance) case insensitive - it's not true/the case for grpc meta data? the grpc lib doesn't automatically set that lowercase for us?

ldemailly: I checked indeed if you don't you get
```
14:53:17 W grpcrunner.go:109> Error making grpc call: rpc error: code = Internal desc = header key "Foo" contains illegal characters not in [0-9a-z-_.]
```

fortio_main.go
ldemailly: it does:
```
$ go run . load -c 1 -n 3 -payload foo -grpc -ping -H fOo:bAr -H user-agent:foo 127.0.0.1:8079
14:54:58 I http_client.go:288> User-Agent being Set to "foo"
Fortio dev running at 8 queries per second, 10->10 procs, for 3 calls: 127.0.0.1:8079
14:54:58 I fortio_main.go:555> Skipping setting metadata content-type:[application/octet-stream]
14:54:58 I fortio_main.go:555> Skipping setting metadata content-length:[3]
```

fgrpc/grpcrunner.go
```diff
@@ -299,3 +313,23 @@ func grpcDestination(dest string) (parsedDest string) {
...
 	return parsedDest
 }
+
+// extractDialOptions extract special MD and convert them to dial options.
+func extractDialOptions(in metadata.MD) (out []grpc.DialOption) {
+	for k, v := range in {
+		switch k {
+		case "user-agent":
+			delete(in, k)
+			if v[0] == jrpc.UserAgent {
```
ldemailly, line 325: any idea why both user agents are present when set?

```
$ go run . grpcping -loglevel debug -c 1 -n 1 -payload foo -grpc -ping -H fOo:bAr -H host:foo-bar -H user-agent:test123 127.0.0.1:8079
```
yields (`go run . server -loglevel debug&` with my addition to ping srv logging of meta data:
```
15:37:31 V pingsrv.go:48> Ping called {Seq:1 Ts:1673653051146622000 Payload:foo DelayNanos:0} (meta map[:authority:[foo-bar] content-type:[application/grpc] foo:[bAr] user-agent:[test123 grpc-go/1.52.0]])
```

`user-agent:[test123 grpc-go/1.52.0]` instead of just test123

bincommon/commonflags.go
ldemailly: I simplified further / as above

ldemailly (approved):
that user-agent thing sort of bothers me but we can figure it out later

fgrpc/grpcrunner.go
ldemailly: if you have the answer please put it in #680

fgrpc/pingsrv.go
ldemailly: will address (later/maybe) in #681

fortio_main.go
ktalg: (sorry for late reply) 
As I mentioned above
https://github.com/fortio/fortio/pull/671#discussion_r1062558141
There _may_ be some problems with grpc-go's medata validation,
https://github.com/grpc/grpc-go/issues/5913
so I think we should ensure the validity of the metadata ourselves instead of relying on external conversions

Issue #681, opened 2023-01-13
further refactor common options and reduce grpc client arguments list length


ldemailly:
grpc client arguments list is quite long

At some point we moved some common stuff in TLSOptions and wondering if we should have something else common for payload, n (called Exactly in http options), destination/serverAddr (though that last one could be in TLSOptions) - maybe CommonOptions or TLSAndWebOptions or something

Or use RunnerOptions

_Originally posted by @ldemailly in https://github.com/fortio/fortio/pull/671#discussion_r1060217805_

PR #697, opened 2023-01-30
fix special grpc metadata handling

### LINKED ISSUE #700: Bug about handling grpc authority metadata
fortio version: 1.40.0
I run a fortio server with `fortio server -grpc-port=9001 -loglevel=debug` and locally with `fortio load -grpc -ping -n=5 -H="host:google.com" localhost:9001`.After running a client, the server prints:
```
16:52:04 I logger.go:150> Log level is now 0 Debug (was 2 Info)
16:52:04 V stats.go:528> Will use [50 75 90 99 99.9] for percentiles
16:52:04 Fortio 1.40.0 tcp-echo server listening on tcp [::]:8078
16:52:04 Fortio 1.40.0 udp-echo server listening on udp [::]:8078
16:52:04 Fortio 1.40.0 grpc 'ping' server listening on tcp [::]:9001
16:52:04 Fortio 1.40.0 https redirector server listening on tcp [::]:8081
16:52:04 Fortio 1.40.0 http-echo server listening on tcp [::]:8080
16:52:04 Data directory is /home/kevin
16:52:04 REST API on /fortio/rest/run, /fortio/rest/status, /fortio/rest/stop
UI started - visit:
http://localhost:8080/fortio/
(or any host/ip reachable on this server)
16:52:04 I fortio_main.go:297> Note: not using dynamic flag watching (use -config to set watch directory)
16:52:04 I fortio_main.go:305> All fortio 1.40.0 h1:jSDO/jGcyC/qTpMZZ84EZbn9BQawsWM9/RMQ9s6Cn3w= go1.19.2 amd64 linux servers started!
16:52:13 V pingsrv.go:48> Ping called {Seq:1 Ts:1675155133396903376 Payload: DelayNanos:0} (meta map[:authority:[localhost:9001] content-type:[application/grpc] user-agent:[grpc-go/1.52.1]])
16:52:13 V pingsrv.go:48> Ping called {Seq:0 Ts:1675155133396903376 Payload: DelayNanos:0} (meta map[:authority:[google.com] content-type:[application/grpc] user-agent:[fortio.org/fortio-1.40.0 grpc-go/1.52.1]])
16:52:13 V pingsrv.go:48> Ping called {Seq:0 Ts:1675155133396903376 Payload: DelayNanos:0} (meta map[:authority:[google.com] content-type:[application/grpc] user-agent:[fortio.org/fortio-1.40.0 grpc-go/1.52.1]])
16:52:13 V pingsrv.go:48> Ping called {Seq:1 Ts:1675155133396903376 Payload: DelayNanos:0} (meta map[:authority:[localhost:9001] content-type:[application/grpc] user-agent:[grpc-go/1.52.1]])
16:52:14 V pingsrv.go:48> Ping called {Seq:0 Ts:1675155133396903376 Payload: DelayNanos:0} (meta map[:authority:[google.com] content-type:[application/grpc] user-agent:[fortio.org/fortio-1.40.0 grpc-go/1.52.1]])
```
As we've seen, the metadata "authority" for some requests is not "google.com," but "localhost:9001",there's obviously something wrong here

ktalg:
fix https://github.com/fortio/fortio/issues/700

There is a bug in the current handling of special grpc metadata (user-agent,host): 
The MD itself is modified in the Dial method so that other threads cannot read the original MD while establishing a connection 

Modifications to the MD are now moved outside of the Dial and need to be called separately when the request is sent

### DIFF
--- a/fgrpc/grpcrunner.go
+++ b/fgrpc/grpcrunner.go
@@ -57,7 +57,7 @@ func Dial(o *GRPCRunnerOptions) (*grpc.ClientConn, error) {
 			return net.Dial(fnet.UnixDomainSocket, o.UnixDomainSocket)
 		}))
 	}
-	opts = append(opts, extractDialOptions(o.Metadata)...)
+	opts = append(opts, o.dialOptions...)
 	conn, err := grpc.Dial(serverAddr, opts...)
 	if err != nil {
 		log.Errf("failed to connect to %s with certificate %s and override %s: %v", serverAddr, o.CACert, o.CertOverride, err)
@@ -89,8 +89,8 @@ func (grpcstate *GRPCRunnerResults) Run(outCtx context.Context, t periodic.Threa
 	var err error
 	var res interface{}
 	status := grpc_health_v1.HealthCheckResponse_SERVING
-	if grpcstate.Metadata.Len() != 0 {
-		outCtx = metadata.NewOutgoingContext(outCtx, sanitize(grpcstate.Metadata))
+	if len(grpcstate.Metadata) != 0 { // filtered one
+		outCtx = metadata.NewOutgoingContext(outCtx, grpcstate.Metadata)
 	}
 	if grpcstate.Ping {
 		res, err = grpcstate.clientP.Ping(outCtx, &grpcstate.reqP)
@@ -121,15 +121,17 @@ type GRPCRunnerOptions struct {
 	periodic.RunnerOptions
 	fhttp.TLSOptions
 	Destination        string
-	Service            string        // Service to be checked when using grpc health check
-	Profiler           string        // file to save profiles to. defaults to no profiling
-	Payload            string        // Payload to be sent for grpc ping service
-	Streams            int           // number of streams. total go routines and data streams will be streams*numthreads.
-	Delay              time.Duration // Delay to be sent when using grpc ping service
-	CertOverride       string        // Override the cert virtual host of authority for testing
-	AllowInitialErrors bool          // whether initial errors don't cause an abort
-	UsePing            bool          // use our own Ping proto for grpc load instead of standard health check one.
-	Metadata           metadata.MD   // metadata that will be added to the request
+	Service            string            // Service to be checked when using grpc health check
+	Profiler           string            // file to save profiles to. defaults to no profiling
+	Payload            string            // Payload to be sent for grpc ping service
+	Streams            int               // number of streams. total go routines and data streams will be streams*numthreads.
+	Delay              time.Duration     // Delay to be sent when using grpc ping service
+	CertOverride       string            // Override the cert virtual host of authority for testing
+	AllowInitialErrors bool              // whether initial errors don't cause an abort
+	UsePing            bool              // use our own Ping proto for grpc load instead of standard health check one.
+	Metadata           metadata.MD       // input metadata that will be added to the request
+	dialOptions        []grpc.DialOption // grpc dial options extracted from Metadata (authority and user-agent extracted)
+	filteredMetadata   metadata.MD       // filtered version of Metadata metadata (without authority and user-agent)
 }
 
 // RunGRPCTest runs an http test and returns the aggregated stats.
@@ -160,11 +162,13 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
 	r := periodic.NewPeriodicRunner(&o.RunnerOptions)
 	defer r.Options().Abort()
 	numThreads := r.Options().NumThreads // may change
+	o.dialOptions, o.filteredMetadata = extractDialOptions(o.Metadata)
 	total := GRPCRunnerResults{
 		RetCodes:    make(HealthResultMap),
 		Destination: o.Destination,
 		Streams:     o.Streams,
 		Ping:        o.UsePing,
+		Metadata:    o.Metadata, // the original one
 	}
 	grpcstate := make([]GRPCRunnerResults, numThreads)
 	out := r.Options().Out // Important as the default value is set from nil to stdout inside NewPeriodicRunner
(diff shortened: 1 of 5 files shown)


ldemailly:
Hi @ktalg how does the issue show up? (grpc run - missing headers in some? do you have a human visible repro?)

fgrpc/grpcrunner.go
```diff
@@ -316,14 +316,26 @@ func grpcDestination(dest string) (parsedDest string) {
...
 	for k, v := range in {
 		switch k {
-		// Transfer these 2 and avoid having them duplicated in original MD
+		// Transfer these 2
 		case "user-agent":
-			delete(in, k)
 			out = append(out, grpc.WithUserAgent(v[0]))
 		case "host":
-			delete(in, k)
 			out = append(out, grpc.WithAuthority(v[0]))
```
ldemailly, line 328: is this only these 2 that were missing on next dial because removed or something else?

fgrpc/grpcrunner.go
```diff
@@ -316,14 +316,26 @@ func grpcDestination(dest string) (parsedDest string) {
 func extractDialOptions(in metadata.MD) (out []grpc.DialOption) {
```
ldemailly: seems maybe this should just return 2 things, the dial options and the filtered MD? (avoiding the separate sanitize() or at least rename sanitize for something clearer like filterMetaData() but it seems odd the 2 can't be combined)

fgrpc/grpcrunner.go
ktalg: > is this only these 2 that were missing on next dial because removed or something else?

Yes, I don't think it's a good idea to modify metadata in Dial

fgrpc/grpcrunner.go
ktalg: If I understand this correctly, then in terms of function naming, the "Dial" used for the connection should not contain side effects

ldemailly: ic the issue is the one time setup is here but somehow the grpc meta data keeps getting set in the context each time, it's not very efficient to keep "sanitizing" it though even if we keep assigning it

fgrpc/grpcrunner_test.go
```diff
@@ -546,7 +557,7 @@ func Test_extractDialOptions(t *testing.T) {
 			if !reflect.DeepEqual(len(gotOut), tt.wantOutLen) {
 				t.Errorf("extractDialOptions() = %v, want %v", len(gotOut), tt.wantOutLen)
 			}
-			if !reflect.DeepEqual(tt.args.in, tt.wantMD) {
+			if !reflect.DeepEqual(sanitize(tt.args.in), tt.wantMD) {
```
ldemailly: using the function that ends up being tested in the test isn't super great... it's tautological

ldemailly: I wanted to see the test fail first and pass with the changes in the main but the test doesn't fail, it doesn't compile without the new code

ldemailly: but removing sanitize (because the old extractDialOptions mutates tt.args.in) "works" for that purpose

ldemailly:
I'm testing a variant with setting things only once

fgrpc/grpcrunner.go
ldemailly: @ktalg please look at https://github.com/fortio/fortio/pull/701/commits/416e9d5e0a3781a95efa31e1aede251ff6bdd0cd for what I meant

(tried to push to your branch but failed so it's in 701)

ktalg: > ic the issue is the one time setup is here but somehow the grpc meta data keeps getting set in the context each time, it's not very efficient to keep "sanitizing" it though even if we keep assigning it

I'm aware of this, and it seems like a much bigger change would be needed to handle it well, and I have to be honest and say I can't assess whether such a change would be worth it 🥲 It involves a lot of design considerations...

fgrpc/grpcrunner_test.go
ktalg: > using the function that ends up being tested in the test isn't super great... it's tautological

thanks a lot for the suggestion, but what do you mean by "the function that ends up being tested"? This test function just reuses the test case to test "extractDialOptions" and "sanitize" I didn't use these 2 functions to create the expected "want result"

fgrpc/grpcrunner.go
ldemailly: well I did the change, can you review it in #701

ldemailly: I will try again to push it to your fork in a bit

ldemailly: @ktalg it worked this time, PTAL at https://github.com/fortio/fortio/pull/697/commits/d906484c90afbcc25ed1d17aa2010660916c5402
and if you think it makes sense we can merge this and put it in 1.40.1

fgrpc/grpcrunner_test.go
ldemailly: ic, yes now that it's one function it's clearer I think

ldemailly (approved):
(though I'm kinda approving my own changes too so KT ptal)

fgrpc/grpcrunner.go
ktalg: > @ktalg it worked this time, PTAL at [d906484](https://github.com/fortio/fortio/commit/d906484c90afbcc25ed1d17aa2010660916c5402) and if you think it makes sense we can merge this and put it in 1.40.1

it makes sense, thanks a lot!

ktalg:
@ldemailly  and how about improving the name of `extractDialOptions`, because it does two things

ldemailly:
> @ldemailly  and how about improving the name of `extractDialOptions`, because it does two things

yes that's a good point. will change it. code wise though, thoughts/ok?

ldemailly:
oh you did already, cool

ktalg:
> > @ldemailly  and how about improving the name of `extractDialOptions`, because it does two things
> 
> yes that's a good point. will change it. code wise though, thoughts/ok?

I think our changes are good enough, there's nothing more to suggest! 

But I seem to have violated 'DRY' in some places, and I think I should avoid these if there is a big refactoring in the future

ldemailly:
thx we can further cleanup in #681 (repeat in tests is somewhat fine btw)

PR #701, opened 2023-02-01
fix special grpc metadata handling #697 + fixes

ldemailly:
See #697 and #700


ldemailly:
managed to push to #697

Issue #266, opened 2018-06-28
blog post(s)/article about fortio (features)


ldemailly:
fortio is a pretty great tool and library (If I may say so myself) and has a lot of features, some a bit harder to discover than other. 

There is sometimes the misconception it depends on istio or kubernetes (neither are true, it has 0 dependencies except base golang and optionally grpc)

A series of blog posts or some articles in tech journals would help raise the visibility and thus usefulness

ldemailly:
from #570:

docs, blogs, book... about all the many many features of fortio

fortio has so many features that -h or the README.md or the web UI are not making for an easy onboarding or discovery of all it can do and how to best use it

default values of many flags too (like now using -a -uniform -nocatchup being more or less required for serious use etc

for now I added these to https://github.com/fortio/fortio/wiki/FAQ#i-want-to-get-the-best-results-what-flags-should-i-pass\



----


not that some chapters of the istio book from christian does have fortio examples

ldemailly:
Also of note https://github.com/fortio/fortio/pull/706#discussion_r1104522768

ie a challenge is giving more context and examples on how to use the many flags, while keeping the flag help concise

PR #706, opened 2023-02-11
moved logger to toplevel fortio.org/log, dflag and version - avoid flags in library modules

### LINKED ISSUE #433: echo-server-default-params dflag showing up when fortio is used as library
fhttp/http_server.go:   defaultEchoServerParams = dflag.DynString(flag.CommandLine, "echo-server-default-params", "",

should be in fortio_main / commandline flags instead and passed

ldemailly:
This is quite a big change, a bit hard to review because it contains both shallow change (package import change and file deletion + interesting new ones like the new dflag split API use; might be easier to look at individual commits)

Essentially to help reuse and lower pulled in dependencies for the building blocks

As well as flags pollution: 
- [x] move flags out of fnet/
- [x] move flags out of fhttp/
- [x] check other packages for flags and move as needed

Change summary for callers/users of fortio log / dflag / version:
- Remove `/fortio` in the imports. eg `fortio.org/fortio/log` becomes `fortio.org/log`
- Replace `log.SetFlagDefaultsForClientTools()` by `log.SetDefaultsForClientTools()` (for cli)
- Add either `log.LoggerStaticFlagSetup()` (cli) or `dynloglevel.LoggerFlagSetup()` (servers/dflag users, the package is `fortio.org/dflag/dynloglevel`)

Fixes #433


bincommon/commonflags.go
```diff
@@ -132,6 +133,22 @@ func SharedMain(usage func(io.Writer, ...interface{})) {
 		"Size of the buffer (max data size) for the optimized http client in `kbytes`")
 	flag.BoolVar(&fhttp.CheckConnectionClosedHeader, "httpccch", fhttp.CheckConnectionClosedHeader,
 		"Check for Connection: Close Header")
+	// FlagResolveIPType indicates which IP types to resolve.
+	// With round robin resolution now the default, you are likely to get ipv6 which may not work if
+	// use both type (`ip`). In particular some test environments like the CI do have ipv6
+	// for localhost but fail to connect. So we made the default ip4 only.
+	dflag.Flag("resolve-ip-type", fnet.FlagResolveIPType)
```
wiardvanrij, line 140: I'm trying to understand the re-use of the `fnet.FlagResolveIPType` into a new `dflag.Flag`. Partially because it's my first time checking these things.

I guess I'm wondering if there shouldn't be a more 'central' init place for all flags? Why we have some flags (?) "defined" in a different package that we can 're-init' as a dynamic flag in the 'main' part? Idk if my question makes sense but.. yea.

bincommon/commonflags.go
```diff
@@ -132,6 +133,22 @@ func SharedMain(usage func(io.Writer, ...interface{})) {
...
+	// default assumes one gets all the ips in the first call and does round robin across these.
+	// first just picks the first answer, rr rounds robin on each answer.
+	dflag.Flag("dns-method", fnet.FlagResolveMethod)
+	dflag.Flag("echo-server-default-params", fhttp.DefaultEchoServerParams)
+	dflag.FlagBool("proxy-all-headers", fhttp.Fetch2CopiesAllHeader)
+	dflag.Flag("server-idle-timeout", fhttp.ServerIdleTimeout)
+	// MaxDelay is the maximum delay allowed for the echoserver responses.
+	// It is a dynamic flag with default value of 1.5s so we can test the default 1s timeout in envoy.
+	dflag.Flag("max-echo-delay", fhttp.MaxDelay)
+	dynloglevel.LoggerFlagSetup()
```
wiardvanrij, line 152: Initial thought is; what/why is this? (haven't checked it locally and followed the reference, but might be nice to comment).

bincommon/commonflags.go
wiardvanrij: Also sort of makes it easier I guess? Not having to comment both 'sides'

wiardvanrij (commented):
Main nit would be to add additional documentation. I.e. move parts of code comments into flag docs.

Review wise.. LGTM? (I mean I do like the dynamic flags, but hard to judge if the code is correct :) )

bincommon/commonflags.go
ldemailly: So the reasoning here is the libraries define a config variable (using dflag.New) which includes the contract that they'll call .Get() to get the typed values, may setup validators, and the value can come from string using the same interface as flags.

Then a program, using the library, decides if they want to configure it as a dflag (or not)

In this case fortio is both a set of libraries, and the main user of them, so the flags are moved to bincommon for the most part

The comment on both sides is mostly that go doc wants exported variables commented (even though the 2nd argument of dflag.New is the comment/usage (which will be dumped on the flags UI and command line help) and when I copied it here I preserved it but yes it should only be needed once.

ldemailly: In short in fnet/ and fhttp/ the "flags" aren't flags anymore they are dynamic config objects, that happen to be backed by dflags

This is to avoid that if you use a small part of fnet or fhttp you end up with your small cli `-help` showing a bunch of stuff you do not use. 

Ditto btw I got rid of log's many flag in favor of a Config object and a single optional flag

bincommon/commonflags.go
ldemailly: let me add a comment

it's https://pkg.go.dev/fortio.org/dflag/dynloglevel#LoggerFlagSetup

ldemailly:
I was wondering if this should be 1.50.0 to show a big gap from current 1.40.1 - or 1.42 at least... thoughts?

bincommon/commonflags.go
ldemailly: > move parts of code comments into flag docs.

any example of which?

wiardvanrij: > > move parts of code comments into flag docs.
> 
> any example of which?

```
  -resolve-ip-type type
    	Resolve type: ip4 for ipv4, ip6 for ipv6 only, use ip for both (default ip4)
```

vs 

```
// With round robin resolution now the default, you are likely to get ipv6 which may not work if
	// use both type (`ip`). In particular some test environments like the CI do have ipv6
	// for localhost but fail to connect. So we made the default ip4 only.
```

like the "code lacks" info about which options there are and the documentation "lacks" background information / reasoning. 

Having both in a single place is nice.

wiardvanrij: but imo, unrelated to the MR.. just some food for thought on improving docs in general which is already an open issue.

wiardvanrij:
> I was wondering if this should be 1.50.0 to show a big gap from current 1.40.1 - or 1.42 at least... thoughts?

yea, kinda worth bumping it to 1.50 :)

wiardvanrij (approved):

bincommon/commonflags.go
ldemailly: So comments can be as long and as anecdotal or detailed as we want but the flag help text needs to be somewhat concise. I kinda wish the inverse: that if we don't document a config var / flag, the go doc would show the value / line for self doc

ldemailly: we need what you're working on "fortio the book/tutorials" which can be informed from both flags help line and comments with more context added.

ldemailly: noted in #266

PR #718, opened 2023-02-25
initial design/api for progressive ramp up

### LINKED ISSUE #717: ramp up qps/connections progressively
Many system can't handle say 0->10kqps  (or connections) at once so we end up having to start multiple smaller fortio

Having a ramp up would be nice addition

Api tbd... maybe 

`-ramp-up duration` will linearly increase qps and connections from 0 to duration? [or maybe seperately as

`-c start:end` `-qps start:end` over `-ramp` ?

ldemailly:
(will) Fix #717


ldemailly:
https://github.com/fortio/fortio/blob/master/scripting_example.gr sorta does that though not continously/much differently than if it was a script outside fortio

PR #721, opened 2023-02-26
Stream input back in echo h2 + h2c curl streaming payload + minimal prometheus metrics (number of fd)

ldemailly:
This is quite a big change/refactor to essentially support Reader/Writer instead of Payload input and byte[] output

Also added h2c for client (based on getting it working on https://github.com/fortio/h2cli first)

Also includes #723 (stats about open file descriptor and number of in flight runs under a new metrics endpoint) separated out for ease of review

Also fixed #305 because it was getting tiring to keep manually changing flags output and version

Also fixed divide by 0 panic when trying to resolve "" host (discovered from adding tests).

```
curl --limit-rate 1000 --data-binary @README.md -H content-type:text/plain --http2-prior-knowledge "http://localhost:8080"
```

Server
```
$ go run . server -loglevel debug
14:26:30 Echo: POST / HTTP/2.0 127.0.0.1:65090 ()  "curl/7.86.0"
14:26:30 Header Host: localhost:8080
14:26:30 Header Content-Length: 51497
14:26:30 Header User-Agent: curl/7.86.0
14:26:30 Header Accept: */*
14:26:30 Header Content-Type: text/plain
14:26:30 D http_utils.go:385> Parsing delay  -> []
14:26:30 D http_server.go:177> Request # 2
14:26:30 D http_server.go:70> FlushWriter wrote 1000
14:26:31 D http_server.go:70> FlushWriter wrote 1000
14:26:32 D http_server.go:70> FlushWriter wrote 1000
14:26:33 D http_server.go:70> FlushWriter wrote 1000
14:26:34 D http_server.go:70> FlushWriter wrote 1000
14:26:35 D http_server.go:70> FlushWriter wrote 1000
14:26:36 D http_server.go:70> FlushWriter wrote 1000
14:26:37 D http_server.go:70> FlushWriter wrote 1000
14:26:38 D http_server.go:70> FlushWriter wrote 1000
[...]
```
and on curl window you see the readme coming back in chunks with 1s pause (from the rate limit)

in h1.1 mode results would only come back at the end

also now working:
```
 (echo "foo"; sleep 3; echo "bar") | go run . curl -stream -quiet localhost:8080
foo
# ^ shows right away, then 3s pause and:
bar
```

ie h2c streaming works from `fortio curl` too (not just h2cli)

PR #726, opened 2023-03-08
fix 2XX responses reporting as errors

jneate:
I just got fed up of seeing HTTP 202 reported as an error

### DIFF
--- a/fhttp/httprunner.go
+++ b/fhttp/httprunner.go
@@ -17,7 +17,6 @@ package fhttp
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"os"
 	"runtime"
 	"runtime/pprof"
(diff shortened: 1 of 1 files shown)


fhttp/httprunner.go
```diff
@@ -69,8 +68,8 @@ func (httpstate *HTTPRunnerResults) Run(ctx context.Context, t periodic.ThreadID
 		httpstate.aborter.Abort(false)
 		log.Infof("Aborted run because of code %d - data %s", code, DebugSummary(body, 1024))
 	}
-	if code == http.StatusOK {
-		return true, "200"
+	if code >= 200 && code <= 299 {
```
ldemailly, line 71: probably should use https://github.com/fortio/fortio/blob/6d264d850c06fd305c566afe545b1197b77900bb/fhttp/http_client.go#L999-L1001 ? (trying to remember why it isn't already)

fhttp/httprunner.go
```diff
@@ -69,8 +68,8 @@ func (httpstate *HTTPRunnerResults) Run(ctx context.Context, t periodic.ThreadID
...
+		return true, fmt.Sprint(code)
```
ldemailly, line 72: should be strconv.Itoa(code) (and line 74 too so a preexisting problem), also noticing line 61 (calling Fetch instead of StreamFetch is also wrong - puzzled how I missed that in #721 refactor)

ldemailly:
Thanks for the bug report, fixing it in #727 - ptal

jneate:
> Thanks for the bug report, fixing it in #727 - ptal

Do you want this closing if you're fixing it in #727?

I'm happy to make the changes here but happy either way

jneate:
Closing as fixed in #727

PR #727, opened 2023-03-09
go1.19.7 + codeOK fix + streaming misses

ldemailly:
- new go version
- refactor of client api usage from always returning the bytes to streaming data had 2 misses, including critical one in Run()
- code to tell access logger success/failure was hardcoded to 200s only instead of using same range as elsewhere

### DIFF
--- a/fhttp/httprunner.go
+++ b/fhttp/httprunner.go
@@ -17,11 +17,11 @@ package fhttp
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"os"
 	"runtime"
 	"runtime/pprof"
 	"sort"
+	"strconv"
 	"sync"
 
 	"fortio.org/fortio/periodic"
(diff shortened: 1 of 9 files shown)


bseenu (approved):

fhttp/httprunner.go
```diff
@@ -59,20 +59,16 @@ type HTTPRunnerResults struct {
...
 	if httpstate.AbortOn == code {
 		httpstate.aborter.Abort(false)
-		log.Infof("Aborted run because of code %d - data %s", code, DebugSummary(body, 1024))
+		log.Infof("Aborted run because of code %d - data len %d", code, size)
 	}
-	if code == http.StatusOK {
-		return true, "200"
-	}
-	return false, fmt.Sprint(code)
+	return codeIsOK(code), strconv.Itoa(code)
```
ldemailly, line 71: that initial implementation when I added support for success/error of each run/access log was (my) bad

fixed now, thx for noticing @jneate - I will now close #726

bseenu (approved):

PR #768, opened 2023-06-12
using json structured logging, add `log-errors` option setting to rapi and UI

ldemailly:
integrates

https://github.com/fortio/log/pull/22
https://github.com/fortio/scli/pull/11

also show runid in relevant logs so errors and end of run can be attached to corresponding run


wuhaoyujerry (approved):

fhttp/http_client.go
```diff
@@ -203,6 +203,7 @@ type HTTPOptions struct {
 	Payload          []byte        // body for http request, implies POST if not empty.
 	LogErrors        bool          // whether to log non 2xx code as they occur or not
 	ID               int           `json:"-"` // thread/connect id to use for logging (thread id when used as a runner)
+	UniqueID         int64         `json:"-"` // Run identifier when used through a runner, copied from RunnerOptions.RunID
```
bseenu, line 206: why is this needed when you already have runID ?

bseenu (approved):

fhttp/http_client.go
ldemailly: So there is RunID on the RunnerOptions indeed but this (HTTPOptions) is what the client (which is often called by the runner code but not always) need - and I called it UniqueID so when using HTTPRunnerOptions you don't end up with ambiguous double name (as HTTPRunnerOptions has both RunnerOptions which has RunID and HTTPOptions)

Short version is it's indeed a bit ugly

PR #785, opened 2023-06-30
do abort early even with stdclient on bad urls/hosts. allow most same query args on fetchers.

### LINKED ISSUE #784: putting an empty or invalid (non resolving) url doesn't abort early - UI and rapi
also somewhat related "allow initial errors" is harcoded to true and while it's a good default it's another reason run proceeds with 100% errors

ldemailly:
- fixes #784 

Screenshot 2023-06-30 at 5 13 15 PM

instead of a run with 100% errors

- Also use UI form params for fetch and fetch2 endpoints (so one can pass extra headers with &H=foo:bar and turn off secure fetch using `https-insecure=on` etc


fhttp/http_client.go
```diff
@@ -380,6 +380,20 @@ func newHTTPRequest(o *HTTPOptions) (*http.Request, error) {
 	}
 	//nolint:noctx // we pass context later in Run()/Fetch()
 	req, err := http.NewRequest(method, o.URL, body)
+	if err == nil { //nolint:nestif
+		// Additional validation for the URL so we abort early on fatal errors even for the std client.
+		// fixes #784
+		if req.URL == nil || req.URL.Host == "" {
```
wiardvanrij, line 386: Doesn't it seem more logical to do this sanity check even before newHttpRequest? I.e., somewhere in the flag handling?

fhttp/http_client.go
```diff
@@ -380,6 +380,20 @@ func newHTTPRequest(o *HTTPOptions) (*http.Request, error) {
...
+			err = fmt.Errorf("invalid url '%s'", o.URL)
```
wiardvanrij, line 387: It took me (too) long to realise that you are setting `err` here for line 383 if I'm honest. Same goes for the logic at line 394. 

I do understand you want to reuse the contex of the logger... but then maybe make a small func as in "return err with context" and then do something like

```
return nil, returnErrorWithContext("invalid url")
```

where `returnErrorWithContext`  func is the contents of:

```
		log.S(log.Error, "Unable to make request",
			log.Attr("method", method), log.Attr("url", o.URL), log.Attr("err", err),
			log.Attr("thread", o.ID), log.Attr("run", o.UniqueID))
```

wiardvanrij (approved):

fhttp/http_client.go
ldemailly: flag handling? this can be called (is called) by UI handler and rapi so no flag - this is to catch/fix #784

fhttp/http_client.go
ldemailly: yeah it was to avoid repeating / share the log.S indeed. but you're right it'd be a bit clearer/cleaner with a function, specially if used in more places. there is a small downside though which is the line number in the log will be the one of the function (though maybe that's also something I could try to change, expose the depth)

fhttp/http_client.go
wiardvanrij: 'whatever input' handling then :p 

More my point is; can't we check earlier than this point?

ldemailly: it's the first thing done in NewStdClient

https://github.com/fortio/fortio/blob/v1.55.2/fhttp/http_client.go#L559-L561

so there isn't really anything "earlier" no

ldemailly: trying and failing to get github to show code preview from links, by hand:

```go
func NewStdClient(o *HTTPOptions) (*Client, error) {
	o.Init(o.URL) // also normalizes NumConnections etc to be valid.
	req, err := newHTTPRequest(o)
	if req == nil {
		return nil, err
	}
```

fhttp/http_client.go
ldemailly: Will address (maybe ;-) in #787

fhttp/http_client.go
ldemailly: in theory it could be inside Init but that wouldn't cover the case the url is changed after init etc

PR #792, opened 2023-07-12
fix workflow too

ldemailly:
release build https://github.com/fortio/fortio/actions/runs/5534055141/jobs/10098486224 failed because of ppc64le

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,8 @@ jobs:
         uses: docker/build-push-action@2eb1c1961a95fc15694676618e422e8ba1d63825 # pin@v3
         with:
           context: .
-          platforms: linux/amd64,linux/arm64,linux/ppc64le,linux/s390x
+          # platforms: linux/amd64,linux/arm64,linux/ppc64le,linux/s390x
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -54,7 +54,8 @@ jobs:
         uses: docker/build-push-action@2eb1c1961a95fc15694676618e422e8ba1d63825 # pin@v3
         with:
           context: .
-          platforms: linux/amd64,linux/arm64,linux/ppc64le,linux/s390x
+          # platforms: linux/amd64,linux/arm64,linux/ppc64le,linux/s390x
+          platforms: linux/amd64,linux/arm64,linux/s390x
```
bseenu, line 58: who uses linux/s390x now a days ? :)

bseenu (approved):

.github/workflows/main.yml
ldemailly: you'd be surprised.... but IBM which is a major contributor to Istio is... :) 

dunno who uses ppc64le... hopefully nobody until they fix that error

PR #797, opened 2023-07-18
Save the async result file even on init error (like invalid host/target)

### LINKED ISSUE #796: rapi calls should return failure or put the run results on init errors (e.g invalid host)
right now

curl "localhost:8080/fortio/rest/run?url=http://doesntexist.fortio.org&save=on&async=on&log-errors=on"

returns 
```json
{"message":"started","RunID":1,"Count":1,"ResultID":"2023-07-17-163631_1","ResultURL":"http://localhost:8080/fortio/data/2023-07-17-163631_1.json"}
```

but in logs:
```
16:36:31.536 E restHandler.go:417> Init error for http mode with url http://doesntexist.fortio.org and options &{RunType: Runners:[] QPS:0 Duration:0s NumThreads:0 Percentiles:[50 75 90 99 99.9] Resolution:0 Out:0x1400012e010 Labels: Stop:<nil> Exactly:0 Jitter:false Uniform:false RunID:1 Offset:0s AccessLogger:<nil> NoCatchUp:false ID:2023-07-17-163631_1 genTime:0x1400040e018} : lookup doesntexist.fortio.org: no such host
```

and then later empty "stop" call / no result url 

also
```
curl http://localhost:8080/fortio/data/2023-07-17-163631_1.json
```
404 not found

ldemailly:
Fixes #796 

Opted for async save of json so there is a trace and also because of concern of messing up the aborting/init order logic (which isn't the best reason)

Issues

- [x] json file id generation different async vs not vs error
- [x] trying to stop an init erroring run with &wait=true seems to hang: fixed
- [ ] maybe need to revisit the whole aborter (probably not as that's a huge change) but at least fix above point
- [ ] might need to make "async" an option of periodic runner directly so init errors can be handled and returned before actual async run start
- [x] check the UI can display the error json files - done including backward compatibility, will show as error on older fortio report too
- [x] `go test -v -coverprofile=profile.out -covermode=atomic -timeout 15s -race ./rapi 2>&1  | logc` fails while it passes without coverprofile - no race detected either way (stop seems to hang) - need to check the go routine dump / stack traces


wuhaoyujerry (approved):

ldemailly:
Only the coverage mode unearth a concurrency hang/bug (yet -race doesn't complain)
```bash
go test -v -coverprofile=profile.out -covermode=atomic -timeout 15s -race ./rapi 2>&1  | ~/go/bin/logc
```

ldemailly:
currently issue of stop concurrency issue sort of reproducible with
```
go test -test.count=1 -timeout 8s -run "^TestHTTPRunnerRESTApiBadHost$" ./rapi | ~/go/bin/logc
```

edit - fixed

ldemailly:
codecov is a bit drunk as it started to count 
```
}
```
lines as not covered while not before, this dropping the % more because of some log fatal/bug cases...etc..

PR #817, opened 2023-08-07
[jrpc] Add support for optional custom allowed http status codes and custom TLS options

ldemailly:
For instance to skip cert verification or only accept very specific code instead of the 200, 201, 202 defaults

See the new test

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<!-- 1.58.0 -->
+<!-- 1.59.0 -->
 # Fortio
 
 [![Awesome Go](https://fortio.org/mentioned-badge.svg)](https://github.com/avelino/awesome-go#networking)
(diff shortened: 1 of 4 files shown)


wiardvanrij (approved):
This is good, however, consider passing along the entire httpClient. 

I'm used to this: https://github.com/thanos-io/thanos/blob/main/pkg/httpconfig/http.go#L33 
While obviously it's not needed in 99% of the cases, allowing the clients freedom to pass along the entire httpClient is pretty neat.

ldemailly:
> This is good, however, consider passing along the entire httpClient.
> 
> I'm used to this: https://github.com/thanos-io/thanos/blob/main/pkg/httpconfig/http.go#L33 While obviously it's not needed in 99% of the cases, allowing the clients freedom to pass along the entire httpClient is pretty neat.

Done: 

https://github.com/fortio/fortio/pull/817/commits/e1ae993215e252c7899d7de22e03ca74a4e63e85


ps: aint it fun to decrease coverage on unchanged files... codecov is having issues

PR #818, opened 2023-08-08
Remove the empty file .gitmodules

haiyuewa:
The commit c2b5531cd3ea ("Switching to go1.14 and go modules and update all dependencies (#363)") has removed the git submodule directory.

Just remove the empty file .gitmodules to make git clean.


ldemailly:
Thanks for noticing, I'm curious... how does it show up / is there a downside to the empty file (beside being a relic/unnecessary, which is enough to remove it indeed)

ldemailly (approved):
thx!

haiyuewa:
> Thanks for noticing, I'm curious... how does it show up / is there a downside to the empty file (beside being a relic/unnecessary, which is enough to remove it indeed)

Just found two strange git update message by my simple script[1], but no "Fetching submodule" message
for git repo which has real submodule. So deep into the fortio git repo metadata for detail. ; - )
   Already up to date.
   Already up to date.

git-up-all.sh [1]:
                cd $repository
                git remote prune origin
                git up

                if [ -f .gitmodules ]; then
                        git submodule update --init --recursive
                        git submodule foreach --recursive git remote prune origin
                        git pull --recurse-submodules
                fi

PR #820, opened 2023-08-10
cli grpc support profile option

wulianglongrd:


ldemailly:
Thanks, does it end up being used for grpc though?

ldemailly (approved):
See Question above

ldemailly (approved):
I checked it does work already

PR #866, opened 2023-12-02
adds support to have nowUTCToken 

tomParty:
4d47b5598f2f5dee8b50c5f948bbb4c5740c7cb2 was needed as there were refernces to atomic.Bool which is introduced in 1.19 and currently mod was set to 1.18

4d47b5598f2f5dee8b50c5f948bbb4c5740c7cb2 introduces nowUTCToken to set all instances in a payload to the same now in utc


fhttp/http_client.go
```diff
@@ -513,12 +517,25 @@ func (c *Client) StreamFetch(ctx context.Context) (int, int64, uint) {
...
 		req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
 	} else if len(c.body) > 0 {
 		req.Body = io.NopCloser(bytes.NewReader(c.body))
 	}
+
+	if c.bodyContainsNOWUTC {
+		nowUTC := nowFn().UTC()
+		body := string(c.body)
+		for strings.Contains(body, nowUTCToken) {
+			body = strings.Replace(body, nowUTCToken, nowUTC.String(), 1)
```
ldemailly: .String() once, outside of the loop

but why not replaceall

fhttp/http_client.go
```diff
@@ -513,12 +517,25 @@ func (c *Client) StreamFetch(ctx context.Context) (int, int64, uint) {
 		for strings.Contains(body, uuidToken) {
 			body = strings.Replace(body, uuidToken, generateUUID(), 1)
 		}
+		c.body = []byte(body)
```
ldemailly, line 520: why is this needed but wasn't for {uuid}

ldemailly (changes requested):
The string nowUTC is a bit odd - I am wondering also if the format is a bit arbitrary, what if I want to insert unix timestamps or something else

so maybe... {custom} calling a custom function passed in instead? (and that function can return time.Now().UTC().String() for your use case but something else for someone else?

fhttp/http_client.go
tomParty: yea sounds good i'l just use replaceall, I think I was trying to just keep similar footprint as was was done for uuid

fhttp/http_client.go
tomParty: when there's a body that has both payloads it would overwrite the req.Body without the {uuid} interpolation. so I made sure that we updated it before processing down the chain

tomParty:
> The string nowUTC is a bit odd - I am wondering also if the format is a bit arbitrary, what if I want to insert unix timestamps or something else
> 
> so maybe... {custom} calling a custom function passed in instead? (and that function can return time.Now().UTC().String() for your use case but something else for someone else?

This would be awesome. I'm curious how we'd register the custom func or exec. or you imagining something like
```
{time.Now().UTC().String()}
{custom | time.Now().UTC().String()}
```

fhttp/http_client.go
ldemailly: maybe uuid can use fixing too :) that also was contributed and I might have missed it in the then review

ldemailly:
> > The string nowUTC is a bit odd - I am wondering also if the format is a bit arbitrary, what if I want to insert unix timestamps or something else
> > so maybe... {custom} calling a custom function passed in instead? (and that function can return time.Now().UTC().String() for your use case but something else for someone else?
> 
> This would be awesome. I'm curious how we'd register the custom func or exec. or you imagining something like
> 
> ```
> {time.Now().UTC().String()}
> {custom | time.Now().UTC().String()}
> ```

Well I was thinking in the library case one can set what "custom" function is in the HTTPClientOption, but for the default cli that's indeed a good question how it could be used; we could have a flag with a few built in (unixts,utc,sequence,...) as I don't think go can support compiling at run time - on the other hand... that's what templates are also?

ldemailly:
We could have a couple of variants and use templates with a function map also but it wouldn't be fully dynamic (kinda by design of template safety)

fhttp/http_client.go
tomParty: hahah i'll look into it. I think I remember seeing uuid are unique.. so if you had a payload with multiple, they'd be all different..

tomParty:
> We could have a couple of variants and use templates with a function map also but it wouldn't be fully dynamic (kinda by design of template safety)

yea, hoping to have some time tomorrow or this weekend to tinker around with it more, at a min, I might just add a set of functions/tokens  like you mentioned before `unixts,utc,sequence,..`

ldemailly:
Btw one way to do extensibility could be https://github.com/ldemailly/go-scratch/blob/main/plugin_example/extensible_main.go

ldemailly:
We all can get busy (me too!), no problem, just curious @tomParty  if you plan on getting back to this?

ldemailly:
@tomParty  ?

PR #881, opened 2024-01-15
re-enable ifElseChain in go critic

ldemailly:
whether readability is improved (outside of the clear case like when it's indeed a switch on a single value) is tbd


fhttp/http_utils.go
```diff
@@ -202,11 +202,12 @@ func ParseChunkSize(inp []byte) (int64, int64) {
 		if inDigits { //nolint:nestif
 			b := toUpper(inp[off])
 			var digit int64
-			if b >= 'A' && b <= 'F' {
+			switch {
+			case b >= 'A' && b <= 'F':
```
PrabhdeepsGill, line 206: 👍

PrabhdeepsGill (approved):

PR #908, opened 2024-03-22
Using (partially) the shared workflows

ldemailly:
Like done in many other fortio repos over last 2 days, use https://github.com/fortio/workflows
but keep the build using the fortio build image and not goreleaser

Introduced a shared action for checkout+qemu+docker login+docker buildx shared between releaser workflow and here

https://github.com/fortio/workflows/blob/main/.github/actions/common_setup/action.yml

This will eliminate most of the dependabot PRs except docker/build-push-action and action/create-release
(which are only used here)

also switched coverage to be same with added `dependencies` step in workflows/


ldemailly:
looking at why the coverage dropped, it seems it's not doing the rerun with debug on or some such (not counting debug only lines)

ldemailly:
> looking at why the coverage dropped, it seems it's not doing the rerun with debug on or some such (not counting debug only lines)

seems to have to do with atomic coverage though... that seems like a bug

bseenu (approved):

ldemailly:
> > looking at why the coverage dropped, it seems it's not doing the rerun with debug on or some such (not counting debug only lines)
> 
> seems to have to do with atomic coverage though... that seems like a bug

so -race is what removes the debug lines... somehow... anyway... fixed through https://github.com/fortio/workflows/pull/10

PR #917, opened 2024-04-03
bump grpc and prep for 1.63.7 

ldemailly:
~Tbd (ie tests results) for whether NewClient() is a drop… in replacement for grpc.Dial()~

Edit: looks like `grpc.NewClient` isn't a drop in replacement for `grpc.Dial` so keeping the working one for now


ldemailly:
addressed in #917

ldemailly:
not sure what's up with coverage drop by ~ 10% but seems to be go1.22 related

tejaspbajaj (commented):
lgtm

PR #926, opened 2024-05-30
Avoid warmup grpc connection multiple times

ktalg:
Currently, each new grpc connection is used to send multiple warmup requests (when grpc-stream>0), which is unexpected, especially for scenarios where the server takes a long time to process, and the whole warmup process will consume a lot of time.

### DIFF
--- a/fgrpc/grpcrunner.go
+++ b/fgrpc/grpcrunner.go
@@ -186,7 +186,8 @@ func RunGRPCTest(o *GRPCRunnerOptions) (*GRPCRunnerResults, error) {
 	ts := time.Now().UnixNano()
 	for i := 0; i < numThreads; i++ {
 		r.Options().Runners[i] = &grpcstate[i]
-		if (i % o.Streams) == 0 {
+		newConn := i%o.Streams == 0
+		if newConn {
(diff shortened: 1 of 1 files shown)


ldemailly:
Thanks - I find it surprising that warming only the first stream of each multiplexed connections makes a difference, specially with the ping server itself, is that what you see?

ldemailly:
can you add also the nolint so gocyclo / ci passes?

ktalg:
> Thanks - I find it surprising that warming only the first stream of each multiplexed connections makes a difference, specially with the ping server itself, is that what you see?

In a testing scenario where the client and server are far apart, each request takes 50ms to reach its destination. When conducting the test: `fortio load -grpc -ping -c=10 -s=100 -qps=0 -H="host: server" -gomaxprocs=2 -t=30s <server addr>`, fortio requires 10x100x50=50000ms=50s for warmup! This time cost is even greater than the duration of the test itself, which seems unreasonable. I believe that warming up each connection only once should suffice, as this would not cause any adverse effects.

ldemailly (approved):
Thanks !

PR #938, opened 2024-06-24
No need to copy cert bundle anymore

ldemailly:
- [x] Shorter Docker - confirm we have a test failing
    confirmed:`"tls: failed to verify certificate: x509: certificate signed by unknown authority`
- [x] go get -u fortio.org/scli -> fixed test


Dockerfile
```diff
@@ -9,11 +9,8 @@ RUN make -C fortio official-build-version BUILD_DIR=/build MODE=${MODE}
 
 # Minimal image with just the binary and certs
 FROM scratch as release
-# NOTE: the list of files here, if updated, must be changed in release/Dockerfile.in too
-COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
-# TODO: get rid of *.bak, *~ and other spurious non source files
-#COPY --from=build /build/fortio/ui/static /usr/share/fortio/static
-#COPY --from=build /build/fortio/ui/templates /usr/share/fortio/templates
```
ldemailly, line 16: also removed this as that's commented out since go:embed was released

hhhapz (approved):

PR #941, opened 2024-06-29
chore: fix acronyms, typos, and style by using jargonLint

ccoVeille:
I used my own project to spot the changes

https://github.com/jargonLint/jargonLint

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,16 +14,16 @@ Fortio (Φορτίο) started as, and is, [Istio](https://istio.io/)'s load test
 
 Fortio is also used by, among others, [Meshery](https://docs.meshery.io/extensibility/load-generators).
 
-Fortio runs at a specified query per second (qps) and records an histogram of execution time
-and calculates percentiles (e.g. p99 ie the response time such as 99% of the requests take less than that number (in seconds, SI unit)).
+Fortio runs at a specified query per second (qps) and records a histogram of execution time
+and calculates percentiles (e.g., p99 i.e., the response time such as 99% of the requests take less than that number (in seconds, SI unit)).
 It can run for a set duration, for a fixed number of calls, or until interrupted (at a constant target QPS, or max speed/load per connection/thread).
 
-The name fortio comes from greek [φορτίο](https://fortio.org/fortio.mp3) which means load/burden.
+The name fortio comes from Greek [φορτίο](https://fortio.org/fortio.mp3) which means load/burden.
 
-Fortio is a fast, small (4Mb docker image, minimal dependencies), reusable, embeddable go library as well as a command line tool and server process,
+Fortio is a fast, small (4Mb Docker image, minimal dependencies), reusable, embeddable go library as well as a command line tool and server process,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,13 +37,13 @@ We also have moved some of the library to their own toplevel package, like:
 - Version helper: [fortio.org/version](https://github.com/fortio/version#version)
 - CLI helpers integrating the above to reduce toil making new tools [fortio.org/cli](https://github.com/fortio/cli#cli) and servers [fortio.org/scli](https://github.com/fortio/scli#scli) for arguments, flags, usage, dynamic config, etc...
 
-If you want to connect to fortio using https and fortio to provide real TLS certificates, or to multiplex grpc and regular http behind a single port, check out [Fortio Proxy](https://github.com/fortio/proxy#fortio-proxy).
+If you want to connect to fortio using HTTPS and fortio to provide real TLS certificates, or to multiplex gRPC and regular HTTP behind a single port, check out [Fortio Proxy](https://github.com/fortio/proxy#fortio-proxy).
```
ldemailly, line 40: thanks!

https and http I'm not sure... it "screams" a bit in English and has become somewhat of a noun ?
(gRPC ok)

README.md
```diff
@@ -70,7 +70,7 @@ rpm -i https://github.com/fortio/fortio/releases/download/v1.65.0/fortio-1.65.0-
 # and more, see assets in release page
 ```
 
-On a MacOS you can also install Fortio using [Homebrew](https://brew.sh/):
+On a macOS you can also install fortio using [Homebrew](https://brew.sh/):
```
ldemailly: this probably should be
```
On a Mac ...
```
or
```
On Mac OS ...
```
But I guess you're technically correct officially it's macOS (or ... Darwin I guess but nobody knows that term)

What's wrong with `Fortio` btw

ldemailly (commented):
Nice thanks - how did you find all these btw? some tool?

README.md
ccoVeille: They are acronym

Hypertext Transfer Protocol Secure (HTTP): https://en.wikipedia.org/wiki/HTTP
Hypertext Transfer Protocol Secure (HTTPS): https://en.wikipedia.org/wiki/HTTPS

They might be screaming, but they are expected to be acronym.

I can discard them if you want no problem anyway.

please finish the review and tell me

README.md
ccoVeille: it's no longer Mac OS, but macOS

https://en.wikipedia.org/wiki/MacOS

> macOS, originally Mac OS X, previously shortened as OS X

Side note, I'm not an Apple-fanboy :grin: 

But you are right, I should have used "On macOS"

ccoVeille: about Fortio I hesitated, your documentation and code comments use randomly Fortio or fortio, I found fortio more often, so I changed it here, unsure why, maybe because I was updating the line

ldemailly: It's Fortio (Φορτίο even) when talking about the project but the cli is `fortio`

so here it's (I think) Fortio even if it's to end up with `fortio` binary

ldemailly:
```suggestion
On macOS you can also install Fortio using [Homebrew](https://brew.sh/):
```

ldemailly (approved):

PR #950, opened 2024-07-02
go 1.22.5 and pull log/cli/scli updates and prep for 1.66.0

ldemailly:
Edit: image finally is there (took 18+h) 

Edit: google.golang.org/grpc 1.65.0 forces go 1.21 :/ sigh but... so we get to 1.66.0 - maybe it's better anyway because of build tags etc

I usually rebuild the build image but the base isn't there yet

```
$ docker pull golang:1.22.5
Error response from daemon: manifest for golang:1.22.5 not found: manifest unknown: manifest unknown
```

so using the `toolchain` trick to get the right version


hhhapz (approved):
looks fine, although i would just wait for the docker image to be available before pushing to not require another release once it is.

ldemailly:
it's taking forever for docker and go-setup to update :/

ccoVeille (approved):

ldemailly:
> i would just wait for the docker image to be available 

Me too but it's been > 5h...

https://github.com/docker-library/golang/issues/525

PR #969, opened 2024-09-06
go1.22.7 (and newer testscript,grpc,...)

ldemailly:
Plus the usual "new linters" stuff

Plus changes to make the new windows CI pass


tejaspbajaj (approved):

Makefile
```diff
@@ -49,9 +50,17 @@ certs-clean:
 
 TEST_TIMEOUT:=90s
 
+OS:=$(shell go env GOOS)
+
 # Local test
+ifeq ($(OS),windows)
+test:
+	@echo "Skipping most tests on Windows until we can get cert-gen to work there."
```
ccoVeille, line 58: Did you try mkcert?

https://github.com/FiloSottile/mkcert

It works pretty well

ldemailly: looks nice, but I don't want to put it the store, the test certs are needed as files

I should figure out why it fails on windows git bash but... later

ldemailly: #970

PR #974, opened 2024-09-12
Introduce DelayedErrorReader to avoid the case where we get both EOF and some data in a single read

### LINKED ISSUE #972: EOF error for non keep alive HTTP 1.1 requests
**The version of the project you were using:** v1.66.2 (582d82a75a3172cc40eb0fe8e807c753b2a53770)

**Operating system you are using:** Arch Linux (6.10.5-arch1-1)

**The exact, minimal, steps needed to reproduce the issue:** Unfortunately I am not able to implement a minimal example that reliable reproduces the issue.

**Problem description:**
When i try to load test a specific application using the following command, i always get a `EOF` error resulting in a `Code  -1` statistic:
```
./fortio load -n 1 -c 1 --qps 0 -keepalive=false  https://somehost/api/infos
10:04:14.138 r1 [INF] scli.go:122> Starting, command="Φορτίο", version="dev  go1.23.0 amd64 linux", go-max-procs=8
Fortio dev running at 0 queries per second, 8->8 procs, for 1 calls: https://somehost/api/infos
10:04:14.139 r1 [INF] httprunner.go:121> Starting http test, run=0, url="https://somehost/api/infos", threads=1, qps="-1.0", warmup="parallel", conn-reuse=""
Starting at max qps with 1 thread(s) [gomax 8] for exactly 1 calls (1 per thread + 0)
10:04:14.153 r1 [ERR] http_client.go:1100> Read error, err="EOF", size=0, header_len=0, dest={"IP":"0.0.0.0","Port":443,"Zone":""}, url="https://somehost/api/infos", thread=0, run=0
10:04:14.154 r1 [INF] periodic.go:851> T000 ended after 12.940561ms : 1 calls. qps=77.27640246817738
Ended after 13.004645ms : 1 calls. qps=76.896
10:04:14.154 r1 [INF] periodic.go:581> Run ended, run=0, elapsed=13004645, calls=1, qps=76.8956015331445
Aggregated Function Time : count 1 avg 0.01293497 +/- 0 min 0.01293497 max 0.01293497 sum 0.01293497
>= 0.012935 <= 0.012935 , 0.012935 , 100.00, 1
# target 50% 0.012935
# target 75% 0.012935
# target 90% 0.012935
# target 99% 0.012935
# target 99.9% 0.012935
Error cases : count 1 avg 0.01293497 +/- 0 min 0.01293497 max 0.01293497 sum 0.01293497
>= 0.012935 <= 0.012935 , 0.012935 , 100.00, 1
# target 50% 0.012935
# target 75% 0.012935
# target 90% 0.012935
# target 99% 0.012935
# target 99.9% 0.012935
# Socket and IP used for each connection:
[0]   1 socket used, resolved to 0.0.0.0:443, connection timing : count 1 avg 0.007703781 +/- 0 min 0.007703781 max 0.007703781 sum 0.007703781
Connection time histogram (s) : count 1 avg 0.007703781 +/- 0 min 0.007703781 max 0.007703781 sum 0.007703781
>= 0.00770378 <= 0.00770378 , 0.00770378 , 100.00, 1
# target 50% 0.00770378
# target 75% 0.00770378
# target 90% 0.00770378
# target 99% 0.00770378
# target 99.9% 0.00770378
Sockets used: 1 (for perfect keepalive, would be 1)
Uniform: false, Jitter: false, Catchup allowed: true
IP addresses distribution:
0.0.0.0:443: 1
Code  -1 : 1 (100.0 %)
Response Header Sizes : count 1 avg 0 +/- 0 min 0 max 0 sum 0
Response Body/Total Sizes : count 1 avg 0 +/- 0 min 0 max 0 sum 0
All done 1 calls (plus 0 warmup) 12.935 ms avg, 76.9 qps
```
Initially I debugged the client-server communication using wireshark, but did not find any abnormalities. That said, from the network perspective everything worked correctly, which means that all data was transferred completely from the server to the client and the and the connections (TLS as well as TCP) were closed normally.
I then started debugging fortio on the latest master branch (582d82a75a3172cc40eb0fe8e807c753b2a53770) and noticed that `fhttp/http_client.go:1074`  (`nI, err := conn.Read(c.buffer[c.size:])`) was actually returning the `EOF` error. When inspecting the content of `c.buffer` right after the `conn.Read` call I also noticed the the buffer already contained all the data sent by the server. `c.size` however was still 0 since this variable is increased at some later point (`fhttp/http_client.go:1104`). As far as I understand, `fhttp/http_client.go:1089` should handle the case where we get a `EOF` error as part of a normal end of stream connection termination. In this case however, the condition cannot evaluate to true since `c.size` is still `0`. As a result the request is interpreted as failed event though it was successful.
As far as I understand the issue is caused by the rather unusual behavior of the TLS socket's `Read` function which in this case reads the data completely and returns `EOF` error. Would it instead read the data and return `EOF` error on the next iteration call (just like in regular C code), then the current implementation would work correctly.

**Reproducing the problem:**
In my setup I was connecting to a nginx:1.27.1-alpine on the local lan that implemented TLS termination (TLS 1.2) and proxied the requests to a node.js application on the same lan, serving a static in-memory response. I noticed, that the problem could be reproduced more easily if the response from the node.js application was rather small (~413 bytes including headers) and the response was sent with very low latency.
NB: The same request using HTTP instead of HTTPS (still passing the nginx reverse proxy) never had these problems.

**Possible solution:**
IMO the following patch should fix the problem (at least it did in my case), however I'm not sure about the implications of such a change:
```
diff --git a/fhttp/http_client.go b/fhttp/http_client.go
index 21d5958..021a740 100644
--- a/fhttp/http_client.go
+++ b/fhttp/http_client.go
@@ -1086,7 +1086,7 @@ func (c *FastClient) readResponse(conn net.Conn, reusedSocket bool) {
                                        c.code = RetryOnce // special "retry once" code
                                        return
                                }
-                               if errors.Is(err, io.EOF) && c.size != 0 {
+                               if errors.Is(err, io.EOF) && (c.size > 0 || n > 0) {
                                        // handled below as possibly normal end of stream after we read something
                                        break
                                }
```

ldemailly:
fixes #972 but a bit messy way/too generic


ldemailly:
https://badsslcom is having some issues causing test failures

ldemailly:
tests are more stable with `netgo` build tag

tejaspbajaj (approved):
nice, lgtm

PR #976, opened 2024-09-16
Use safecast / eliminate most nolint:gosec

ldemailly:
https://pkg.go.dev/fortio.org/safecast

Tested locally still around 180k qps for fast client so it doesn't seem to affect noticeably performance


fhttp/http_client.go
```diff
@@ -1087,9 +1088,9 @@ func (d *DelayedErrorReader) Close() error {
 
 // Response reading:
 //
-//nolint:nestif,funlen,gocognit,gocyclo,maintidx,gosec // TODO: refactor - unwiedly/ugly atm.
+//nolint:nestif,funlen,gocognit,gocyclo,maintidx // TODO: refactor - unwiedly/ugly atm.
```
ldemailly, line 1091: and it's now more ugly... probably should refactor the types and/or remember the len->int64 conversions

asha-oss (approved):

ldemailly:
going to merge this but if you find anything odd lmk and I can follow up

PR #977, opened 2024-09-16
Static assign larger payload

wolgastj:


fhttp/http_client.go
```diff
@@ -647,6 +647,7 @@ func NewStdClient(o *HTTPOptions) (*Client, error) {
 		DialContext:         dialCtx,
 		TLSHandshakeTimeout: o.HTTPReqTimeOut,
 		ForceAttemptHTTP2:   o.H2,
+		WriteBufferSize:     1048576,
```
ldemailly, line 650: curious how was that needed? of note btw the fast client does have a fixed buffer 
```
  -httpbufferkb kbytes
    	Size of the buffer (max data size) for the optimized HTTP client in kbytes (default 128)
```

PR #999, opened 2024-11-19
Use flag set.Func() instead of custom type for -H

### LINKED ISSUE #993: use flag set.Func() instead of custom type for -H, perclist etc
"new" since go 1.16

https://pkg.go.dev/flag#FlagSet.Func

### LINKED ISSUE #1000: doc contributing part is deprecated
At the contributing part of readme write following command: `gofumpt -s -w *.go` 
But it should be use: `gofumpt -w *.go`
Because at the doc of gofumpt said, that -s is hidden and used as default https://github.com/mvdan/gofumpt/blob/86bffd62437a3c437c0b84d5d5ab244824e762fc/README.md?plain=1#L27

proooooogiba:
fix #993
fix #1000

### DIFF
--- a/bincommon/commonflags.go
+++ b/bincommon/commonflags.go
@@ -34,19 +34,6 @@ import (
 	"fortio.org/safecast"
 )
 
-// -- Support for multiple instances of -H flag on cmd line.
-type headersFlagList struct{}
-
-func (f *headersFlagList) String() string {
-	return ""
-}
-
-func (f *headersFlagList) Set(value string) error {
-	return httpOpts.AddAndValidateExtraHeader(value)
-}
-
-// -- end of functions for -H support
-
 // FortioHook is used in fortio/cli and fortio/rapi to customize the run and introduce for instance clienttrace
 // and fortiotel access logger.
 type FortioHook func(*fhttp.HTTPOptions, *periodic.RunnerOptions)
@@ -63,7 +50,6 @@ var (
 	httpsInsecureFlag   = flag.Bool("k", false, "Do not verify certs in HTTPS/TLS/gRPC connections")
 	httpsInsecureFlagL  = flag.Bool("https-insecure", false, "Long form of the -k flag")
 	resolve             = flag.String("resolve", "", "Resolve host name to this `IP`")
-	headersFlags        headersFlagList
 	httpOpts            fhttp.HTTPOptions
 	followRedirectsFlag = flag.Bool("L", false, "Follow redirects (implies -std-client) - do not use for load test")
 	userCredentialsFlag = flag.String("user", "", "User credentials for basic authentication (for HTTP). Input data format"+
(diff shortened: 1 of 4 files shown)


ldemailly (commented):
Thanks ! there are more in fortio_main for the proxies too - can you change these too please?

and did you manually test to confirm it works the same?

bincommon/commonflags.go
```diff
@@ -114,8 +100,9 @@ var (
 // It sets up the common flags, the rest of usage/argument/flag handling
 // is now moved to the [fortio.org/cli] and [fortio.org/scli] packages.
 func SharedMain() {
-	flag.Var(&headersFlags, "H",
-		"Additional HTTP header(s) or gRPC metadata. Multiple `key:value` pairs can be passed using multiple -H.")
+	flag.Func("H",
+		"Additional HTTP header(s) or gRPC metadata. Multiple `key:value` pairs can be passed using multiple -H.",
+		func(s string) error { return httpOpts.AddAndValidateExtraHeader(s) })
```
ldemailly: what the linter says, the extra func is not doing anything

ldemailly (changes requested):

proooooogiba:
1) Replace at fortio_main custom type by Func
2) I find [implementation](https://github.com/golang/go/blob/382b20a09e90d3d96003bd3d22418d79d2e9d2d3/src/flag/flag.go#L980) of flag.Func and under the hood it uses flag.Var, so i don't check it manually. Could you please say, what is the best way to check it manually? I thought, that cli_test will be enough.

bincommon/commonflags.go
proooooogiba: fix it

ldemailly:
I don’t remember if I do have a test for multiple headers through CLI using the echo from debug.fortio.org for instance, maybe should add such txtar test if there isn’t. basically that multiple (and single) -H still works , ditto for proxies etc

ldemailly:
this does test the multiple -H https://github.com/fortio/fortio/blob/master/cli_test.txtar#L21

but for proxies not sure we have a test outside of https://github.com/fortio/fortio-demo-chart/blob/main/helm/fortio/templates/fortio.yaml#L205-L208 but that chart is used but not part of tests

ldemailly (approved):

proooooogiba:
> this does test the multiple -H https://github.com/fortio/fortio/blob/master/cli_test.txtar#L21
> 
> but for proxies not sure we have a test outside of https://github.com/fortio/fortio-demo-chart/blob/main/helm/fortio/templates/fortio.yaml#L205-L208 but that chart is used but not part of tests

Add tests to cli_test.txtar file for the flags

cli_test.txtar
```diff
@@ -23,3 +23,13 @@ stderr 'HTTP/1.1 200 OK'
 stdout 'Xyz: bar blah'
 stdout 'Content-Type: foo/bar'
 stdout 'POST /test-path HTTP/1.1'
+
+# HTTP multi proxy to run (-M flag)
+fortio curl -M '8088 http://fortio-c1-http:8080 http://fortio-c2-http:8080' https://debug.fortio.org/test-path
```
ldemailly: these flags only apply to server mode

proooooogiba: @ldemailly I try to write test for server mode, but i find a problem, that the process of server mode can't be stop and test go to endless loop. 
As i see in testscript [doc](https://arc.net/l/quote/ofdmazkc), using exec with & token terminate command in background process, but it doesn't seem to be truth, because using kill command doesn't help.
So, i have reverted commit with test and checked flags manually - it works same.

ldemailly: server tests exist in webtest.sh (predating txtar which does not do great with multiple processes). Let me see I can do a manual test or add one there

ldemailly (approved):

ldemailly:
did a manual test to confirm

-M
```
09:45:07.309 r1 [INF] http_forwarder.go:317> Multi-server on [::]:5554 running with &{Targets:[{Destination:http://localhost:8080 MirrorOrigin:true} {Destination:http://localhost:8080 MirrorOrigin:true}] Serial:false Name:Multi on [::]:5554 client:0x140003e0d50}
09:45:07.309 r1 Fortio dev Multi on 5555 server listening on tcp [::]:5555
09:45:07.309 r1 [INF] http_forwarder.go:317> Multi-server on [::]:5555 running with &{Targets:[{Destination:http://localhost:8080 MirrorOrigin:true} {Destination:http://localhost:8080 MirrorOrigin:true}] Serial:false Name:Multi on [::]:5555 client:0x140003e0e70}
```

-P
```
09:45:55.830 r1 Fortio dev proxy for 127.0.0.1:8080 server listening on tcp [::]:5554
09:45:55.830 r1 Fortio dev proxy for 127.0.0.1:8080 server listening on tcp [::]:5555
```

PR #1026, opened 2025-04-01
Build image with go 1.23.8

ldemailly:


Dockerfile.build
```diff
@@ -1,5 +1,5 @@
 # Dependencies and linters for build:
-FROM golang:1.23.7@sha256:7fb9682dbc6c9743cd911cedb39c84790e8dfeaaf87d229b78f3f6658fce507a
+FROM golang:1.23.8@sha256:a5339982f2e78b38b26ebbee35139854e184a4e90e1516f9f636371e720b727b
```
ldemailly, line 2: This is the only non mechanical change and the rest stems from that and the steps in release/README.md

geofpwhite (approved):

PR #1053, opened 2025-05-17
Use go 1.23.9 based build image. latest grol and terminal. prep for release

ldemailly:
In addition to latest go patch, the release will have updated x/net, grpc and grol


tejaspbajaj (approved):
lgtm

PR #1060, opened 2025-06-10
refactor: replace sequential assignments with explicit variable decla…

wlynxg:
During my review of the fortio source code, I initially attempted to trace the assignment and usage relationships of fields in the RunnerResults struct. However, I found it challenging to locate the corresponding assignment logic due to the use of sequential struct assignments, which are not properly indexed by LSP (Language Server Protocol) tools.

After further investigation, I discovered the assignment logic in https://github.com/fortio/fortio/blob/master/periodic/periodic.go, where fields were being initialized using sequential struct literal syntax:

```go
// Original sequential assignment (hard to track with LSP)
results := RunnerResults{
    startTime, // LSP cannot resolve which field this corresponds to
    endTime,
    numThreads,
    // ... other fields
}
```

This syntax, while concise, makes it difficult for developers to:
- Quickly identify which struct fields are being initialized
- Use IDE features like "Find Usages" or "Go to Definition"
- Maintain the code safely when struct fields are added/removed

To address this, I refactored the assignments to use explicit field names:

```go
// Refactored to explicit field assignments
results := RunnerResults{
    StartTime:    startTime,
    EndTime:      endTime,
    NumThreads:   numThreads,
    // ... other fields with explicit names
}
```

While this change does not affect runtime behavior, it significantly enhances the developer experience for anyone studying or contributing to fortio's source code.


periodic/periodic.go
```diff
@@ -535,12 +535,27 @@ func (r *periodicRunner) Run() RunnerResults {
...
 		log.Warnf("Run requested to stop before even starting")
 		aborter.Reset()
-		return RunnerResults{ // A bit ugly this is almost the same as the big init below in the normal not early abort case.
-			r.RunType, r.Labels, start, requestedQPS, requestedDuration,
-			0, 0, r.NumThreads, version.Short(), functionDuration.Export().CalcPercentiles(r.Percentiles),
-			errorsDuration.Export().CalcPercentiles(r.Percentiles),
-			r.Exactly, r.Jitter, r.Uniform, r.NoCatchUp, r.RunID, loggerInfo, r.ID,
-			*jrpc.NewErrorReply("Aborted before even starting", nil),
+		
+		return RunnerResults{
```
ldemailly, line 540:
```suggestion
		// A bit ugly this is almost the same as the big init below in the normal not early abort case.
		return RunnerResults{
```

ldemailly (commented):
Thanks, yeah it was quite bad to just do unnamed - the comment about the other badness should remain though

ldemailly (approved):
thx! will merge when tests pass

PR #1090, opened 2025-09-12
Use the new duration for uptime and -t

ldemailly:


asha-oss (approved):

geofpwhite (commented):
👍

README.md
```diff
@@ -1,4 +1,4 @@
-<!-- 1.71.2 -->
+<!-- 1.71.3 -->
```
ldemailly, line 1: that turns out was the wrong bump as it was then tagged 1.72.0 in releases/ so have to now fix #1097 to use 1.73.0

ldemailly: fixed in #1098

PR #1160, opened 2026-08-29
addressed cveCVE-2026-39822

morampudisouji:


ldemailly:
Thanks. but I will make a new image (including with the right sha as per build readme). 1.25.12 is wrong anyway, should be .14

ldemailly:
oh and hi soujanya, sorry didn't realize it was you :) the build image itself isn't pushed by CI (it could/should I guess) but I do it

ldemailly:
See #1161 and accept github fortio org invite to review/approve it

